PR #467, opened 2017-01-13
Typo fix in documentation

agl:
Changing `an` to `a` seems correct, but I took a guess at what the next bit should have been.


cesarghali:
Thanks Adam for the fixes!

PR #469, opened 2017-01-14
docs: Fix link to protobuf definitions

aaronlehmann:
design.md is located a level below project root, so the link needs to start with `../impl` instead of `impl`.

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -116,7 +116,7 @@ recipient has installed.
 
 ## API Reference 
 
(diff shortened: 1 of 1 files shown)


aaronlehmann:
I believe Docker Inc has a CLA in place that would cover this contribution. I'm a bit fuzzy on the details though.

gdbelvin:
Fixed in #470 
Thank you for your contribution though

PR #470, opened 2017-01-16
Fix protobuf link in design.md.

rektide:
Currently 404. This PR fixes the link in design.md to the protocol buffers spec for the service.


rektide:
Not signing a CLA and handing over a bunch of personal info to fix this. Your stuff is broke. Have a nice day.

PR #477, opened 2017-01-16
Upgrade to pkg/context

gdbelvin:
Closes #473

### DIFF
--- a/cmd/key-transparency-server/frontend.go
+++ b/cmd/key-transparency-server/frontend.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"context"
 	"database/sql"
 	"flag"
 	"io/ioutil"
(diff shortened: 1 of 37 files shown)


cmd/key-transparency-server/frontend.go
```diff
@@ -40,7 +41,7 @@ import (
 	"github.com/golang/protobuf/jsonpb"
 	"github.com/golang/protobuf/proto"
 	"github.com/grpc-ecosystem/grpc-gateway/runtime"
-	"golang.org/x/net/context"
+	oldcontext "golang.org/x/net/context"
```
cesarghali: Hopefully gRPC will close issue grpc/grpc-go#711 and we won't have to use two context objects.

gdbelvin:
Looks like there isn't a clean way to do this until Go 1.8 or later. I'll close this PR until gRPC updates.

Issue #479, opened 2017-01-16
Add RSA Signature Scheme in core/signatures


cesarghali:
This requires:

* [x] Add RSA implementation of `Signer` and `Verifier` in `core/crypto/signatures`.
* [x] Add entry point to `core/crypto/signatures/factory`.
* [ ] Add client ability to generate RSA keys and store them in KeyStore.

Implementing this feature will be done with (at least) three Pull Requests.

cesarghali:
For the time being, we will not allow client to use RSA keys to sign updates.

PR #480, opened 2017-01-17
Add RSA Implementation of Signer and Verifier in core/signatures

cesarghali:
Necessary tests are added.

Partial #479.


gdbelvin:
This is a low priority feature at the moment. Let's hold off on adding older and less efficient crypto primitives  until we have a fully working and tested system.

cesarghali:
Sounds good. I will keep this PR open until we decide to merge it.

However, if we consider RSA to be old and less efficient, should we consider not supporting it in the first place?

gdbelvin:
I'm happy for this to go in now. 
Could you rebase and resubmit?

cesarghali:
This PR is rebased. PTAL.

gdbelvin:
Thanks for this PR!

PR #490, opened 2017-02-10
Minor fix

cesarghali:
This PR changes the following:
* All RPC call handlers return gRPC errors after logging the actual error.
* All functions that are called by RPC call handlers return standard `error` objects.

In other words, all exported functions (which are indeed gRPC handlers) return gRPC errors. All unexported functions don't return gRPC errors.

### DIFF
--- a/core/keyserver/keyserver.go
+++ b/core/keyserver/keyserver.go
@@ -263,7 +263,7 @@ func (s *Server) UpdateEntry(ctx context.Context, in *tpb.UpdateEntryRequest) (*
 		log.Printf("Enqueue error: %v", err)
 		return nil, grpc.Errorf(codes.Internal, "Write error")
 	}
-	return &tpb.UpdateEntryResponse{Proof: resp}, err
+	return &tpb.UpdateEntryResponse{Proof: resp}, nil
 }
 
 func (s *Server) saveCommitment(ctx context.Context, kv *tpb.KeyValue, committed *tpb.Committed) error {


core/keyserver/keyserver.go
```diff
@@ -102,8 +103,7 @@ func (s *Server) getEntry(ctx context.Context, userID string, epoch int64) (*tpb
 
 	neighbors, err := s.tree.NeighborsAt(ctx, index[:], epoch)
 	if err != nil {
-		log.Printf("Cannot get neighbors list: %v", err)
```
gdbelvin: code in /keyserver/ should only report grpc errors.

cesarghali: This error is returned by an unexported function `getEntry`. All exported functions, which are gRPC handlers return gRPC errors.

cesarghali: In other words, all exported functions (which are indeed gRPC handlers) return gRPC errors. All unexported functions don't return gRPC errors.

gdbelvin: Hmm... Let's chat more about this Monday. I worry that the exported / unexported rule will result in a lot of error conversion code.  I think I prefer the server module / non-server-module deliniation

cesarghali: I don't have a preference between exported/unexported and server-module/non-server-module. We can decide on either. The important thing is to have a consistent error generation/handling rule. Currently, `keyserver.go` functions return both `error` and gRPC error objects. The good thing is that gRPC handlers always return gRPC errors.

PR #491, opened 2017-02-11
Add Service Key to Client RPC Calls

cesarghali:
Both `GetEntry` and `ListEntryHistory` APIs are unauthenticated. However, in some situations, a service key is required for them to work. This PR allows the gRPC client to use a service key to generate RPC call credentials.


gdbelvin:
This is great. One tweak though: 
Since the service key is a simple string, I'd prefer to have it as a setting in the .json config file rather than it's own separate file.

PR #493, opened 2017-02-13
Separate and move Signature to crypto/signatures

gdbelvin:
This supports separating digital signing from other functions.
In the future, this module can be moved or an external library could be used.


core/proto/signature/gen.go
```diff
@@ -0,0 +1,17 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
```
cesarghali: We call this file `doc.go` in other places. Should we keep the same naming consistent? I'm fine with either.

gdbelvin: we should probably migrate them all to gen.go at some point, but it's not high priority. 
Thanks for noticing!

PR #496, opened 2017-02-15
Ensure that pb.go files are up-to-date

cesarghali:
If we run `go generate ./...` some proto output files gets updated. This PR submits those updates.


AlCutter:
You might be interested in the [travis config](https://github.com/google/trillian/blob/master/.travis.yml) and [presubmit script](https://github.com/google/trillian/blob/master/scripts/presubmit.sh) we have over in Trillian-land for automatically detecting/failing PRs which have out-of-date generated files.

gdbelvin:
Since the old pb files were not creating errors, I'd prefer to not put this
in the presubmit.
Developers should run generate only as needed. The unit tests should verify
that the pb and the rest of the code base is consistent.

PR #501, opened 2017-02-22
Start a Trillian Log Server inside the front end.

gdbelvin:
In prod, this line can be replaced with a trillian/ecosystem/proxy

### DIFF
--- a/cmd/keytransparency-server/frontend.go
+++ b/cmd/keytransparency-server/frontend.go
@@ -36,6 +36,11 @@ import (
 	"github.com/google/keytransparency/impl/sql/sqlhist"
 	"github.com/google/keytransparency/impl/transaction"
 
+	"github.com/google/trillian"
+	"github.com/google/trillian/extension/builtin"
+	"github.com/google/trillian/server"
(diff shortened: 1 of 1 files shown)


cmd/keytransparency-server/frontend.go
```diff
@@ -208,6 +229,7 @@ func main() {
 	}
 	grpcServer := grpc.NewServer(opts...)
 	pb.RegisterKeyTransparencyServiceServer(grpcServer, svr)
+	trillian.RegisterTrillianLogServer(grpcServer, logServer)
```
cesarghali: This PR looks good to me. I am just wondering what are pros and cons of running trillian in the same binary with Key Transparency as opposed to running it in its own binary. Wouldn't this line expose all trillian gRPC APIs on the same address and port of Key Transparency? Is that what we want to do?

gdbelvin: The pros:
- External users of our code won't have to run multiple binaries. 

Cons: 
- A little less process separation, but this can be adjusted by replacing this bit of code with the proxy.

PR #503, opened 2017-02-24
Use RSA in Client Binary

### LINKED ISSUE #479: Add RSA Signature Scheme in core/signatures

cesarghali:
Should merge #502 first.
Closes #479.

**Note:**
Please do not update this branch or merge this PR. I need to clean it up after merging #502. Thanks.

### DIFF
--- a/cmd/keytransparency-client/cmd/keys.go
+++ b/cmd/keytransparency-client/cmd/keys.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/google/keytransparency/core/crypto/keymaster"
 	"github.com/google/keytransparency/core/crypto/signatures/p256"
+	"github.com/google/keytransparency/core/crypto/signatures/rsa"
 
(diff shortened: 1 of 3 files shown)


gdbelvin:
Hmm...  I think I'd prefer to not have this option in the client.  
Fewer options makes the client easier to use, and with good defaults like the ones we've chosen, also safer and more efficient.  What do you think?

Thanks for adding RSA support to the signature library though!

cesarghali:
I was actually debating whether to add this feature to clients and I wanted to get your opinion on it. I am closing this PR now.

PR #513, opened 2017-03-15
Save SignedMapHeads to Trillian Log

### LINKED ISSUE #384: Replace XJSON server with Trillian Append Only Log
Transition to use Google Trillian's Verifiable Map implementation to achieve at-scale performance for 10^12 accounts.  

Components: 
- [x] Trillian Log Client google/trillian#297
- [ ] Send SMH's to a Trillian endpoint.
- [ ] Run co-located Trillian Log. 
- [ ] Proxy Trillian Log requests in the front-ends.
- [ ] Host Trillian Log Process on same port to avoid running a second binary?

gdbelvin:
This replaces the Certificate Transparency client with a Trillian Log Client for sending Signed Map Heads to an append only log. 

- Launches a Trillian front and backend in `goreman`. 
- Removes SCT verification logic from the client, since Trillian does not support SCTs. 

Closes #384


cmd/keytransparency-signer/backend.go
```diff
@@ -97,10 +124,22 @@ func main() {
 		log.Fatalf("Failed to create SQL history: %v", err)
 	}
 	mutator := entry.New()
-	sths, err := appender.New(context.Background(), sqldb, *mapID, *mapLogURL, nil)
+
+	// STH Appender.
+	// The log should be in a restricted access backend environment.
+	cc, err := grpc.Dial(*logURL, grpc.WithInsecure())
```
cesarghali: By the way, I really like that we can now talk to the log using gRPC instead of HTTP.

impl/sql/appender/trillian.go
```diff
@@ -0,0 +1,79 @@
...
+		// TODO(gbelvin): Add leaf at a specific index. trillian#423
+		if err := t.log.AddLeaf(ctx, b); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// Epoch retrieves a given object at a particular index.
+func (t *Trillian) Epoch(ctx context.Context, epoch int64, obj interface{}) ([]byte, error) {
```
cesarghali: I'm not sure but I think tests will not pass without implementing `Epoch` and `Latest`, right?

impl/sql/appender/trillian.go
```diff
@@ -0,0 +1,79 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
```
cesarghali: Can you please add tests for this?

cmd/keytransparency-signer/backend.go
```diff
@@ -15,14 +15,20 @@
...
+	"encoding/pem"
 	"flag"
+	"fmt"
 	"io/ioutil"
 	"log"
 	"strings"
 	"sync"
 	"time"
 
+	"google.golang.org/grpc"
```

cmd/keytransparency-signer/backend.go
gdbelvin: Great point. Trillian does not not current support HTTPS connections. 
I raised this question in https://github.com/google/trillian/issues/428 

I think this is because Trillian is not designed to ever be a front end / exposed service. It also lacks other front end type things like authentication, quota, and anti-abuse mechanisms.

core/signer/signer.go
```diff
@@ -94,6 +94,7 @@ func (s *Signer) StartSigning(interval time.Duration) {
 // ProcessMutation saves a mutation and adds it to the append-only log and tree.
 func (s *Signer) ProcessMutation(ctx context.Context, txn transaction.Txn, index, mutation []byte) error {
 	// Send mutation to append-only log.
+	// TODO(gbelvin): Store in databse. #507
```
gdbelvin: done

scripts/prepare_server.sh
```diff
@@ -212,7 +212,7 @@ fi
 
 if ((BACKEND == 1)); then
     PROCFILE="${PROCFILE}
-backend: ./keytransparency-signer --domain=\$DOMAIN --db=\$DB --maplog=\$CTLOG --etcd=\$LISTEN --period=\$SIGN_PERIOD_SEC --key=\$SIGN_KEY"
+backend: ./keytransparency-signer --domain=\$DOMAIN --db=\$DB --maplog=\$CTLOG --etcd=\$LISTEN --period=\$SIGN_PERIOD --key=\$SIGN_KEY"
```
gdbelvin: google/keytransparency/issues#516

impl/sql/appender/objhasher.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"crypto"
+
+	"github.com/benlaurie/objecthash/go/objecthash"
+)
+
+// Domain separation prefixes
+const (
+	RFC6962LeafHashPrefix = 0
```
gdbelvin: Good question.  These prefixes provide "domain separation" in the Merkle Tree, preventing leaf hashes from being glued to interior nodes etc. 

ObjectHash uses prefixes as well to distinguish types. None of these prefixes use `1` as a value, so I believe we retain the domain separation property. 

I'm not sure it makes sense to try to add an additional domain separator to the output of a hash.  We'd end up with a double hash like `H(1 || ObjectHash(obj))`

gdbelvin:
This is currently blocked on google/trillian#465

.env
```diff
@@ -20,3 +20,8 @@ CERT="genfiles/server.crt"
 VRF_PRIV="genfiles/vrf-key.pem"
 VRF_PUB="genfiles/vrf-pubkey.pem"
 SIGN_KEY="genfiles/p256-key.pem"
+TRILLIAN_LOG_KEY_PWD=""
+TRILLIAN_LOG_KEY_FILE="genfiles/p256-key.pem"
+TRILLIAN_LOG_PUBKEY_FILE="genfiles/p256-pubkey.pem"
```
gdbelvin: Done

benlaurie:
Meta-comment. I still think not all maps require this, so can it be optional? Maybe it is already?

gdbelvin:
Key Transparency would be considered a Trillian Map personality, so having it here doesn't force it on other maps.

gdbelvin:
This is now ready for another look!  I know this PR is big. Let me know if you'd like it broken up.

cmd/keytransparency-client/cmd/cache.go
```diff
@@ -1,44 +0,0 @@
-// Copyright 2016 Google Inc. All Rights Reserved.
```
gdbelvin: Trillian log does not support SignedCertificateTimestamps (SCTs). Previously, SCTs needed to be saved to disk and verified later. Now all verification is immediate, thus no need to verify previously retrieved results anymore.

cmd/keytransparency-client/cmd/root.go
```diff
@@ -29,7 +29,9 @@ import (
 	"github.com/google/keytransparency/core/crypto/signatures"
 	"github.com/google/keytransparency/core/vrf"
 	"github.com/google/keytransparency/core/vrf/p256"
+	"github.com/google/keytransparency/impl/config"
 	"github.com/google/keytransparency/impl/google/authentication"
+	"github.com/google/trillian/client"
```
gdbelvin: done

cmd/keytransparency-client/cmd/root.go
```diff
@@ -79,9 +77,9 @@ func init() {
 	cobra.OnInitialize(initConfig)
 	RootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.keytransparency.yaml)")
 	RootCmd.PersistentFlags().String("vrf", "testdata/vrf-pubkey.pem", "path to vrf public key")
-	RootCmd.PersistentFlags().String("ct-url", "", "URL of Certificate Transparency server")
-	RootCmd.PersistentFlags().String("ct-key", "testdata/ct-server-key-public.pem", "Path to public key PEM for Certificate Transparency server")
-	RootCmd.PersistentFlags().String("ct-scts", ".keytransparency-scts.dat", "Path to load/save unverified SCT state from")
+	RootCmd.PersistentFlags().Int64("logid", 0, "Log ID of the backend log server")
+	RootCmd.PersistentFlags().String("log-url", "", "URL of Certificate Transparency server")
+	RootCmd.PersistentFlags().String("log-key", "testdata/ct-server-key-public.pem", "Path to public key PEM for Certificate Transparency server")
```
gdbelvin: I'll just remove the default value because this file isn't provided by our repo

cmd/keytransparency-client/grpcc/grpc_client.go
```diff
@@ -25,14 +25,14 @@ import (
...
 
-	"github.com/google/keytransparency/core/client/ctlog"
 	"github.com/google/keytransparency/core/client/kt"
 	"github.com/google/keytransparency/core/crypto/signatures"
 	"github.com/google/keytransparency/core/mutator"
 	"github.com/google/keytransparency/core/mutator/entry"
 	"github.com/google/keytransparency/core/tree/sparse"
 	tv "github.com/google/keytransparency/core/tree/sparse/verifier"
 	"github.com/google/keytransparency/core/vrf"
+	"github.com/google/trillian/client"
```
gdbelvin: done

gdbelvin:
Updated and ready for another review. 
This PR focuses on integrating Trillian Log into the verification client.

scripts/prepare_client.sh
```diff
@@ -70,8 +70,8 @@ cd "${GOPATH}/src/github.com/google/keytransparency"
 go build ./cmd/keytransparency-client
 
 # Generate .keytransparency.yaml file.
-KTYAML="ct-key: \"../certificate-transparency/test/testdata/ct-server-key-public.pem\"
-ct-url: \"http://107.178.246.112\"
```
cesarghali: Do we need to provide a log ID here?

cesarghali: Also, can you please modify `.keytransparency.yaml` to reflect the same default values?

cmd/keytransparency-client/cmd/root.go
```diff
@@ -264,22 +242,28 @@ func dial(ktURL, caFile, clientSecretFile string, serviceKeyFile string) (*grpc.
...
 func GetClient(clientSecretFile string) (*grpcc.Client, error) {
+	vrfFile := viper.GetString("vrf")
 	ktURL := viper.GetString("kt-url")
 	ktPEM := viper.GetString("kt-key")
 	ktSig := viper.GetString("kt-sig")
 	mapID := viper.GetInt64("mapid")
-	ctURL := viper.GetString("ct-url")
-	ctPEM := viper.GetString("ct-key")
-	vrfFile := viper.GetString("vrf")
+	logID := viper.GetInt64("logid")
```
gdbelvin: done

scripts/prepare_client.sh
gdbelvin: done

Issue #507, opened 2017-03-08
Supply API for listing Mutations


gdbelvin:
- [x] Store Mutations in a local DB table.
- [x] Create API for listing mutations.

Issue #487, opened 2017-01-31
Include Sequence Number in Mutations


gdbelvin:
Certificate Transparency does not guarantee that items will be sequenced in the same order they are submitted. We can either provide our own definitive sequencing, or wait for the Trillian Log to sequence. In order to avoid taking on a new production performance dependency, let's provide our own ordering via our own sequence number before submitting to Trillian.  

- We might even be able to supply our own sequence number to Trillian. 
- Monitors will need to explicitly order by the provided sequence number rather than Trillian ordering.  @cesarghali

cesarghali:
Although we discussed this offline and realized that we might not need sequence numbering here, I think we still need it to allow paging. Monitors might call the `GetMutations` API (or whatever we end up calling it) and request to get mutations in pages. There should be a way to index into the list of mutations in order to support paging.

gdbelvin:
Sounds good. We can either page by mutation index or epoch.

cesarghali:
Correct. This is what I have in mind for the API:

* Monitors call `GetMutations` in epoch `e` starting from the "beginning" of mutation index and with page size of `p`.
* Server returns the requested data and sets the next index to be "beginning" + `p`.

This is very similar to what we have in history. I am working now on adding sequence numbers to mutations.

PR #514, opened 2017-03-16
Store Mutations in Local Database

cesarghali:
This PR stores mutations in the local database after being processed by the queue/signer. It also stop sending mutations to a CT log.

Partial #507.

### DIFF
--- a/core/mutator/mutator.go
+++ b/core/mutator/mutator.go
@@ -55,7 +55,7 @@ type Mutator interface {
 // Mutation reads and writes mutations to the database.
 type Mutation interface {
 	// Read reads all mutations for a specific given mapID, epoch, and index.
-	Read(ctx context.Context, txn transaction.Txn, epoch int64, index []byte) ([][]byte, error)
+	Read(ctx context.Context, txn transaction.Txn, epoch int64, index []byte) ([]byte, error)
 	// Write saves the mutation in the database.
 	Write(ctx context.Context, txn transaction.Txn, epoch int64, index, mutation []byte) error
 }
--- a/core/mutator/mutator.go
+++ b/core/mutator/mutator.go
@@ -13,10 +13,16 @@
 // limitations under the License.
 
 // Package mutator defines the operations to transform mutations into changes in
-// the map.
+// the map as well as operations to write and read mutations to and from the
+// database.
 package mutator
 
-import "errors"
+import (
+	"context"
+	"errors"
+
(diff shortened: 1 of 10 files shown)


gdbelvin:
It might be helpful to first sketch out the API that we'd like to support for mutations first:
- How should clients "page" through mutations, especially if there are a lot of them? 
- How should clients know what mutations belong to what epoch?

core/signer/signer.go
```diff
@@ -93,11 +93,6 @@ func (s *Signer) StartSigning(interval time.Duration) {
 
 // ProcessMutation saves a mutation and adds it to the append-only log and tree.
 func (s *Signer) ProcessMutation(ctx context.Context, txn transaction.Txn, index, mutation []byte) error {
-	// Send mutation to append-only log.
```
gdbelvin: Consider whether it might be possible to reuse or extend the appender library by just a bit? Might save some work.

core/mutator/mutator.go
```diff
@@ -45,3 +51,11 @@ type Mutator interface {
 	// Mutate applies mutation to value
 	Mutate(value, mutation []byte) ([]byte, error)
 }
+
+// Mutation reads and writes mutations to the database.
+type Mutation interface {
+	// Read reads all mutations for a specific given mapID, epoch, and index.
+	Read(ctx context.Context, txn transaction.Txn, epoch int64, index []byte) ([][]byte, error)
```
cesarghali: For now, there's no ordering, this will be addressed in #487.
However, it's a good idea to modify the database to only store a single mutation per epoch per index.

impl/sql/mutations/mutations.go
```diff
@@ -0,0 +1,151 @@
...
+		`
+	CREATE TABLE IF NOT EXISTS Maps (
+		MapID   VARCHAR(32) NOT NULL,
+		PRIMARY KEY(MapID)
+	);`,
+		`
+	CREATE TABLE IF NOT EXISTS Mutations (
+		MapID      VARCHAR(32)   NOT NULL,
+		Epoch      INTEGER       NOT NULL,
+                MutationID VARBINARY(32) NOT NULL,
```
gdbelvin: It looks like mutation ID is the index, this is slightly confusing

cesarghali:
Thanks for the approval Gary. Travis is broken due to grpc-gateway update. Since you mentioned you're working on a fix, I will wait for that before merging this PR.

PR #524, opened 2017-03-20
Add Mutation Sequence Number

### LINKED ISSUE #487: Include Sequence Number in Mutations

cesarghali:
This PR includes (in separate commits):
* Implementation of a configuration module as a key-value store in the database. This will be used to store the mutation sequence number as well as other Key Transparency server-side configurations.
* Save mutation sequence number along with the mutations themselves. The sequence number is incremented for every mutation.

Closes #487.

### DIFF
--- a/core/mutator/mutator.go
+++ b/core/mutator/mutator.go
@@ -18,7 +18,6 @@
 package mutator
 
 import (
-	"context"
 	"errors"
 
 	"github.com/google/keytransparency/core/transaction"
@@ -52,15 +51,23 @@ type Mutator interface {
 	Mutate(value, mutation []byte) ([]byte, error)
 }
 
+// MutationInfo contains information related mutations, specifically the index
+// and the mutation data.
+type MutationInfo struct {
+	// Index determines the index of the mutation in the map.
+	Index []byte
+	// Data contains the actual mutation data.
+	Data []byte
+}
+
 // Mutation reads and writes mutations to the database.
 // TODO: Add mapID to this interface to support multiple maps per server.
 type Mutation interface {
-	// Read reads all mutations for a specific given mapID, sequence, and
-	// index.
-	Read(ctx context.Context, txn transaction.Txn, sequence uint64, index []byte) ([]byte, error)
-	// Write saves the mutation in the database. Sequence might not be used
-	// by all databases engines. For instance, in MySQL and SQLite, sequence
-	// numbers are automatically added at the database level. Write returns
-	// the sequence number that is written.
-	Write(ctx context.Context, txn transaction.Txn, sequence uint64, index, mutation []byte) (uint64, error)
+	// ReadRange reads all mutations for a specific given mapID and sequence
+	// range. The range is identified by a starting sequence number and a
+	// count.
+	ReadRange(txn transaction.Txn, startSequence uint64, count int) ([]MutationInfo, error)
+	// Write saves the mutation in the database. Write returns the sequence
+	// number that is written.
+	Write(txn transaction.Txn, index, mutation []byte) (uint64, error)
 }
--- a/core/mutator/mutator.go
+++ b/core/mutator/mutator.go
@@ -53,10 +53,14 @@ type Mutator interface {
 }
 
 // Mutation reads and writes mutations to the database.
+// TODO: Add mapID to this interface to support multiple maps per server.
 type Mutation interface {
-	// Read reads all mutations for a specific given mapID, epoch, and index.
-	// TODO(cesarghali): Add mapID to this interface to support multiple maps per server.
-	Read(ctx context.Context, txn transaction.Txn, epoch int64, index []byte) ([]byte, error)
-	// Write saves the mutation in the database.
-	Write(ctx context.Context, txn transaction.Txn, epoch int64, index, mutation []byte) error
+	// Read reads all mutations for a specific given mapID, sequence, and
+	// index.
+	Read(ctx context.Context, txn transaction.Txn, sequence uint64, index []byte) ([]byte, error)
+	// Write saves the mutation in the database. Sequence might not be used
+	// by all databases engines. For instance, in MySQL and SQLite, sequence
+	// numbers are automatically added at the database level. Write returns
+	// the sequence number that is written.
+	Write(ctx context.Context, txn transaction.Txn, sequence uint64, index, mutation []byte) (uint64, error)
 }
--- a/core/mutator/mutator.go
+++ b/core/mutator/mutator.go
@@ -21,6 +21,8 @@ import (
 	"errors"
 
 	"github.com/google/keytransparency/core/transaction"
+
+	tpb "github.com/google/keytransparency/core/proto/keytransparency_v1_types"
 )
 
 var (
@@ -51,23 +53,14 @@ type Mutator interface {
 	Mutate(value, mutation []byte) ([]byte, error)
 }
 
-// MutationInfo contains information related mutations, specifically the index
-// and the mutation data.
-type MutationInfo struct {
-	// Index determines the index of the mutation in the map.
-	Index []byte
-	// Data contains the actual mutation data.
-	Data []byte
-}
-
 // Mutation reads and writes mutations to the database.
 // TODO: Add mapID to this interface to support multiple maps per server.
 type Mutation interface {
 	// ReadRange reads all mutations for a specific given mapID and sequence
 	// range. The range is identified by a starting sequence number and a
 	// count.
-	ReadRange(txn transaction.Txn, startSequence uint64, count int) ([]MutationInfo, error)
+	ReadRange(txn transaction.Txn, startSequence uint64, count int) ([]*tpb.SignedKV, error)
 	// Write saves the mutation in the database. Write returns the sequence
 	// number that is written.
-	Write(txn transaction.Txn, index, mutation []byte) (uint64, error)
+	Write(txn transaction.Txn, mutation *tpb.SignedKV) (uint64, error)
 }
--- a/impl/etcd/queue/etcd.go
+++ b/impl/etcd/queue/etcd.go
@@ -164,9 +164,7 @@ func processEntry(ctx context.Context, txn ctxn.Txn, cbs callbacks, dataKV kv) e
 	if dataKV.AdvanceEpoch {
 		return cbs.advanceFunc(ctx, txn)
 	}
-	// Sequence number is automatically added. However, something, e.g., has
-	// to be passed to satisfy the API.
-	return cbs.processFunc(ctx, txn, 0, dataKV.Key, dataKV.Val)
+	return cbs.processFunc(ctx, txn, dataKV.Key, dataKV.Val)
 }
 
 // Close stops the receiver from receiving items from the queue.
--- a/impl/etcd/queue/etcd.go
+++ b/impl/etcd/queue/etcd.go
@@ -164,7 +164,9 @@ func processEntry(ctx context.Context, txn ctxn.Txn, cbs callbacks, dataKV kv) e
 	if dataKV.AdvanceEpoch {
 		return cbs.advanceFunc(ctx, txn)
 	}
-	return cbs.processFunc(ctx, txn, dataKV.Key, dataKV.Val)
+	// Sequence number is automatically added. However, something, e.g., has
+	// to be passed to satisfy the API.
+	return cbs.processFunc(ctx, txn, 0, dataKV.Key, dataKV.Val)
 }
 
 // Close stops the receiver from receiving items from the queue.
--- a/impl/sql/mutations/mutations.go
+++ b/impl/sql/mutations/mutations.go
@@ -20,8 +20,11 @@ import (
 	"database/sql"
 	"fmt"
 
+	"github.com/gogo/protobuf/proto"
 	"github.com/google/keytransparency/core/mutator"
 	"github.com/google/keytransparency/core/transaction"
+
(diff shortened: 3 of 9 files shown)


core/config/config.go
```diff
@@ -0,0 +1,36 @@
...
+	"golang.org/x/net/context"
+)
+
+// Config defines a key-value store for Key Transparency configuration that is
+// stored in the database. The key is a string and value is a byte slice.
+type Config interface {
+	// Create creates a new config key-value pair in the database and return
+	// the corresponding value. If the key already exists, its current value
+	// is returned.
+	Create(ctx context.Context, txn transaction.Txn, key string, value []byte) ([]byte, error)
```
gdbelvin: This sounds more like WriteOrIgnore?

core/mutator/mutator.go
```diff
@@ -55,7 +55,7 @@ type Mutator interface {
 // Mutation reads and writes mutations to the database.
 type Mutation interface {
 	// Read reads all mutations for a specific given mapID, epoch, and index.
-	Read(ctx context.Context, txn transaction.Txn, epoch int64, index []byte) ([]byte, error)
+	Read(ctx context.Context, txn transaction.Txn, epoch int64, index []byte) (uint64, []byte, error)
 	// Write saves the mutation in the database.
-	Write(ctx context.Context, txn transaction.Txn, epoch int64, index, mutation []byte) error
+	Write(ctx context.Context, txn transaction.Txn, epoch int64, index []byte, sequence uint64, mutation []byte) error
```
gdbelvin: I would put the sequence closer to epoch in the argument ordering.

impl/etcd/queue/etcd.go
```diff
@@ -145,11 +170,26 @@ func (q *Queue) dequeue(key, value []byte, rev int64, cbs callbacks) error {
 		return err
 	}
 
+	// Read mutation sequence number.
+	bSequence, err := q.srvConfig.Read(q.ctx, txn, sequenceConfigKey)
```
gdbelvin: I think this is problematic because it creates a single row that's operating as a global sequence counter. That row will get hot. A few ideas:

- Use the MySQL incrementing counter here. 
- Add support for time based ordering with an implicit sequence number (ala spanner)
- Write the sequence number in batches within an epoch. ie. reset the sequence number to 0 at the start of each epoch. The idea is that we should probably write all the mutation that are happening in an epoch together, and we'll know for sure which order we want them in.

impl/etcd/queue/etcd.go
```diff
@@ -145,11 +170,26 @@ func (q *Queue) dequeue(key, value []byte, rev int64, cbs callbacks) error {
...
+	if err != nil {
+		return err
+	}
+	sequence := binary.LittleEndian.Uint64(bSequence)
```
gdbelvin: It loos like these binary conversions are creeping into the code. It would be better to have a slightly custom interface with the types that we want and to do all the serialization there.

cesarghali:
This PR has been significantly modified, hence the force push. It is much simpler and cleaner now. PTAL.

core/mutator/mutator.go
```diff
@@ -55,7 +55,9 @@ type Mutator interface {
 // Mutation reads and writes mutations to the database.
 type Mutation interface {
 	// Read reads all mutations for a specific given mapID, epoch, and index.
```
gdbelvin: Please document what the return values are

core/mutator/mutator.go
```diff
@@ -55,7 +55,9 @@ type Mutator interface {
...
-	Read(ctx context.Context, txn transaction.Txn, epoch int64, index []byte) ([]byte, error)
-	// Write saves the mutation in the database.
-	Write(ctx context.Context, txn transaction.Txn, epoch int64, index, mutation []byte) error
+	Read(ctx context.Context, txn transaction.Txn, epoch int64, index []byte) (uint64, []byte, error)
+	// Write saves the mutation in the database. Sequence might not be used
+	// by all databases engines. For instance, in MySQL and SQLite, sequence
+	// numbers are automatically added at the database level.
+	Write(ctx context.Context, txn transaction.Txn, sequence uint64, epoch int64, index, mutation []byte) error
```
gdbelvin: It seems to me that getting a reliable sequence number at this point is really hard, especially when there are multiple readers from the queue.  What DB engines would need this?

impl/etcd/queue/etcd.go
```diff
@@ -163,7 +163,9 @@ func processEntry(ctx context.Context, txn ctxn.Txn, cbs callbacks, dataKV kv) e
 	if dataKV.AdvanceEpoch {
 		return cbs.advanceFunc(ctx, txn)
 	}
-	return cbs.processFunc(ctx, txn, dataKV.Key, dataKV.Val)
+	// Sequence number is automatically added. However, something, e.g., has
```
gdbelvin: Passing fake data to an API seems less than ideal...

core/mutator/mutator.go
gdbelvin: If possible, it would also be nice to avoid needing to know the current epoch when writing mutation here too.

gdbelvin:
Thanks for this update! I think we can loosen the API requirements around paging through mutations to make the mutation API here simpler and less likely for callers to mix up.

Specifically, if we relax the requirements from paginating through mutations to paginating through epochs, I think we can remove the need for the mutation API callers to know what the sequence number, or even what the current epoch is.

impl/etcd/queue/etcd.go
cesarghali: I don't like this neither, but this is the only way I was able to make the API consistent for both external and internal implementations.

core/mutator/mutator.go
cesarghali: I like the idea of not having epoch number in the `Mutations` table. Though, knowing this epoch number here is free. Signer's `ProcessMutation` function already knows the epoch number by calling `QueueLeaf`. This happens before writing mutations in the database.

core/mutator/mutator.go
cesarghali: Done.

core/mutator/mutator.go
```diff
@@ -53,10 +53,14 @@ type Mutator interface {
...
+// TODO: Add mapID to this interface to support multiple maps per server.
 type Mutation interface {
-	// Read reads all mutations for a specific given mapID, epoch, and index.
-	// TODO(cesarghali): Add mapID to this interface to support multiple maps per server.
-	Read(ctx context.Context, txn transaction.Txn, epoch int64, index []byte) ([]byte, error)
-	// Write saves the mutation in the database.
-	Write(ctx context.Context, txn transaction.Txn, epoch int64, index, mutation []byte) error
+	// Read reads all mutations for a specific given mapID, sequence, and
+	// index.
+	Read(ctx context.Context, txn transaction.Txn, sequence uint64, index []byte) ([]byte, error)
```
gdbelvin: Do we need to pass both `ctx` and `txn`?  Transactions contain contexts. Consider just passing `txn`

core/mutator/mutator.go
```diff
@@ -53,10 +53,14 @@ type Mutator interface {
...
+	// Write saves the mutation in the database. Sequence might not be used
+	// by all databases engines. For instance, in MySQL and SQLite, sequence
+	// numbers are automatically added at the database level. Write returns
+	// the sequence number that is written.
+	Write(ctx context.Context, txn transaction.Txn, sequence uint64, index, mutation []byte) (uint64, error)
```
gdbelvin: When the `Write` API gets called, the caller won't know what the sequence number is, because it gets added by the DB. This is the case for both internal and external implementations. Consider removing it.

core/mutator/mutator.go
gdbelvin: Clients of this API won't necessarily know what sequence number to use here. All they will know is that the sequence number of the last thing they submitted will be greater than anything else in the DB. 

The sequencer and monitors will need to `Read` a set of mutations from `A` to either `B` or `latest`. 

Consider `ReadRange(txn, mapID, start, count/end)` as an alternative?

impl/etcd/queue/etcd.go
gdbelvin: See above comments on API

impl/sql/mutations/mutations.go
```diff
@@ -95,17 +76,24 @@ func (m *mutations) Read(ctx context.Context, txn transaction.Txn, epoch int64,
 	return mutation, nil
 }
 
-// Write saves the mutation in the database.
-func (m *mutations) Write(ctx context.Context, txn transaction.Txn, epoch int64, index, mutation []byte) error {
+// Write saves the mutation in the database. In MySQL or SQLite implementation,
+// sequence numbers are automatically added. Therefore, sequence variable is not
```
gdbelvin: I think we need to remove `sequence` from the `Write` API. WDYT?

impl/sql/mutations/mysql.go
```diff
@@ -0,0 +1,42 @@
...
+	_ "github.com/go-sql-driver/mysql" // Set database engine.
+)
+
+var (
+	driverName = "mysql"
+
+	createStmt = []string{
+		`
+	CREATE TABLE IF NOT EXISTS Maps (
+		MapID   VARCHAR(32) NOT NULL,
```
gdbelvin: I think our MapIDs are now BIGINTs

impl/sql/mutations/mysql.go
```diff
@@ -0,0 +1,42 @@
...
+		PRIMARY KEY(MapID)
+	);`,
+		`
+	CREATE TABLE IF NOT EXISTS Mutations (
+		MapID    VARCHAR(32)   NOT NULL,
+		Sequence INTEGER       NOT NULL PRIMARY KEY AUTO_INCREMENT,
+                MIndex   VARBINARY(32) NOT NULL,
+		Mutation BLOB          NOT NULL,
+		UNIQUE(MapID, Sequence),
```
gdbelvin: It looks like this UNIQUE constraint is the only difference between MySQL and MySQLite. 
If so, consider merging the two, since the UNIQUE constraint is not actually doing anything. The PRIMARY KEY and NOT NULL won't allow duplicates anyhow.

cesarghali: Good catch. I removed `UNIQUE` from SQLite but forgot to remove it from MySQL. I will remove it as well.

The two queries can't be merged because MySQL and SQLite uses different auto increment keywords, i.e., AUTO_INCREMENT and AUTOINCREMENT, respectively. This is the main reason for the separation.

impl/sql/mutations/mysql.go
cesarghali: Good catch, I missed that.

cesarghali:
Thanks @gdbelvin for the review. The PR is updated and rebased. PTAL.

core/mutator/mutator.go
```diff
@@ -52,11 +51,23 @@ type Mutator interface {
...
-	Read(ctx context.Context, txn transaction.Txn, epoch int64, index []byte) ([]byte, error)
-	// Write saves the mutation in the database.
-	Write(ctx context.Context, txn transaction.Txn, epoch int64, index, mutation []byte) error
+	// ReadRange reads all mutations for a specific given mapID and sequence
+	// range. The range is identified by a starting sequence number and a
+	// count.
+	ReadRange(txn transaction.Txn, startSequence uint64, count int) ([]MutationInfo, error)
+	// Write saves the mutation in the database. Write returns the sequence
+	// number that is written.
+	Write(txn transaction.Txn, index, mutation []byte) (uint64, error)
```
gdbelvin: It looks like the index is being stored here and inside the mutation itself. 
Consider just passing the mutation itself as the argument to Write. Then the internal implementation can pull out the index if it needs to.  It also saves us a proto serialization too.

cesarghali: Good call. This is definitely a cleaner API.

impl/sql/mutations/mutations.go
```diff
@@ -77,35 +61,61 @@ func New(db *sql.DB, mapID int64) (mutator.Mutation, error) {
...
+		return nil, err
+	}
+	defer rows.Close()
 
-	var mutation []byte
-	err = readStmt.QueryRow(m.mapID, epoch, index).Scan(&mutation)
-	if err == sql.ErrNoRows {
-		return nil, nil
-	} else if err != nil {
+	var results []*tpb.SignedKV
```
gdbelvin: This could be slightly improved to avoid dynamic array-resizing at runtime:
`results := make([]*tpb.SignedKV, 0, count)`

impl/sql/mutations/mutations_test.go
```diff
@@ -35,110 +37,210 @@ func newDB(t testing.TB) *sql.DB {
...
-		index    []byte
-		mutation []byte
+		mutation    *tpb.SignedKV
+		outSequence uint64
 	}{
-		{1, []byte("index1"), []byte("mutation1")},
-		{2, []byte("index2"), []byte("mutation2")},
+		{
+			&tpb.SignedKV{
+				KeyValue: &tpb.KeyValue{
```
gdbelvin: nit: to save screen space, and slightly increase readability, each `KeyValue` could all be on one line.

PR #525, opened 2017-03-23
Use int64 for mapID

### LINKED ISSUE #508: Use int64 for MapID to match Trillian

gdbelvin:
To support migrating to Trillian Maps, use an int64 to indicate the backend mapID.

This also updates the way leaf hashes in the map are calculated, since the mapID is part of the leaf hash.

Closes #508


core/mutator/mutator.go
```diff
@@ -55,6 +55,7 @@ type Mutator interface {
 // Mutation reads and writes mutations to the database.
 type Mutation interface {
 	// Read reads all mutations for a specific given mapID, epoch, and index.
+	// TODO(cesarghali): Add mapID to this interface.
```
cesarghali: Why do we need `mapID` here? Mutation objects are initialized with a map ID. If this is related to supporting multiple maps per server, can you please add that to the comment?

core/tree/sparse/verifier/verifier.go
```diff
@@ -61,8 +64,9 @@ func (v *Verifier) VerifyProof(neighbors [][]byte, index, leaf []byte, root spar
 	}
 
 	// Verify that calculated and provided roots match.
-	if !bytes.Equal(calculatedRoot.Bytes(), root.Bytes()) {
-		return ErrInvalidProof
+	if got, want := calculatedRoot.Bytes(), root.Bytes(); !bytes.Equal(got, want) {
+		return fmt.Errorf("got \n%x, want \n%x", got, want)
+		//return ErrInvalidProof
```
cesarghali: Can you please uncomment this return and remove the currently returned error?

core/tree/sparse/verifier/verifier.go
```diff
@@ -38,13 +38,16 @@ var (
...
 	hasher sparse.TreeHasher
 }
 
 // New returns a new tree proofs verifier object.
-func New(mapID []byte, hasher sparse.TreeHasher) *Verifier {
-	return &Verifier{mapID, hasher}
+func New(mapID int64, hasher sparse.TreeHasher) *Verifier {
+	return &Verifier{
+		mapID:  mapID,
+		hasher: hasher,
```
cesarghali: +1 on using named literals. Thanks for the change.

core/tree/sparse/verifier/verifier_test.go
```diff
@@ -41,55 +43,116 @@ type Leaf struct {
...
+					[]byte("0"),
+					[][]byte{
+						dh("0a1db26319a01c8cef10b06c48d6dccce9ac5b951d5b91583a8346530d3d32a5"),
+					}},
+				{
+					dh(defaultIndex[0]),
+					[]byte("3"),
+					[][]byte{
+						dh("1cf3d1d90436e6ad75eb2ff97816ffcad6f024c6eceb7b86b2c507efdf3567d7"),
+						dh("22ed79bb41e8dbaf185691deee13645e6bd1d6faee90dc6d33b7a1342aa35555"),
```
cesarghali: I'm curious why we now have two neighbors while we only had one before? The test is passing now and it used to pass before. Also, the index looks the same to me.

scripts/prepare_client.sh
```diff
@@ -76,6 +76,7 @@ vrf:    \"${VRF}\"
 kt-key: \"${KTKEY}\"
 kt-sig: \"${SIGKEY}\"
 domain: \"${DOMAIN}\"
+mapid: 0
```
cesarghali: Can you also make the same change in the checked-in default `.keytransparency.yaml` file?

core/mutator/mutator.go
gdbelvin: This is to support multiple maps per server.  I'll add it to the comment.

core/tree/sparse/verifier/verifier.go
gdbelvin: Woops. Thanks. This was a debugging thing.

scripts/prepare_client.sh
gdbelvin: done

core/tree/sparse/verifier/verifier_test.go
gdbelvin: This caught me off guard too.  
We changed the way leaf hashes and empty hashes are calculated, but on the surface, that shouldn't change the index, should it?

gdbelvin: After looking at the indexes, though I think it's doing the right thing: 

`1000`
`1100`
`0010`

```
   / \
  2  / \
     1  0
```

1 and 0 should have two neighbors, which they now do.

cesarghali: You're right, it shouldn't. As you mentioned, not sure this deserves more investigation now since we might switch maps soon.

PR #526, opened 2017-03-23
Proto Cleanup

cesarghali:
We have few protobuf messages that are not used in `ctmap.go`. This PR removes them and renames `GetLeafResponse` to `LeafProof` in `keytransparency_v1_types.proto` since it is not related to CT map.


core/proto/keytransparency_v1_types/keytransparency_v1_types.proto
```diff
@@ -99,6 +99,15 @@ message SignedKV {
   bytes previous = 3;
 }
 
+// LeafProof for a verifiable map leaf.
+message LeafProof {
```
gdbelvin: This is great! 
FYI, we'll trying to move towards using Trillian.MapLeafInclusion here. 
Up to you whether you want to do this in this pull request or save it for later.

cesarghali: Thanks for the notification. I think I'll save this for another PR.

PR #528, opened 2017-03-27
Create a Sequenced table

gdbelvin:
Creates a place for storing generic items that have been sequenced.
Particuarly of the form: mapID, epoch.

Technically, a Trillian Log could serve the same purpose, but...
- You can't do an atomic transaction across an API call.
- We need to separate storing signed map heads from putting them in
  an append only log in order to match the trillian Map API.

This API supports the trillian Map function for storing Signed Map Heads.

### DIFF
--- a/impl/sql/sequenced/sequenced.go
+++ b/impl/sql/sequenced/sequenced.go
@@ -51,7 +51,7 @@ var (
 		`
 	CREATE TABLE IF NOT EXISTS Sequenced (
 		MapID   BIGINT      NOT NULL,
-		Epoch   BIGINT      NOT NULL,
+		Epoch   INTEGER     NOT NULL,
 		Data    BLOB(1024)  NOT NULL,
 		PRIMARY KEY(MapID, Epoch),
 		FOREIGN KEY(MapID) REFERENCES Maps(MapID) ON DELETE CASCADE
@@ -67,12 +67,16 @@ type Sequenced struct {
 }
 
 // New returns an object that can store sequenced items for multiple maps.
-func New(db *sql.DB) (sequenced.Sequenced, error) {
+// mapID will be the only allowed mapID.
+func New(db *sql.DB, mapID int64) (sequenced.Sequenced, error) {
 	if err := db.Ping(); err != nil {
 		return nil, fmt.Errorf("No DB connection: %v", err)
 	}
 
 	if err := create(db); err != nil {
+		return nil, err
+	}
+	if err := insertMapRow(db, mapID); err != nil {
 		return nil, err
 	}
 	return &Sequenced{
@@ -91,9 +95,9 @@ func create(db *sql.DB) error {
 	return nil
 }
 
-func (s *Sequenced) insertMapRow(txn transaction.Txn, mapID int64) error {
+func insertMapRow(db *sql.DB, mapID int64) error {
 	// Check if a map row does not exist for the same MapID.
-	countStmt, err := txn.Prepare(countMapRowExpr)
+	countStmt, err := db.Prepare(countMapRowExpr)
 	if err != nil {
 		return fmt.Errorf("insertMapRow(): %v", err)
 	}
@@ -107,7 +111,7 @@ func (s *Sequenced) insertMapRow(txn transaction.Txn, mapID int64) error {
 	}
 
 	// Insert a map row if it does not exist already.
-	insertStmt, err := txn.Prepare(insertMapRowExpr)
+	insertStmt, err := db.Prepare(insertMapRowExpr)
 	if err != nil {
 		return fmt.Errorf("insertMapRow(): %v", err)
 	}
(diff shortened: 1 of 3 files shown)


impl/sql/sequenced/sequenced.go
```diff
@@ -0,0 +1,182 @@
...
+	createStmt = []string{
+		`
+	CREATE TABLE IF NOT EXISTS Maps (
+		MapID   BIGINT NOT NULL,
+		PRIMARY KEY(MapID)
+	);`,
+		`
+	CREATE TABLE IF NOT EXISTS Sequenced (
+		MapID   BIGINT      NOT NULL,
+		Epoch   BIGINT      NOT NULL,
```
cesarghali: Please make this `INTEGER` to be consistent with the other database table definitions.

impl/sql/sequenced/sequenced.go
```diff
@@ -0,0 +1,182 @@
...
+	_, err = insertStmt.Exec(mapID)
+	if err != nil {
+		return fmt.Errorf("insertMapRow(): %v", err)
+	}
+	return nil
+}
+
+// Append adds an object to the append-only data structure.
+func (s *Sequenced) Write(txn transaction.Txn, mapID, epoch int64, obj interface{}) error {
+	if err := s.insertMapRow(txn, mapID); err != nil {
```
cesarghali: Can `insertMapRow` be called in `New` once instead of doing that for every write? Does this support multi-homed maps? If that's the case we need to change this in all other places while we're at it. I can do that after this PR is merged.

impl/sql/sequenced/sequenced_test.go
```diff
@@ -0,0 +1,141 @@
...
+			}
+			var readData []byte
+			err = a.Read(txn2, tc.mapID, tc.epoch, &readData)
+			if err != nil {
+				t.Errorf("Read(%v, %v): %v, want nil", tc.epoch, tc.data, err)
+			}
+			if err := txn2.Commit(); err != nil {
+				t.Errorf("txn2.Commit() failed: %v", err)
+			}
+			if got := readData; !bytes.Equal(got, tc.data) {
```
cesarghali: Readability: If got, want := readData, tc.data; ...

impl/sql/sequenced/sequenced_test.go
```diff
@@ -0,0 +1,141 @@
...
+		{0, 0, []byte("foo"), false},
+		{0, 1, []byte("foo"), true},
+	} {
+		txn, err := factory.NewDBTxn(context.Background())
+		if err != nil {
+			t.Errorf("factory.NewDBTxn() failed: %v", err)
+			continue
+		}
+		err = a.Write(txn, tc.mapID, tc.epoch, tc.data)
+		if got := err == nil; got != tc.want {
```
cesarghali: Readability: if got, want := err == nil, tc.want; ...

impl/sql/sequenced/sequenced_test.go
```diff
@@ -0,0 +1,141 @@
...
+			continue
+		}
+		epoch, err := a.Latest(txn2, tc.mapID, &obj)
+		if err != nil {
+			t.Errorf("Latest(): %v, want nil", err)
+		}
+		if err := txn2.Commit(); err != nil {
+			t.Errorf("txn.Commit() failed: %v", err)
+		}
+		if got := epoch; got != tc.want {
```
cesarghali: Readability: if got, want := epoch, tc.want; ...

impl/sql/sequenced/sequenced.go
gdbelvin: done

impl/sql/sequenced/sequenced_test.go
gdbelvin: done

impl/sql/sequenced/sequenced_test.go
gdbelvin: done

impl/sql/sequenced/sequenced_test.go
gdbelvin: done

gdbelvin:
Updated for readability, PTAL

impl/sql/sequenced/sequenced.go
```diff
@@ -91,9 +95,9 @@ func create(db *sql.DB) error {
 	return nil
 }
 
-func (s *Sequenced) insertMapRow(txn transaction.Txn, mapID int64) error {
+func insertMapRow(db *sql.DB, mapID int64) error {
```
cesarghali: You can make this a member of Sequenced object instead of passing db and mapID. This makes it consistent with the rest of the code. For example:
https://github.com/google/keytransparency/blob/master/impl/sql/mutations/mutations.go#L122

Issue #534, opened 2017-03-28
Multi-tenant support 


gdbelvin:
We would like to support key transparency for multiple domains using the same server.
This would provide a level of API matching to Trillian's multi-tenant support. 
This is a low-priority wish list item, not a strong requirement.

1. Incoming requests need to be mapped to a particular domain.
   1. Use `@domain` in the userID to do separation.
   1. Use a URL prefix for domain separation. 
1. A domain -> config database needs to be setup and configurable. 
1. Queries for objects configured from the config table need to be efficient.
   1. Trillian Log ids, urls, and pubkeys
   1. Trillian Map ids, urls, and pubkeys

gdbelvin:
Need mapping between 
domain -> mapID -> logID

gdbelvin:
Workplan:
- Config table
- Admin API for creating new mappings
- Update URLs to include ID

PR #536, opened 2017-03-29
Multi-tenant configuration support

gdbelvin:
Support multiple backend logs by storing their configuration
in an in-memory table.

In the future, the Add* commands of the Admin interface will perform
the needed setup such as inserting master rows in the various
sub-components.

Further in the future, this in-memory table will be backed by a database
rather than configured in-memory from the commandline.

#534

### DIFF
--- a/core/admin/static_test.go
+++ b/core/admin/static_test.go
@@ -1,0 +1,44 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package admin
+
+import (
+	"testing"
+
+	"github.com/google/keytransparency/integration/fake"
+)
+
+func TestStatic(t *testing.T) {
+	admin := NewStatic()
+	client := fake.NewFakeTrillianClient()
+
+	for _, tc := range []struct {
+		logID     int64
+		add, want bool
+	}{
+		{logID: 0, add: false, want: false},
+		{logID: 1, add: true, want: true},
+	} {
+		if tc.add {
+			if err := admin.AddLog(tc.logID, client); err != nil {
+				t.Errorf("AddLog(): %v, want nil", err)
+			}
+		}
+		_, err := admin.LogClient(tc.logID)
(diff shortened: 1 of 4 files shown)


integration/fake/fake_trilian_log.go
```diff
@@ -0,0 +1,70 @@
...
+func (f *FakeClient) AddLeaf(ctx context.Context, data []byte) error {
+	f.leaves = append(f.leaves, &trillian.LogLeaf{
+		LeafValue: data,
+	})
+	return nil
+}
+
+// GetByIndex returns the requested leaf.
+func (f *FakeClient) GetByIndex(ctx context.Context, index int64) (*trillian.LogLeaf, error) {
+	if got, want := index, int64(len(f.leaves)); got > want {
```
cesarghali: Since `index` is `int64`, this should check for negative values.

integration/fake/fake_trilian_log.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+	return f.leaves[index], nil
+}
+
+// ListByIndex returns the set of requested leaves.
+func (f *FakeClient) ListByIndex(ctx context.Context, start int64, count int64) ([]*trillian.LogLeaf, error) {
+	if got, want := start+count, int64(len(f.leaves)); got > want {
+		return nil, fmt.Errorf("Index out of range. Got %v, want %v", got, want)
+	}
+	return f.leaves[start : start+count], nil
```
cesarghali: Should check for negative start and end indexes.

core/admin/static_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+func TestStatic(t *testing.T) {
+	admin := NewStatic()
+	client := fake.NewFakeTrillianClient()
+
+	for _, tc := range []struct {
+		logID     int64
+		add, want bool
+	}{
+		{logID: 0, add: false, want: false},
```
gdbelvin: I think the names make it slightly easier to read in this case. 
Agreed that they're not strictly necessary in tests.

integration/fake/fake_trilian_log.go
```diff
@@ -0,0 +1,70 @@
...
+package fake
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/google/trillian"
+)
+
+// FakeClient implements trillian/client.LogVerifier
```
gdbelvin: done

integration/fake/fake_trilian_log.go
gdbelvin: Check my reasoning here:
- `len(f.leaves)` will be >= 0
- `int64(len(f.leaves))` upcasts length from `int` to `int64` to make sure that we're comparing to items of the same type.

Therefore a negative index will always fail this test.

gdbelvin: done

integration/fake/fake_trilian_log.go
gdbelvin: done

Issue #522, opened 2017-03-20
Consider adding a DB backed Queue


gdbelvin:
This would allow easier integration in other production environments. 
It would also eliminate our cross-domain transaction problems. 

On the other hand, this would tend to produce db tables with global, increasing counters which causes problems for scaling properly. 
- MySQL in master / slave setup can do this.

PR #546, opened 2017-04-11
Remove the Queue

cesarghali:
This PR removes the etcd queue which was responsible of processing mutations and creating epochs. The code in this PR works as follow:

* `keyserver` directly writes mutations in the database rather than enqueuing them so they can be written to the database.
* The signer's `StartSigning` function is the one that directly calls `CreateEpoch` instead of calling `AdvanceEpoch` in the queue which used to enqueue an advance epoch signal and eventually calls `CreateEpoch` when this signal is dequeued.

This should also stabilize the integration tests since everything is running in the same goroutine (no race can happen).

Next steps (each in a separate PR):

* Remove the cross-domain transaction object and convert all of its instances to database transactions.
* Add a master election functionality.

Parial #522.


cmd/keytransparency-server/frontend.go
```diff
@@ -188,7 +189,11 @@ func main() {
 	if err != nil {
 		log.Fatalf("Failed to create committer: %v", err)
 	}
-	queue := queue.New(context.Background(), etcdCli, *mapID, factory)
+	mutations, err := mutations.New(sqldb, *mapID)
+	if err != nil {
+		log.Fatalf("Failed to create mutations object: %v", err)
+	}
+	queue := queue.New(context.Background(), etcdCli, *mapID, factory, mutations)
```
gdbelvin: Does giving `queue` access to `mutations` invite scope creep? 
Would it be better to have `keyserver` write to `mutations` directly?

impl/etcd/queue/etcd.go
```diff
@@ -17,29 +17,46 @@ package queue
...
+	tpb "github.com/google/keytransparency/core/proto/keytransparency_v1_types"
 )
 
-// leaseTTL is in seconds.
-var leaseTTL = int64(30)
+var (
+	// leaseTTL is in seconds.
+	leaseTTL = int64(30)
+	// saveMutation will be overwritten in tests.
+	saveMutation = func(txn ctxn.Txn, mutations mutator.Mutation, mutation []byte) error {
```
gdbelvin: Do you think it would be better to kill the `Enqueue` function instead?

cesarghali: I wasn't thinking of doing this at this stage but it seems like a good idea. However, I think I'll devote a separate PR for this. That PR will include:

* Getting rid of `Enqueue` and moving storing mutations in the database to `keyserver.go`.
* Moving `AdvanceEpoch` and its functionality to `signer.go`.

This will basically remove our dependency on the queue so we can delete it completely.

cmd/keytransparency-server/frontend.go
cesarghali: Please see my comment below. The entire queue is going away soon!

gdbelvin:
LGTM

cesarghali:
Please take another look at this PR, specifically the second commit. I also changed the PR's name and description.

core/keyserver/keyserver.go
```diff
@@ -258,9 +254,21 @@ func (s *Server) UpdateEntry(ctx context.Context, in *tpb.UpdateEntryRequest) (*
...
-		return nil, grpc.Errorf(codes.Internal, "Write error")
+	// Save mutation to the database.
+	txn, err := s.factory.NewDBTxn(ctx)
+	if err != nil {
+		return nil, grpc.Errorf(codes.Internal, "Cannot create transaction")
+	}
+	if _, err := s.mutations.Write(txn, in.GetEntryUpdate().GetUpdate()); err != nil {
+		log.Printf("mutations.Write failed: %v", err)
+		if err := txn.Rollback(); err != nil {
+			log.Printf("Cannot rollback the transaction: %v", err)
```
gdbelvin: Include previous error and rollback error

core/signer/signer.go
```diff
@@ -84,17 +84,27 @@ func (s *Signer) FakeTime() {
...
 }
 
-// StartSigning inserts epoch advancement signals into the queue.
-func (s *Signer) StartSigning(interval time.Duration) {
+// StartSigning advance epochs once per interval.
+func (s *Signer) StartSigning(ctx context.Context, interval time.Duration) {
 	for range time.NewTicker(interval).C {
-		if err := s.queue.AdvanceEpoch(); err != nil {
-			log.Fatalf("Advance epoch failed: %v", err)
+		txn, err := s.factory.NewDBTxn(ctx)
```
gdbelvin: We might help ourselves if we put this internal bit in it's own function

core/signer/signer.go
```diff
@@ -115,20 +125,42 @@ func (s *Signer) ProcessMutation(ctx context.Context, txn transaction.Txn, index
...
+		startSequence = smh.GetMapHead().MaxSequenceNumber + 1
+	} else if err != sql.ErrNoRows {
+		return 0, fmt.Errorf("Latest err: %v", err)
+	}
+	maxSequence, mutations, err := s.mutations.ReadRange(txn, startSequence, math.MaxInt32)
+	if err != nil {
+		return 0, fmt.Errorf("ReadRange err: %v", err)
+	}
+	for _, mutation := range mutations {
+		mData, err := proto.Marshal(mutation)
```
gdbelvin: Add TODO or issue to convert mutation API from `[]byte to interface{}`

impl/sql/mutations/mutations.go
```diff
@@ -62,35 +62,41 @@ func New(db *sql.DB, mapID int64) (mutator.Mutation, error) {
...
+	maxSequence := uint64(0)
 	for rows.Next() {
+		var sequence uint64
 		var mData []byte
-		if err := rows.Scan(&mData); err != nil {
-			return nil, err
+		if err := rows.Scan(&sequence, &mData); err != nil {
+			return 0, nil, err
+		}
+		if maxSequence < sequence {
```
gdbelvin: It might read a bit clearer to swap the two arguments

integration/client_test.go
```diff
@@ -202,27 +209,35 @@ func TestUpdateValidation(t *testing.T) {
 			t.Fatalf("Update(%v): %v != %v, want %v", tc.userID, err, want, tc.want)
 		}
 		if tc.want {
-			if err := env.Queue.AdvanceEpoch(); err != nil {
-				t.Errorf("AdvanceEpoch: %v", err)
+			txn, err := env.Factory.NewDBTxn(bctx)
```
gdbelvin: use the function I referenced pulling out above.  This seems like a lot of boiler plate code

core/keyserver/keyserver.go
cesarghali: Previous error is already logged two lines above. The same behavior is used throughout keyserver.go

core/signer/signer.go
cesarghali: Mutation APIs take and return `SignedKV` and not `[]byte`, do we want to use `interface{}` instead?

cesarghali:
This PR is rebased and ready for another pass. All comments are addressed or responded to.
PTAL, specifically at the second commit "Add ReadAll API to mutator.Mutation". This fix the problem of memory drainage in the integration tests.

core/keyserver/keyserver.go
gdbelvin: cool

core/signer/signer.go
```diff
@@ -115,20 +114,59 @@ func (s *Signer) ProcessMutation(ctx context.Context, txn transaction.Txn, index
...
+	return nil
+}
+
+// processMutations reads mutations from the database and adds them to the tree.
+// processMutations returns the maximum mutation sequence number processed.
+func (s *Signer) processMutations(ctx context.Context, txn transaction.Txn) (uint64, error) {
+	startSequence := uint64(0)
+	smh := new(ctmap.SignedMapHead)
+	if _, _, err := s.sths.Latest(ctx, smh); err == nil {
+		startSequence = smh.GetMapHead().MaxSequenceNumber + 1
```
gdbelvin: Just thinking about what this would mean for timestamps... 
- Timestamps use the time uncertainty at a particular moment to figure out what +1 means. 
- I think there are APIs for making open / closed interval requests, eg. (  ], [ ], [ ), ( )

Consider defining the `ReadAll` API to be a ( ] interval?

cesarghali: While this works for sequence numbers, it also works for timestamp. As long as we provide a starting point that is greater than `MaxSequenceNumber` but not greater than the next sequence number after `MaxSequenceNumber` then the SQL query will cover it. Note that the SQL query has `Sequence >= ?` as a condition. This with the `+1` is basically a ( ] interval.

The SQL query is already collecting a ( ] interval.

cesarghali: By the way, the timestamp sequence number is just an integer, so regardless of the time granularity stored in that integer, doing `+1` will skip `MaxSequenceNumber` and include the next sequence number in the returned result.

PR #547, opened 2017-04-12
Use time.Duration for sequencing period

gdbelvin:


cmd/keytransparency-signer/backend.go
```diff
@@ -41,11 +41,11 @@ import (
 var (
 	serverDBPath  = flag.String("db", "db", "Database connection string")
 	etcdEndpoints = flag.String("etcd", "", "Comma delimited list of etcd endpoints")
-	epochDuration = flag.Uint("period", 60, "Seconds between epoch creation")
 	domain        = flag.String("domain", "example.com", "Distinguished name for this key server")
 	mapID         = flag.Int64("mapid", 0, "ID for backend map")
 	mapLogURL     = flag.String("maplog", "", "URL of CT server for Signed Map Heads")
 	signingKey    = flag.String("key", "", "Path to private key PEM for STH signing")
+	epochDuration = flag.Duration("period", time.Second*60, "Time between epoch creation")
```
cesarghali: Please add the unit to the description since it's always second.

gdbelvin: I believe that units are now specified on the command line. eg. "5s"

PR #549, opened 2017-04-14
Create Travis presubmit make rule without coverage

cesarghali:
Travis is failing when running coverage. Coverage is not adding any benefits in Travis so this PR removes it. We still have coverage when running presubmit locally.


gdbelvin:
I think you also need to update .travis

PR #550, opened 2017-04-14
Create Trillian Interfaces

gdbelvin:
- Appender Local and Remote interfaces
- Create Trillian Client
- Fake
- Create LogClient helper

### DIFF
--- a/core/appender/trillian_test.go
+++ b/core/appender/trillian_test.go
@@ -23,6 +23,7 @@ import (
 )
 
 func TestLatest(t *testing.T) {
+	ctx := context.Background()
 	fakeLog := fake.NewFakeTrillianClient()
 	admin := admin.NewStatic()
 	if err := admin.AddLog(0, fakeLog); err != nil {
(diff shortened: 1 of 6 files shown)


core/appender/trillian_test.go
```diff
@@ -0,0 +1,56 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
```
cesarghali: Can you please add a test for the `Read` API?

gdbelvin: done

core/appender/trillian_test.go
```diff
@@ -0,0 +1,61 @@
...
+		if err := a.Write(ctx, tc.logID, tc.epoch, tc.data); err != nil {
+			t.Errorf("Write(%v, %v): %v, want nil", tc.epoch, tc.data, err)
+		}
+
+		var obj []byte
+		if err := a.Read(ctx, tc.logID, tc.epoch, &obj); err != nil {
+			t.Errorf("Read(%v): %v, want nil", tc.epoch, err)
+		}
+
+		epoch, err := a.Latest(context.Background(), tc.logID, &obj)
```
cesarghali: nit: please use `ctx` instead of `context.Background()`. Also `ctx` can be defined outside the loop, e.g., at the beginning of the test.

PR #552, opened 2017-04-14
Use appender.Remote in place of appender.Appender

gdbelvin:
Transition off of Certificate Transparency onto Trillian Logs.

- Trillian does not support Signed Certificate Timestamps.
- Removes CT appender interface.
- Configures servers to explicitly specify the logID for the backend.

The next PR will update the client and server setup scripts


core/client/kt/verify.go
```diff
@@ -111,15 +109,6 @@ func (v *Verifier) VerifyGetEntryResponse(ctx context.Context, userID string, in
...
-	sct := new(ct.SignedCertificateTimestamp)
-	if _, err := tls.Unmarshal(in.SmhSct, sct); err != nil {
-		return err
-	}
-	if err := v.log.VerifySCT(ctx, in.GetSmh(), sct); err != nil {
-		Vlog.Printf("✗ Signed Map Head CT inclusion proof verification failed.")
-		return err
-	}
-	Vlog.Printf("✓ Signed Map Head CT inclusion proof verified.")
+	// TODO: Verify inclusion proof.
```
cesarghali: This reduces the security of the system, do you know when we're planning on adding this verification step?

core/keyserver/keyserver.go
```diff
@@ -77,9 +79,9 @@ func New(committer commitments.Committer, queue queue.Queuer, tree tree.Sparse,
 func (s *Server) GetEntry(ctx context.Context, in *tpb.GetEntryRequest) (*tpb.GetEntryResponse, error) {
 
 	var smh ctmap.SignedMapHead
-	epoch, _, err := s.appender.Latest(ctx, &smh)
+	epoch, err := s.sths.Latest(ctx, s.logID, &smh)
 	if err != nil {
-		log.Printf("Cannot get SMH: %v", err)
+		log.Printf("sth.Latest(%v): %v", s.logID, err)
```
cesarghali: s/sth/sths

cmd/keytransparency-server/frontend.go
```diff
@@ -193,15 +199,21 @@ func main() {
...
 		log.Fatalf("Failed to create SQL history: %v", err)
 	}
-	sths, err := appender.New(context.Background(), sqldb, *mapID, *mapLogURL, nil)
+
+	tlog, err := config.LogClient(*logID, *logURL, *logPubKey)
 	if err != nil {
-		log.Fatalf("Failed to create appender: %v", err)
+		log.Fatalf("LogClient(%v, %v, %v): %v", *logID, *logURL, *logPubKey, err)
+	}
+	admin := admin.NewStatic()
```
cesarghali: The name of this variable is a bit confusing, how about `static` instead, which is what `NewStatic` returns?

cmd/keytransparency-signer/backend.go
```diff
@@ -99,10 +105,15 @@ func main() {
 		log.Fatalf("Failed to create SQL history: %v", err)
 	}
 	mutator := entry.New()
-	sths, err := appender.New(context.Background(), sqldb, *mapID, *mapLogURL, nil)
+	tlog, err := config.LogClient(*logID, *logURL, *logPubKey)
 	if err != nil {
-		log.Fatalf("Failed to create STH appender: %v", err)
+		log.Fatalf("LogClient(%v, %v, %v): %v", *logID, *logURL, *logPubKey, err)
+	}
+	admin := admin.NewStatic()
```
cesarghali: Same here.

core/keyserver/keyserver_test.go
```diff
@@ -116,7 +121,11 @@ func addProfiles(count int, c *fakeCommitter, st *fakeSparseHist, a *fakeAppende
 		committed := &tpb.Committed{Data: pData}
 		c.M[string(commitment)] = committed
 		st.M[int64(i)] = commitment
-		a.CurrentEpoch = int64(i)
+
+		smh := new(ctmap.SignedMapHead)
+		if err := sths.Write(context.Background(), 0, int64(i), smh); err != nil {
```
cesarghali: Please create the context in the test function and pass it down to this one.

core/keyserver/keyserver_test.go
```diff
@@ -56,9 +62,14 @@ func TestListEntryHistory(t *testing.T) {
...
 		st := &fakeSparseHist{make(map[int64][]byte)}
-		a := &fakeAppender{0, 0}
-		srv := New(c, fakeQueue{}, st, a, fakePrivateKey{}, fakeMutator{}, authentication.NewFake(), fakeFactory{})
-		if err := addProfiles(profileCount, c, st, a); err != nil {
+		fakeLog := fake.NewFakeTrillianClient()
+		admin := admin.NewStatic()
+		if err := admin.AddLog(logID, fakeLog); err != nil {
+			t.Fatalf("failed to add log to admin: %v", err)
+		}
+		sths := appender.NewTrillian(admin)
```
cesarghali: Can you please mock `sths` similar to what we did with `fakeAppender`? I don't think testing key server depends on having a real `sths` object. Therefore, a fake `sths` does not need to do anything other than implementing the interface.

Same thing if this applies to other tests in this file.

core/client/kt/verify.go
gdbelvin: I'll add this in the next PR.  To add it here would require expanding this PR to also include all the changes to the client.  For ease of review, I've broken this up.

core/keyserver/keyserver.go
gdbelvin: done

cmd/keytransparency-server/frontend.go
gdbelvin: done

cmd/keytransparency-signer/backend.go

core/keyserver/keyserver_test.go
gdbelvin: Let's chat about this on GVC. 

This test is using a mock - `fakeLog`. Perhaps there's a better way to setup multi-tennant support - or delay multi tenant support till later.

core/keyserver/keyserver_test.go
gdbelvin: done

gdbelvin:
Rebased with comments addressed, PTAL

core/keyserver/keyserver_test.go
cesarghali: That's right. However, the `fakeLog` actually has some functionality which is not needed to test this package, but is needed for integration test. I was thinking the mocks will immediately return `nil`. We can talk about this more, though.

PR #554, opened 2017-04-14
Change mutator.Mutation Read APIs to Return ( ] Range

### LINKED ISSUE #553: Change mutator.Mutation Read APIs to Return ( ] Range.
Currently both `mutator.Mutation` read APIs take the starting sequence number as an input. The signer reads the maximum sequence number in the latest map head and adds one to it before calling `ReadAll` API. This might be problematic with timestamp-based sequence numbers.

Both read APIs (`ReadRange` and `ReadAll`) should return a result in an interval of the form ( ].

cesarghali:
Closes #553.

See the bug for more context.

### DIFF
--- a/core/mutator/mutator.go
+++ b/core/mutator/mutator.go
@@ -58,10 +58,12 @@ type Mutator interface {
 type Mutation interface {
 	// ReadRange reads all mutations for a specific given mapID and sequence
 	// range. The range is identified by a starting sequence number and a
-	// count. ReadRange also returns the maximum sequence number read.
+	// count. Note that startSequence is not included in the result.
+	// ReadRange also returns the maximum sequence number read.
 	ReadRange(txn transaction.Txn, startSequence uint64, count int) (uint64, []*tpb.SignedKV, error)
 	// ReadAll reads all mutations starting from the given sequence number.
-	// ReadAll also returns the maximum sequence number read.
+	// Note that startSequence is not included in the result. ReadAll also
+	// returns the maximum sequence number read.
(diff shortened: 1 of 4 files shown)


gdbelvin:
`ReadAll` is used when generating a new SignedMapHead. If I recall correctly`ReadRange` will be used to enumerate mutations in an epoch? Would it make more sense if `ReadRange` was `[, )` (per the API style guide) while `ReadAll` was `(, ]` ?

cesarghali:
Yes, this makes sense, but I thought making the two APIs similar will avoid confusion about which one does what?

One other thing, theoretically `ReadRange` can return a list of mutation across epoch boundaries. Mutations APIs don't understand epochs. If such boundaries are needed they have to be set by the APIs caller.

gdbelvin:
+1 to avoiding API confusion. 

Agreed that `ReadRange` shouldn't know about epochs. However, is the API simple enough for the caller to use without error or confusion? Would a `ReadRange(from, to)` style be easier?

cesarghali:
I agree. Actually the `to` field semantics will be different depending on the type of sequence numbers used.

* For monotonically incremental sequence numbers, `to` could be the maximum sequence number, which is easily calculated by the caller given a start sequence number and a window size.
* For timestamp-based sequence numbers, `to` should be the maximum number of returned elements, basically a count. This is because the maximum sequence number in this case cannot be known by the caller.

I can do this API change in a separate PR.

gdbelvin:
It sounds like ReadRange is for the mutations API which hasn't been built yet. 
Let's hold off on changing ReadRange until it's driven by meeting those API needs.

cesarghali:
Sounds good. I will change the `ReadRange` API on demand later when we implement the caller.

PR #555, opened 2017-04-14
Remove All References to etcd

cesarghali:
This PR removes all references to `etcd` from the binaries and configuration files. The only reference left is in the cross-domain transaction code. It will be removed along with this transaction object soon.

Partial #522.


gdbelvin:
Do you think we should leave this to make adding leader election easier?

cesarghali:
I think it's better to have a clean code now and then we can add whatever we need for master election. This is better than leaving everything now and then try to trim it and end up leaving behind things we don't need.

PR #559, opened 2017-04-21
Add coverage testing to travis

gdbelvin:
Uploads coverage analysis to https://codecov.io/gh/google/keytransparency

Example:
https://github.com/codecov/example-go

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ tools:
 
 presubmit: coverage fmt tools
 
-travis-presubmit: test fmt tools
+travis-presubmit: fmt tools
 
 proto:
 	go generate ./...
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -53,7 +53,7 @@ tools:
 
 presubmit: coverage fmt tools
 
-travis-presubmit: test fmt tools
```
cesarghali: Can we please keep running tests as part of travis-presubmit? This is the only way we can know that a pull request passes all tests. Without it, we might merge one that breaks the master branch.

cesarghali: I just noticed that tests are running as part of `coverage.sh` script.

PR #562, opened 2017-04-24
Replace proto/ctmap with trillian

gdbelvin:
ctmap was always a placeholder for proper trillian map protos. 
Now that trillian map protos are available, switch to using them.


cmd/keytransparency-client/cmd/history.go
```diff
@@ -21,12 +21,10 @@ import (
 	"text/tabwriter"
 	"time"
 
-	"github.com/golang/protobuf/ptypes"
+	"github.com/google/trillian"
```
cesarghali: Since this is a proto import, can you please move it to the last import group?

cmd/keytransparency-client/grpcc/grpc_client.go
```diff
@@ -34,11 +34,11 @@ import (
 	"github.com/google/keytransparency/core/vrf"
 
 	"github.com/golang/protobuf/proto"
+	"github.com/google/trillian"
```
cesarghali: Same here.

core/signer/signer.go
```diff
@@ -27,11 +27,8 @@ import (
 	"github.com/google/keytransparency/core/transaction"
 	"github.com/google/keytransparency/core/tree"
 
-	"github.com/golang/protobuf/ptypes"
+	"github.com/google/trillian"
```
cesarghali: Same here.

integration/client_test.go
```diff
@@ -26,9 +26,9 @@ import (
 	"github.com/google/keytransparency/core/crypto/signatures"
 	"github.com/google/keytransparency/core/crypto/signatures/factory"
 
+	"github.com/google/trillian"
```
cesarghali: Same here.

cmd/keytransparency-client/cmd/history.go
gdbelvin: done

core/signer/signer.go
gdbelvin: done

integration/client_test.go
gdbelvin: done

Issue #486, opened 2017-01-31
Use Trillian Map


gdbelvin:
Step 1: Have Key Transparency's internals fulfill the TrillianMapClient API.

PR #564, opened 2017-04-26
Use mapserver in signer

gdbelvin:
Use an internal implementation of the trillian map server. 
In the keyserver:
- Use GetLeaves to retrieve leaf values and inclusion proofs.
- Use GetSignedMapRoot to retrieve the current epoch.  (Replaces sths object)

In the signer:
- Collect all mutations that have occurred since the last epoch.
- GetLeaves for the respective indexes
- Verify and Apply the mutations.
- Call SetLeaves to create a new epoch with the new values. 

TODO:
- Pass an expected epoch field to make sure that the read/write cycle to the mapserver is atomic. google/trillian#559
- Use mastership election to make sure that only one signer is active at any point. #565

Contributes to #486

### DIFF
--- a/cmd/keytransparency-server/frontend.go
+++ b/cmd/keytransparency-server/frontend.go
@@ -17,6 +17,7 @@ package main
 import (
 	"database/sql"
 	"flag"
+	"fmt"
 	"io/ioutil"
 	"log"
 	"net/http"
@@ -28,6 +29,7 @@ import (
 	"github.com/google/keytransparency/core/keyserver"
 	"github.com/google/keytransparency/core/mapserver"
 	"github.com/google/keytransparency/core/mutator/entry"
+	ctxn "github.com/google/keytransparency/core/transaction"
 	"github.com/google/keytransparency/impl/google/authentication"
 	"github.com/google/keytransparency/impl/sql/commitments"
 	"github.com/google/keytransparency/impl/sql/engine"
@@ -35,6 +37,7 @@ import (
 	"github.com/google/keytransparency/impl/sql/sequenced"
 	"github.com/google/keytransparency/impl/sql/sqlhist"
 	"github.com/google/keytransparency/impl/transaction"
+	"github.com/google/trillian"
 
 	"github.com/golang/protobuf/jsonpb"
 	"github.com/golang/protobuf/proto"
@@ -156,6 +159,18 @@ func jsonLogger(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo
 	return resp, err
 }
 
+func newReadonlyMapServer(ctx context.Context, mapID int64, sqldb *sql.DB, factory ctxn.Factory) (trillian.TrillianMapClient, error) {
+	tree, err := sqlhist.New(ctx, mapID, factory)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to create SQL history: %v", err)
+	}
+	sths, err := sequenced.New(sqldb, mapID)
+	if err != nil {
+		return nil, fmt.Errorf("sequenced.New(%v): %v", mapID, err)
+	}
+	return mapserver.NewReadonly(mapID, tree, factory, sths), nil
+}
+
 func main() {
 	log.SetFlags(log.LstdFlags | log.Lshortfile)
 	flag.Parse()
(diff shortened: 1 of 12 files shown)


core/mapserver/readonly.go
```diff
@@ -0,0 +1,140 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
```
cesarghali: Can you please add tests for this?

core/signer/signer.go
```diff
@@ -94,110 +71,129 @@ func (s *Signer) StartSigning(ctx context.Context, interval time.Duration) {
...
 
-	smr := &trillian.SignedMapRoot{
-		MapId:          0,
-		TimestampNanos: s.clock.Now().UnixNano(),
-		RootHash:       root,
-		MapRevision:    epoch,
-		Metadata: &trillian.MapperMetadata{
-			HighestFullyCompletedSeq: maxSequence,
-		},
+	// TODO: verify inclusion proofs?
```
cesarghali: Why do we need to verify inclusion proofs here? This will add a lot of computation in this critical path, right?

cmd/keytransparency-signer/backend.go
```diff
@@ -17,23 +17,28 @@ package main
...
 	"github.com/google/keytransparency/impl/sql/engine"
 	"github.com/google/keytransparency/impl/sql/mutations"
+	"github.com/google/keytransparency/impl/sql/sequenced"
 	"github.com/google/keytransparency/impl/sql/sqlhist"
 	"github.com/google/keytransparency/impl/transaction"
 
+	"github.com/google/trillian"
+	"github.com/google/trillian/crypto/keys"
+	"github.com/google/trillian/util"
+
```

cmd/keytransparency-server/frontend.go
```diff
@@ -182,25 +183,31 @@ func main() {
...
-		log.Fatalf("LogClient(%v, %v, %v): %v", *logID, *logURL, *logPubKey, err)
-	}
-	static := admin.NewStatic()
-	if err := static.AddLog(*mapID, tlog); err != nil {
-		log.Fatalf("static.AddLog(%v): %v", *mapID, err)
-	}
-	sths := appender.NewTrillian(static)
-	vrfPriv := openVRFKey()
-	mutator := entry.New()
+	sths, err := sequenced.New(sqldb, *mapID)
```
gdbelvin: done

cmd/keytransparency-server/frontend.go
```diff
@@ -182,25 +183,31 @@ func main() {
...
+	mapsvr := mapserver.NewReadonly(*mapID, tree, factory, sths)
```
gdbelvin: done

core/client/kt/verify.go
```diff
@@ -108,9 +109,9 @@ func (v *Verifier) VerifyGetEntryResponse(ctx context.Context, userID string, in
 	}
 	Vlog.Printf("✓ Sparse tree proof verified.")
 
-	sig := in.GetSmr().Signature
-	in.GetSmr().Signature = nil // Remove the signature from the object to be verified.
-	if err := v.sig.Verify(in.GetSmr(), sig); err != nil {
+	smr := *in.GetSmr()
```
gdbelvin: I'll add clear comments on the caller side. 
@quannguyentm Would love your input on this API design.

core/mapserver/readonly.go
gdbelvin: Added tests to mapserver_test.go

core/mapserver/readonly.go
```diff
@@ -0,0 +1,140 @@
...
+	return &readonly{
+		mapID:   mapID,
+		tree:    tree,
+		factory: factory,
+		sths:    sths,
+	}
+}
+
+func (r *readonly) GetLeaves(ctx context.Context, in *trillian.GetMapLeavesRequest, opts ...grpc.CallOption) (resp *trillian.GetMapLeavesResponse, retErr error) {
+	if got, want := in.MapId, r.mapID; got != want {
```
gdbelvin: done

core/mapserver/readonly.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	return &trillian.GetMapLeavesResponse{
+		MapLeafInclusion: inclusions,
+		MapRoot:          &root,
+	}, nil
+}
+
+func (r *readonly) SetLeaves(ctx context.Context, in *trillian.SetMapLeavesRequest, opts ...grpc.CallOption) (*trillian.SetMapLeavesResponse, error) {
+	panic("not implemented")
```
gdbelvin: done

core/signer/signer.go
gdbelvin: I'm not sure we need to verify inclusion proofs here, hence the question mark. 
If the map server was being run by a different entity, we'd want to verify the inclusion proofs.  However, in our case, we expect the keyserver and the mapserver to be run by the same entity.

gdbelvin:
Addressed comments. PTAL

core/signer/signer.go
cesarghali: Good point. Maybe at some point we can this functionality to be enabled with a flag to the key server binary.

PR #568, opened 2017-04-26
Ensure VRF uniqueness

### LINKED ISSUE #567: VRF uniqueness is violated
Thanks to Leonid Reyzin for pointing this out" can we say "Thanks to Sharon Goldberg and Leonid Reyzin for finding the error. 

An attacker can choose a different `r`s when computing [`s`](https://github.com/google/keytransparency/blob/master/core/vrf/p256/p256.go#L127). Rather than 
`s = H2(m, [r]G, [r]H)`, the attacker computes 
`s = H2(m, [a]G, [b]H)` and claims that the VRF output is 
`[(b-a)/s + k] H` rather than` [k]H` and `t = a-sk`.

This checks out because the verifier checks that
 `s == H2(m, [t]G + [s]([k]G), [t]H + [s]VRF)`.
`s == H2(m, [a]G, [b]H)`
  Because
`[t]G+[s]([k]G) = [a]G`. 
`[t]H + [s]VRF = [a-sk]H + [b-a+sk]H = [b]H`.
 The attack succeeds at producing a non-unique value for VRF.

http://eprint.iacr.org/2012/577.pdf Page 4:
> Goldberg and Reyzin (March 2017) discovered that if one does not hash the unique identifier when computing the challenge of the proof system, the uniqueness of the VRF from DDH assumption is violated. Therefore, when using the VRF, it is important to hash the unique identifier as well.

gdbelvin:
VRF uniqueness could be violated because the VRF output itself was not
included in the the computation of `s`, the commitment to the VRF and
zero knowledge proof. Fixes #567. 

Other changes also included in this PR: 
- H2(.., m, ..) -> H2(.., H1(m),..) for better alignment with the NSEC5 VRF (sec 3.5).  This change was suggested by Leonid Reyzin and approved by @jcb82 (the CONIKS paper has been updated).
- Variable name notation has been updated for better code clarity. 
- Add test vectors #613 

References:
- [NSEC5 Paper](https://eprint.iacr.org/2017/099.pdf)
- [CONIKS Papser](https://eprint.iacr.org/2014/1004.pdf)

### DIFF
--- a/core/crypto/vrf/p256/p256.go
+++ b/core/crypto/vrf/p256/p256.go
@@ -118,23 +118,23 @@ func (k PrivateKey) Evaluate(m []byte) (index [32]byte, proof []byte) {
 	ri := new(big.Int).SetBytes(r)
 
 	// H = H1(m)
-	hx, hy := H1(m)
+	Hx, Hy := H1(m)
 
 	// VRF_k(m) = [k]H
-	vrfx, vrfy := params.ScalarMult(hx, hy, k.D.Bytes())
-	vrf := elliptic.Marshal(curve, vrfx, vrfy) // 65 bytes.
+	kHx, kHy := params.ScalarMult(Hx, Hy, k.D.Bytes())
+	vrf := elliptic.Marshal(curve, kHx, kHy) // 65 bytes.
 
 	// G is the base point
-	// s = H2(G, h, [k]G, [k]H, [r]G, [r]H)
-	gRx, gRy := params.ScalarBaseMult(r)
-	hRx, hRy := params.ScalarMult(hx, hy, r)
+	// s = H2(G, H, [k]G, VRF, [r]G, [r]H)
+	rGx, rGy := params.ScalarBaseMult(r)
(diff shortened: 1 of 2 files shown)


core/crypto/vrf/p256/p256.go
```diff
@@ -123,14 +123,18 @@ func (k PrivateKey) Evaluate(m []byte) (vrf, proof []byte) {
 	// H = H1(m)
 	hx, hy := H1(m)
 
+	// VRF_k(m) = [k]H
+	vrfx, vrfy := params.ScalarMult(hx, hy, k.D.Bytes())
+
 	// G is the base point
-	// s = H2(m, [r]G, [r]H)
+	// s = H2(m, [r]G, [r]H, [k]H)
```
jcb82: Order is not following either the CONIKs paper or Franklin paper (which list [k]H earlier)

Also, why not hash in G (base point) and the public key kG? As is done in both papers

gdbelvin: Added the base point and the public key. 
Updated the order to match the CONIKS paper.

PR #572, opened 2017-04-28
Include log proofs in GetEntryResponse

gdbelvin:
Remove the need for clients to connect separately to the trillian log by having the keyserver proxy all the required log requests. 

Front end:
- Fetches and returns the appropriate log proofs from the trillian log.

Client:
- No longer needs logID or logURL.

KT Client:
- Requests now include the TreeSize of the currently trusted log root.


core/client/kt/verify.go
```diff
@@ -120,13 +121,23 @@ func (v *Verifier) VerifyGetEntryResponse(ctx context.Context, userID string, in
 	}
 	Vlog.Printf("✓ Signed Map Head signature verified.")
 
+	// Update the log root.
+	// TODO(gdbelvin): Gossip root.
+	if err := v.log.UpdateRoot(in.LogRoot, in.LogConsistency); err != nil {
```
cesarghali: This call to `UpdateRoot` seems to me a bit misplaced in a function that verifies things. From its name, `VerifyGetEntryResponse` gives the impression that it's a passive function. It might be better for `VerifyGetEntryResponse` callers to call `UpdateRoot`. What do you think?

integration/testutil.go
```diff
@@ -220,7 +221,7 @@ func NewEnv(t *testing.T) *Env {
 		t.Fatalf("Dial(%v) = %v", addr, err)
 	}
 	cli := pb.NewKeyTransparencyServiceClient(cc)
-	client := grpcc.New(mapID, cli, vrfPub, verifier, fake.NewFakeTrillianClient())
+	client := grpcc.New(mapID, cli, vrfPub, verifier, fake.NewFakeTrillianLogVerifier())
```
cesarghali: Can `tlog` be used here?

cmd/keytransparency-client/cmd/root.go
```diff
@@ -247,19 +248,19 @@ func GetClient(clientSecretFile string) (*grpcc.Client, error) {
...
 	if err != nil {
 		return nil, fmt.Errorf("Error Dialing %v: %v", ktURL, err)
 	}
 
-	log, err := config.LogClient(logID, logURL, logPEM)
+	// Log verifier.
+	logPubKey, err := keys.NewFromPublicPEMFile(logPEM)
 	if err != nil {
-		return nil, fmt.Errorf("config.LogClient(): %v", err)
+		log.Fatalf("Failed to open public key %v: %v", logPubKey, err)
```
gdbelvin: woops, sorry about that. fixed.

cmd/keytransparency-client/grpcc/grpc_client.go
```diff
@@ -83,14 +83,18 @@ type Client struct {
...
+	log        client.LogVerifier
 	mutator    mutator.Mutator
 	RetryCount int
 	RetryDelay time.Duration
 }
 
 // New creates a new client.
-func New(mapID int64, client spb.KeyTransparencyServiceClient, vrf vrf.PublicKey, verifier crypto.PublicKey, log client.VerifyingLogClient) *Client {
+func New(mapID int64,
+	client spb.KeyTransparencyServiceClient,
```
gdbelvin: Good question. I started doing this when the line got unwieldy long.
The Go style guide doesn't have a strict recommendation on when to wrap lines: 
https://github.com/golang/go/wiki/CodeReviewComments#line-length 

Perhaps the inconsistency is ok, or I could go back and try to group the parameters?

core/client/kt/requests.go
```diff
@@ -25,12 +25,24 @@ import (
...
+// Root returns the currently trusted SignedLogRoot.
+func (v *Verifier) Root() trillian.SignedLogRoot {
+	return v.log.Root()
+}
+
 // CreateUpdateEntryRequest creates UpdateEntryRequest given GetEntryResponse,
 // user ID and a profile.
-func CreateUpdateEntryRequest(getResp *tpb.GetEntryResponse, vrf vrf.PublicKey, userID string, profile *tpb.Profile, signers []signatures.Signer, authorizedKeys []*tpb.PublicKey) (*tpb.UpdateEntryRequest, error) {
+func (v *Verifier) CreateUpdateEntryRequest(
+	getResp *tpb.GetEntryResponse,
```
gdbelvin: see below

integration/testutil.go
gdbelvin: Not any more.  Previously, clients needed a `TrillianClient` to connect to the log server itself. Now clients just need to verify the output of Trillian.  Meanwhile, the keyserver itself needs the `TrillianClient` in order to proxy those requests now.

core/client/kt/verify.go
gdbelvin: Good point.  I'll update this.

cmd/keytransparency-client/grpcc/grpc_client.go
cesarghali: I think this is fine.

cmd/keytransparency-server/frontend.go
```diff
@@ -122,6 +121,14 @@ func grpcHandlerFunc(grpcServer *grpc.Server, otherHandler http.Handler) http.Ha
 	})
 }
 
+func connectToLog(endpoint string, opts []grpc.DialOption) (trillian.TrillianLogClient, error) {
```
cesarghali: How about using `opts ...DialOption` as an argument? This will allow you to call connetToLog with no dial options instead of creating an empty slice, e.g., line 213 below.

core/client/kt/verify.go
```diff
@@ -121,9 +123,9 @@ func (v *Verifier) VerifyGetEntryResponse(ctx context.Context, userID string, in
 	}
 	Vlog.Printf("✓ Signed Map Head signature verified.")
 
-	// Update the log root.
+	// Verify consistency proof between root and newroot.
 	// TODO(gdbelvin): Gossip root.
-	if err := v.log.UpdateRoot(in.LogRoot, in.LogConsistency); err != nil {
+	if err := v.log.VerifyRoot(trusted, in.LogRoot, in.LogConsistency); err != nil {
 		return fmt.Errorf("UpdateRoot(%v, %v): %v", in.LogRoot, in.LogConsistency, err)
```
cesarghali: Please update the error message with the correct function name.

cmd/keytransparency-server/frontend.go
gdbelvin: good idea

core/client/kt/verify.go

cmd/keytransparency-server/frontend.go
gdbelvin: good idea

PR #575, opened 2017-05-05
Make AppID part of the VRF input

### LINKED ISSUE #177: Key Filtering
Allow users to request keys for only a particular app id.

### LINKED ISSUE #510: Move AppIDs into the VRF
Querying an account should not leak other appIDs. 

- Incorporate the appID into the identity string fed to the VRF.  
Something like `VRF(appID, userID)` 
- Remove appID from the `Profile` but keep the map structure.  It might be useful in the future.

gdbelvin:
Improve privacy by making users query the VRF separately for each app.
Querying the keys for one app won't reveal info about other apps.

- Each UserID + AppID is stored in a separate map leaf.
- VRF input is now UserID + AppID
- Commitment input is now UserID + AppID + KeyData
- tpb.Profile is removed. Apps can use whatever data format they wish.

Closes #510
Closes #177


core/crypto/vrf/vrf_test.go
```diff
@@ -0,0 +1,36 @@
...
+// limitations under the License.
+
+package vrf
+
+import (
+	"bytes"
+	"testing"
+)
+
+func TestUniqueID(t *testing.T) {
```
cesarghali: Should we also add a test with hard-coded expected unique ID values? Something similar to what we did for the tree verifier.

cmd/keytransparency-client/cmd/post.go
```diff
@@ -35,12 +33,12 @@ var (
...
 // postCmd represents the post command
 var postCmd = &cobra.Command{
-	Use:   "post [user email] -d {profile}",
+	Use:   "post [user email] [app] -d {key data}",
 	Short: "Update the account with the given profile",
 	Long: `Post replaces the current key-set with the provided key-set, 
 and verifies that both the previous and current key-sets are accurate. eg:
 
-./keytransparency-client post foobar@example.com -d '{"app1": "dGVzdA==", "app2": "dGVzdA=="}'
+./keytransparency-client post foobar@example.com app1 -d "01020304"
```
gdbelvin: I thought that hex might be more user friendly. If you think base64 is more friendly it's an easy switch. 

Implicit is this change is the idea that each app will be running their own client, so I think this change is reasonable.

core/crypto/vrf/vrf_test.go
gdbelvin: Hmm...  I don't think so.  (Happy to be convinced otherwise).
Do we want to fix the UniqueID format in the test? 

As is, the tests exercise a possible attack scenarios and make sure that unique values are being produced from unique inputs.

gdbelvin:
Updated the PR info.  Are there particular docs you know that need to be updated?

cmd/keytransparency-client/cmd/post.go
cesarghali: I agree about not having multiple apps per update.

I think base64 is easier since it's directly accessible from the PEM without any format changes. Thanks for the change.

core/crypto/vrf/vrf_test.go
cesarghali: Correct, and I like that we have such a test! However, this test doesn't test the functionality of UniqueID function, i.e., given an input, test that it generates the expected output. I remember in the tree verifiers test, when we changed the verification algorithm, specifically the way we hash, I had to manually recalculate the hashes and change their values in the test. This was a clear sign to reviewers that a major (and potentially dangerous) change is happening. What do you think?

core/crypto/vrf/vrf_test.go
```diff
@@ -0,0 +1,36 @@
...
+	for _, tc := range []struct {
+		userID, appID   string
+		muserID, mappID string
+	}{
+		{"foo", "app", "fooa", "pp"},
+		{"foo", "app", "", "fooapp"},
```
cesarghali: Can we also add the following case?

```
muserID: "fooapp"
mappID: ""
```

gdbelvin: done

PR #576, opened 2017-05-05
Move core/commitments to core/crypto/commitments

gdbelvin:


cmd/keytransparency-client/cmd/post.go
```diff
@@ -35,12 +33,12 @@ var (
...
 // postCmd represents the post command
 var postCmd = &cobra.Command{
-	Use:   "post [user email] -d {profile}",
+	Use:   "post [user email] [app] -d {key data}",
 	Short: "Update the account with the given profile",
 	Long: `Post replaces the current key-set with the provided key-set, 
 and verifies that both the previous and current key-sets are accurate. eg:
 
-./keytransparency-client post foobar@example.com -d '{"app1": "dGVzdA==", "app2": "dGVzdA=="}'
+./keytransparency-client post foobar@example.com app1 -d "01020304"
```
cesarghali: What's the motivation of this change? And can we still pass multiple apps in one command as it used to be before?

cesarghali: Also in the previous form, users can simply copy the public key material from a PEM file if they have one. Now they need to convert that to hexa before using the client, right?

Issue #59, opened 2015-08-13
Implement Monitor APIs


cesarghali:
This includes:
- [x] Add proto-related material.
- [x] Implement GetMutations.
- [ ] Implement GetMutationsStream.

PR #581, opened 2017-05-16
Add End Sequence to ReadRange API

cesarghali:
This allows callers to specify both count and end sequence in mutations `ReadRange` API. The returned range stops when either count or end sequence is reached, whichever comes first.

Partial #59.

### DIFF
--- a/core/keyserver/keyserver_test.go
+++ b/core/keyserver/keyserver_test.go
@@ -217,7 +217,7 @@ func (fakeFactory) NewTxn(ctx context.Context) (transaction.Txn, error) {
 // mutator.Mutation fake
 type fakeMutation struct{}
 
-func (fakeMutation) ReadRange(txn transaction.Txn, startSequence uint64, count int) (uint64, []*tpb.SignedKV, error) {
+func (fakeMutation) ReadRange(txn transaction.Txn, startSequence uint64, endSequence uint64, count int32) (uint64, []*tpb.SignedKV, error) {
 	return 0, nil, nil
 }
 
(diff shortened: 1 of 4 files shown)


core/keyserver/keyserver_test.go
```diff
@@ -217,7 +217,7 @@ func (fakeFactory) NewTxn(ctx context.Context) (transaction.Txn, error) {
 // mutator.Mutation fake
 type fakeMutation struct{}
 
-func (fakeMutation) ReadRange(txn transaction.Txn, startSequence uint64, count int) (uint64, []*tpb.SignedKV, error) {
+func (fakeMutation) ReadRange(txn transaction.Txn, startSequence uint64, endSequence uint64, count int32) (uint64, []*tpb.SignedKV, error) {
```
gdbelvin: `startSequence, endSequence uint64`

impl/sql/mutations/mutations.go
```diff
@@ -35,7 +35,7 @@ const (
 	VALUES (?, ?, ?);`
 	readRangeExpr = `
   	SELECT Sequence, Mutation FROM Mutations
-  	WHERE MapID = ? AND Sequence > ?
+  	WHERE MapID = ? AND Sequence > ? AND Sequence <= ?
```
gdbelvin: Are you sure we don't want 
sequence >= start &&
sequence < end 
?

cesarghali: I don't think so. There's no way to know the exact starting sequence giving that we only store the highest sequence number in each SignedMapRoot. I remember we decided that we want a range of ( ] to cover implementations using time-based sequence numbers.

gdbelvin: Good point.

PR #582, opened 2017-05-16
Trillian integration fixups

gdbelvin:
- Connections to trillian backends do not use TLS.
- The signer wasn't saving the logID.


cmd/keytransparency-server/frontend.go
```diff
@@ -202,7 +202,7 @@ func main() {
 	mutator := entry.New()
 
 	// Connect to log server.
-	tconn, err := grpc.Dial(*logURL)
+	tconn, err := grpc.Dial(*logURL, grpc.WithInsecure())
```
cesarghali: Should we add a flag specifying if security is needed? I'm thinking of the scenario where Trillian and Key Transparency servers are not on the same machine.

There's no need to address this in this PR. If you think it's a good idea to have such a flag, we should open an issue.

gdbelvin: When Trillian adds TLS support, we can modify our code to enforce it, but that should be a later PR.

cesarghali: SG.

PR #583, opened 2017-05-20
Implement GetMutations API

cesarghali:
This PR implements monitors `GetMutations` API. It also adds the necessary tests with all appropriate fakes. The `Server` object in `impl/mutation` will eventually become a wrapper to the `Server` object in `core/mutation`. The former tests will eventually be deleted since they won't be necessary.

Partial #59.


core/mutation/mutation.go
```diff
@@ -0,0 +1,211 @@
...
+		MapId:    s.mapID,
+		Revision: in.Epoch,
+	})
+	if err != nil {
+		log.Printf("GetSignedMapRootByRevision(%v, %v): %v", s.mapID, in.Epoch, err)
+		return nil, grpc.Errorf(codes.Internal, "Get signed map root failed")
+	}
+
+	// Get highest and lowest sequence number.
+	highestSeq := uint64(resp.GetMapRoot().GetMetadata().HighestFullyCompletedSeq)
```
gdbelvin: Might want to use the `Get` form here to avoid dereferencing nil. 
The Trillian Map Server has been known to omit values in the response...

core/mutation/mutation.go
```diff
@@ -0,0 +1,211 @@
...
+	if err := txn.Commit(); err != nil {
+		return nil, fmt.Errorf("txn.Commit(): %v", err)
+	}
+	indexes := make([][]byte, 0, len(mRange))
+	mutations := make([]*tpb.Mutation, 0, len(mRange))
+	for _, m := range mRange {
+		mutations = append(mutations, &tpb.Mutation{Update: m})
+		indexes = append(indexes, m.GetKeyValue().GetKey())
+	}
+	// Get leaf proofs.
```
gdbelvin: Perhaps add a TODO to make this optional?

core/mutation/mutation.go
```diff
@@ -0,0 +1,211 @@
...
+	proofs, err := s.inclusionProofs(ctx, indexes, in.Epoch)
+	if err != nil {
+		return nil, err
+	}
+	for i, p := range proofs {
+		mutations[i].Proof = p
+	}
+
+	// Fetch log proofs.
+	logRoot, logConsistency, logInclusion, err := s.logProofs(ctx, in.FirstTreeSize, resp.GetMapRoot().MapRevision)
```
gdbelvin: Also the Get form here too

core/mutation/mutation.go
```diff
@@ -0,0 +1,211 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	nextPageToken := ""
+	if len(mutations) == int(in.PageSize) && maxSequence != highestSeq {
```
gdbelvin: What happens on the last page of a set when the last page is smaller than the page size?

core/mutation/mutation_test.go
```diff
@@ -0,0 +1,264 @@
...
+		mutations   []*tpb.SignedKV
+		nextToken   string
+		success     bool
+	}{
+		{"working case complete epoch 1", 1, "", 6, signedKV(t, 1, 6), "", true},
+		{"working case complete epoch 2", 2, "", 4, signedKV(t, 7, 10), "", true},
+		{"working case partial epoch 1", 1, "", 4, signedKV(t, 1, 4), "4", true},
+		{"working case partial epoch 2", 2, "", 3, signedKV(t, 7, 9), "9", true},
+		{"working case larger page size", 1, "", 10, signedKV(t, 1, 6), "", true},
+		{"working case with page token", 1, "4", 2, signedKV(t, 5, 6), "", true},
```
gdbelvin: Add a case for 
- Last page of multi-page set where the last page is smaller than the others.

cesarghali: This should be captured in case `working case larger page size`, where the page size is larger than the number of mutations in the response.

core/mutation/mutation.go
cesarghali: Not sure I'm following, but this case should work as expected. The last page will have few mutations than the request page size and the nextPageToken will be empty. There's a test case for this already. Please check my comment below.

core/mutation/mutation.go
cesarghali: 👍

core/mutation/mutation.go
cesarghali: Done.

However, I don't think this will be a problem. Generally, we're using `Get` for all non-trivial types (e.g., integers) and since Trillian uses `proto3`, any missing field will have a default value (there are no `Has` methods in `proto3` AFAIK). If you think this might still cause a problem on the long run, I'll start using `Get` everywhere and gradually change the codebase in that direction.

Using `Get` everywhere might make the code look nice and be consistent.

cesarghali:
Coverage should be maximal. We got a drop in `impl/mutation` because I removed a test that is not needed anymore. `GetMutations` in that package is just a wrapper for the same function in `core/mutation`, which is already heavily tested.

core/mutation/mutation.go
gdbelvin: Correct that proto3 uses zero values for missing fields, so resp.MissingField will return 0 if it's an int for example. There's a problem, however, if resp itself is nil.  Attempting to dereference nil.MissingField will result in a segfault. The Get* functions test for this and return the proper 0 values even when their parent object is missing.

core/mutation/mutation_test.go
gdbelvin: Please add a case for when a token is provided as well. It looks like the existing tests exercise this only for the first page.

core/mutation/mutation.go
cesarghali: Good point. Will start using `Get` from now on then.

core/mutation/mutation_test.go
cesarghali: Done.

PR #586, opened 2017-05-22
VRF API Update: Verify -> ProofToHash

gdbelvin:
To simplify the VRF API and make it easier to use we make the following
changes:

`Evaluate(m) (vrf, proof) -> Evaluate(m) (index, proof)`
`Verify(m, vrf, proof) (error) -> ProofToHash(m, proof) (index, error)`

- Evaluate emits a single proof item that servers need to send back to
  clients.  Previously, vrf and proof needed to be sent separately.

- Evaluate also emits the index, stronly tying the value that the server
  will actually use to the vrf function itself.

- Verify now computes and returns index only if the proof is valid,
  making the API easier to use and more foolproof.


core/proto/keytransparency_v1_types/keytransparency_v1_types.proto
```diff
@@ -125,37 +125,35 @@ message GetEntryResponse {
   // Privacy layer hides user_id and profile data until requested.
   //
 
-  // vrf is the output of VRF on user_id.
-  bytes vrf = 1;
   // vrf_proof is the proof for VRF on user_id.
-  bytes vrf_proof = 2;
+  bytes vrf_proof = 1;
```
cesarghali: Can you include the generated proto file as well?

core/crypto/vrf/p256/p256.go
```diff
@@ -137,33 +138,42 @@ func (k PrivateKey) Evaluate(m []byte) (vrf, proof []byte) {
 	t := new(big.Int).Sub(ri, new(big.Int).Mul(s, k.D))
 	t.Mod(t, params.N)
 
-	// Write s and t to a proof blob. Also write leading zeros before s and t
+	// VRF_k(m) = [k]H
+	vrfx, vrfy := params.ScalarMult(hx, hy, k.D.Bytes())
+	vrf := elliptic.Marshal(curve, vrfx, vrfy) // 65 bytes.
```
gdbelvin: The X and Y coordinates are 32 bytes, plus one byte to specify the encoding.

cesarghali: Sounds good, thanks.

core/proto/keytransparency_v1_types/keytransparency_v1_types.proto
gdbelvin: done

core/client/kt/verify.go
```diff
@@ -94,12 +94,12 @@ func (v *Verifier) VerifyGetEntryResponse(ctx context.Context, userID, appID str
 	}
 	Vlog.Printf("✓ Commitment verified.")
 
-	if err := v.vrf.Verify(vrf.UniqueID(userID, appID), in.Vrf, in.VrfProof); err != nil {
+	index, err := v.vrf.ProofToHash(vrf.UniqueID(userID, appID), in.VrfProof)
+	if err != nil {
 		Vlog.Printf("✗ VRF verification failed.")
 		return fmt.Errorf("vrf.Verify(%v, %v): %v", userID, appID, err)
```

gdbelvin:
Updated. PTAL

PR #591, opened 2017-05-24
Migrate to glog

### LINKED ISSUE #590: Implment leveled logging
Support debugging live systems with leveled logging.

gdbelvin:
- Support multi-level logging.
- Add a few extra log messages in the signer for debugging.
- Closes #590


cmd/keytransparency-signer/backend.go
```diff
@@ -104,42 +102,37 @@ func main() {
...
 
 	signer := signer.New(*domain, *mapID, tmap, *logID, sths, mutator, mutations, factory)
-	go signer.StartSigning(context.Background(), *epochDuration)
-
-	log.Printf("Signer started.")
-
-	var wg sync.WaitGroup
-	wg.Add(1)
-	wg.Wait()
+	glog.Infof("Signer started.")
```
cesarghali: Should this be `Info` instead of `Infof`?

core/keyserver/keyserver.go
```diff
@@ -85,7 +84,12 @@ func New(logID int64,
 // this user and that it is the same one being provided to everyone else.
 // GetEntry also supports querying past values by setting the epoch field.
 func (s *Server) GetEntry(ctx context.Context, in *tpb.GetEntryRequest) (*tpb.GetEntryResponse, error) {
-	return s.getEntry(ctx, in.UserId, in.AppId, in.FirstTreeSize, -1)
+	resp, err := s.getEntry(ctx, in.UserId, in.AppId, in.FirstTreeSize, -1)
```
cesarghali: This should be fine the way it was, `getEntry` already logs the error and returns a gRPC error with not much information. I don't think there's a need to log that.

core/keyserver/keyserver.go
```diff
@@ -132,7 +136,7 @@ func (s *Server) getEntry(ctx context.Context, userID, appID string, firstTreeSi
 			LogId: s.logID,
 		})
 	if err != nil {
-		log.Printf("tlog.GetLatestSignedLogRoot(%v): %v", s.logID, err)
+		glog.Warningf("tlog.GetLatestSignedLogRoot(%v): %v", s.logID, err)
```
cesarghali: Why is this a warning and others are errors? In general, how are we deciding if it's an error or a warning?
Same below in this file.

I would assume if the execution can continue it's a warning, otherwise it's an error. This one seems like an error to me. What do you think?

cmd/keytransparency-server/frontend.go
```diff
@@ -71,22 +71,22 @@ var (
 func openDB() *sql.DB {
 	db, err := sql.Open(engine.DriverName, *serverDBPath)
 	if err != nil {
-		log.Fatalf("sql.Open(): %v", err)
+		glog.Exitf("sql.Open(): %v", err)
```
cesarghali: Why not use `glog.Fatalf` here? We're using it in `signer.go`. I'm fine with either if we don't care about the difference between the two, I'm just concerned about consistency.

cmd/keytransparency-signer/backend.go
gdbelvin: Both are valid in this case, though my personal preference is `Infof` for consistency.

core/keyserver/keyserver.go
gdbelvin: Good point.

core/keyserver/keyserver.go
gdbelvin: This one should have been a warning. 

In general I'm using errors for internal failures, and warnings for input validation coming from the user.

cmd/keytransparency-signer/backend.go
cesarghali: SG.

core/keyserver/keyserver.go
cesarghali: When I check logs, I first look for errors (not warning). An invalid user input would make the API call fail, and I think it should be an error. What do you think? If you thing these are fine as warnings, that's works for me.

cmd/keytransparency-server/frontend.go
gdbelvin: Fatalf prints a stack trace, which I don't think we need here.

core/keyserver/keyserver.go
gdbelvin: woops. typeo.  I meant "This one should have been an error"

PR #592, opened 2017-05-24
Minor Fixes

gdbelvin:
- Better error messages.
- Request the latest revision in signer.

### DIFF
--- a/core/keyserver/keyserver.go
+++ b/core/keyserver/keyserver.go
@@ -85,12 +85,7 @@ func New(logID int64,
 // this user and that it is the same one being provided to everyone else.
 // GetEntry also supports querying past values by setting the epoch field.
 func (s *Server) GetEntry(ctx context.Context, in *tpb.GetEntryRequest) (*tpb.GetEntryResponse, error) {
-	resp, err := s.getEntry(ctx, in.UserId, in.AppId, in.FirstTreeSize, -1)
-	if err != nil {
-		log.Printf("getEntry failed: %v", err)
-		return nil, grpc.Errorf(codes.Internal, "GetEntry failed")
-	}
-	return resp, nil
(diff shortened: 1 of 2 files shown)


core/keyserver/keyserver.go
```diff
@@ -104,7 +99,7 @@ func (s *Server) getEntry(ctx context.Context, userID, appID string, firstTreeSi
 	})
 	if err != nil {
 		log.Printf("GetLeaves(): %v", err)
-		return nil, grpc.Errorf(codes.Internal, "Failed fetching map leaf")
+		return nil, grpc.Errorf(codes.Internal, "Failed fetching map leaf: %v", err)
```
cesarghali: I thought we don't want to return the actual error back to the client, this is why we log it, right? We do this everywhere in this package and also in `core/mutation`. If we start returning the actual error to the client we should do it everywhere, but I think our current approach is fine.

gdbelvin: yes, you're right. Ideally, we'd have a grpc interceptor that could strip out the error messages that would have a debug flag we could turn on and off. but this will do for now.

core/keyserver/keyserver.go
```diff
@@ -99,7 +99,7 @@ func (s *Server) getEntry(ctx context.Context, userID, appID string, firstTreeSi
 	})
 	if err != nil {
 		log.Printf("GetLeaves(): %v", err)
-		return nil, grpc.Errorf(codes.Internal, "Failed fetching map leaf: %v", err)
+		return nil, grpc.Errorf(codes.Internal, "Failed fetching map leaf", err)
```
cesarghali: I think we can remove `err` here, right?
```
return nil, grpc.Errorf(codes.Internal, "Failed fetching map leaf")
```

PR #593, opened 2017-05-25
Only create epochs if there is something to process or the maximum nu…

### LINKED ISSUE #509: Don't create epochs when there is nothing to process
Define minimum and maximum durations for updating epochs. 
Mutations in between the minimum duration should be batched.
No epochs should be created when there are no mutations to process - unless the maximum epoch duration has passed.

Liamsi:
…mber of min epoch intervals has passed

- rename flag period to min-period
- add max-period flag (number of passed min-periods until signer creates epoch independent from mutations)
- add flag to `CreateEpoch` to let the caller enforce creating a new epoch (independent from mutations); update tests
resolves #509

### DIFF
--- a/cmd/keytransparency-signer/backend.go
+++ b/cmd/keytransparency-signer/backend.go
@@ -42,10 +42,10 @@ import (
 )
 
 var (
-	serverDBPath  = flag.String("db", "db", "Database connection string")
-	domain        = flag.String("domain", "example.com", "Distinguished name for this key server")
-	minEpochDuration = flag.Duration("min-period", time.Second*60, "Minimum time between epoch creation (create epochs only if there where mutations)")
-	maxEpochDuration = flag.Duration("max-period", time.Hour*12, "Maximum time between epoch creation (independent from mutations)")
+	serverDBPath     = flag.String("db", "db", "Database connection string")
+	domain           = flag.String("domain", "example.com", "Distinguished name for this key server")
+	minEpochDuration = flag.Duration("min-period", time.Second*60, "Minimum time between epoch creation (create epochs only if there where mutations). Expected to be smaller than max-period.")
+	maxEpochDuration = flag.Duration("max-period", time.Hour*12, "Maximum time between epoch creation (independent from mutations). This value should about half the time guaranteed by the policy.")
 
 	// Info to connect to sparse merkle tree database.
 	mapID  = flag.Int64("map-id", 0, "ID for backend map")
@@ -134,7 +134,7 @@ func main() {
 	mutator := entry.New()
 
 	if *maxEpochDuration < *minEpochDuration {
-		glog.Exitf("Maximum time between epoch creation (%v) must be grater or equal minimum time between epoch creation.")
+		glog.Exitf("maxEpochDuration < minEpochDuration: %v < %v, want maxEpochDuration >= minEpochDuration")
 	}
 
 	signer := signer.New(*domain, *mapID, tmap, *logID, sths, mutator, mutations, factory)
--- a/cmd/keytransparency-signer/backend.go
+++ b/cmd/keytransparency-signer/backend.go
@@ -45,7 +45,7 @@ var (
 	serverDBPath  = flag.String("db", "db", "Database connection string")
 	domain        = flag.String("domain", "example.com", "Distinguished name for this key server")
 	minEpochDuration = flag.Duration("min-period", time.Second*60, "Minimum time between epoch creation (create epochs only if there where mutations)")
-	maxElapsedEpochs = flag.Uint("max-period", 60, "Maximum elapsed min-periods until epoch creation (independent from mutations)")
+	maxEpochDuration = flag.Duration("max-period", time.Hour*12, "Maximum time between epoch creation (independent from mutations)")
 
 	// Info to connect to sparse merkle tree database.
 	mapID  = flag.Int64("map-id", 0, "ID for backend map")
@@ -135,5 +135,5 @@ func main() {
 
 	signer := signer.New(*domain, *mapID, tmap, *logID, sths, mutator, mutations, factory)
 	glog.Infof("Signer started.")
-	signer.StartSigning(context.Background(), *minEpochDuration, *maxElapsedEpochs)
+	signer.StartSigning(context.Background(), *minEpochDuration, *maxEpochDuration)
 }
--- a/cmd/keytransparency-signer/backend.go
+++ b/cmd/keytransparency-signer/backend.go
@@ -133,6 +133,10 @@ func main() {
 	}
 	mutator := entry.New()
 
+	if *maxEpochDuration < *minEpochDuration {
+		glog.Exitf("Maximum time between epoch creation (%v) must be grater or equal minimum time between epoch creation.")
+	}
+
 	signer := signer.New(*domain, *mapID, tmap, *logID, sths, mutator, mutations, factory)
 	glog.Infof("Signer started.")
 	signer.StartSigning(context.Background(), *minEpochDuration, *maxEpochDuration)
--- a/cmd/keytransparency-signer/backend.go
+++ b/cmd/keytransparency-signer/backend.go
@@ -44,7 +44,8 @@ import (
 var (
 	serverDBPath  = flag.String("db", "db", "Database connection string")
 	domain        = flag.String("domain", "example.com", "Distinguished name for this key server")
-	epochDuration = flag.Duration("period", time.Second*60, "Time between epoch creation")
+	minEpochDuration = flag.Duration("min-period", time.Second*60, "Minimum time between epoch creation (create epochs only if there where mutations)")
+	maxElapsedEpochs = flag.Uint("max-period", 60, "Maximum elapsed min-periods until epoch creation (independent from mutations)")
 
 	// Info to connect to sparse merkle tree database.
 	mapID  = flag.Int64("map-id", 0, "ID for backend map")
@@ -134,5 +135,5 @@ func main() {
 
 	signer := signer.New(*domain, *mapID, tmap, *logID, sths, mutator, mutations, factory)
 	glog.Infof("Signer started.")
-	signer.StartSigning(context.Background(), *epochDuration)
+	signer.StartSigning(context.Background(), *minEpochDuration, *maxElapsedEpochs)
 }
--- a/core/signer/signer.go
+++ b/core/signer/signer.go
@@ -44,13 +44,13 @@ type Signer struct {
 
 // New creates a new instance of the signer.
 func New(realm string,
-	mapID int64,
-	tmap trillian.TrillianMapClient,
-	logID int64,
-	sths appender.Remote,
-	mutator mutator.Mutator,
-	mutations mutator.Mutation,
-	factory transaction.Factory) *Signer {
+mapID int64,
+tmap trillian.TrillianMapClient,
+logID int64,
+sths appender.Remote,
+mutator mutator.Mutator,
+mutations mutator.Mutation,
+factory transaction.Factory) *Signer {
 	return &Signer{
 		realm:     realm,
 		mapID:     mapID,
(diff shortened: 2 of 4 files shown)


Liamsi:
I signed it!

cmd/keytransparency-signer/backend.go
```diff
@@ -45,7 +45,8 @@ import (
 var (
 	serverDBPath  = flag.String("db", "db", "Database connection string")
 	domain        = flag.String("domain", "example.com", "Distinguished name for this key server")
-	epochDuration = flag.Duration("period", time.Second*60, "Time between epoch creation")
+	minEpochDuration = flag.Duration("min-period", time.Second*60, "Minimum time between epoch creation (create epochs only if there where mutations)")
+	maxElapsedEpochs = flag.Uint("max-period", 60, "Maximum elapsed min-periods until epoch creation (independent from mutations)")
```
cesarghali: Can this also be `flag.Duration` like `minEpochDuration`?

cesarghali: Can this also be `flag.Duration` like `minEpochDuration`?

core/signer/signer.go
```diff
@@ -145,7 +157,7 @@ func (s *Signer) applyMutations(mutations []*tpb.SignedKV, leaves []*trillian.Ma
 }
 
 // CreateEpoch signs the current map head.
-func (s *Signer) CreateEpoch(ctx context.Context) error {
+func (s *Signer) CreateEpoch(ctx context.Context, alwaysCreateNewEpoch bool) error {
```
gdbelvin: go style prefers shorter names.  Consider `forceNewEpoch`?

core/signer/signer.go
```diff
@@ -64,9 +64,21 @@ func New(realm string,
...
-		if err := s.CreateEpoch(ctx); err != nil {
+func (s *Signer) StartSigning(ctx context.Context, minInterval time.Duration, maxElapsed uint) {
+	// count elapsed minIntervals:
+	elapsed := uint(0)
+	for range time.NewTicker(minInterval).C {
+		elapsed++
+		var createNewEpoch bool
+		if elapsed >= maxElapsed {
+			createNewEpoch = true
+			elapsed = 0
```
cesarghali: How about we use a time here for this instead of a counter, this way the unit of `maxElapsed` is not coupled with `minInterval`? What do you think?

cesarghali:
Also please run `git rebase master` to update the branch. Thanks.

core/signer/signer.go
Liamsi: Done (in https://github.com/google/keytransparency/pull/593/commits/5987035a462e7f080285f51e879b117ae5e2b2b1)

cmd/keytransparency-signer/backend.go
Liamsi: Done (in https://github.com/google/keytransparency/pull/593/commits/08bac3f7af45be342be3a6b758f949e5d4aebefe)

core/signer/signer.go
Liamsi: Thanks for the suggestion! I coupled both like above because I wanted to avoid creating two independent "threads"/go-routines that might "race" for the epoch creation. But your suggestions is simpler and avoid this, too (if I understood it correctly).

Liamsi:
Thanks a lot for your reviews @gdbelvin @cesarghali. I made some changes and I think this might be ready for another review. 

> Please add a test

I've added one that tests the epoch creation logic (counts if calling `CreateEpoch` with `forceNewEpoch` as expected). 
Another thing that might be worthwhile to test is if epochs in between `maxDurations` are really only created if there are mutations. I didn't include that in this PR as one needs to fire up quite a lot to have a `Signer` running. Maybe this is sth. which could into some `integration` package.

cmd/keytransparency-signer/backend.go
```diff
@@ -132,7 +133,11 @@ func main() {
 	}
 	mutator := entry.New()
 
+	if *maxEpochDuration < *minEpochDuration {
+		glog.Exitf("Maximum time between epoch creation (%v) must be grater or equal minimum time between epoch creation.")
```
gdbelvin: Consider following the canonical "got, want" style for error messages. 
https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures

core/signer/signer.go
```diff
@@ -63,10 +63,26 @@ func New(realm string,
...
-	for range time.NewTicker(interval).C {
-		if err := s.CreateEpoch(ctx); err != nil {
+// StartSigning advance epochs once per minInterval, if there were mutations, and at least once per
+// maxElapsed minIntervals.
+func (s *Signer) StartSigning(ctx context.Context, minInterval time.Duration, maxInterval time.Duration) {
+	tc := time.Tick(minInterval)
+	processEpochs(ctx, tc, maxInterval, s.CreateEpoch)
+}
+
+func processEpochs(ctx context.Context, tc <-chan time.Time, maxElapsed time.Duration, sign func(ctx context.Context, enforce bool) error) {
```
gdbelvin: please add a function comment

core/signer/signer.go
```diff
@@ -63,10 +63,26 @@ func New(realm string,
...
+	last := time.Now()
```
gdbelvin: fetch time last from previous map head if it exists?

core/signer/signer.go
```diff
@@ -63,10 +63,26 @@ func New(realm string,
...
+	for now := range tc {
+		var forceNewEpoch bool
+		if time.Since(last) >= maxElapsed {
+			forceNewEpoch = true
+			last = now
+		} else {
+			forceNewEpoch = false
+		}
+		// only create a new epoch if elapsed intervals >= max. elapsed intervals
```
gdbelvin: s/only/Only/

core/signer/signer_test.go
```diff
@@ -0,0 +1,53 @@
+package signer
+
+import (
+	"testing"
+	"time"
+	"golang.org/x/net/context"
+)
+
+func TestEpochCreation(t *testing.T) {
```
gdbelvin: Prefer table driven tests if at all possible. 
https://github.com/golang/go/wiki/TableDrivenTests

core/signer/signer_test.go
```diff
@@ -0,0 +1,53 @@
...
+				t.Errorf("More epochs created (%d) than expected (%d)", got, want)
+			}
+		}
+	}
+
+	// see if only one epoch creation was enforced:
+	if got != want {
+		t.Fatalf("Expected %v maxEpoch(s) but got %v", want, got)
+	}
+}
```
gdbelvin: add newline

core/signer/signer.go
```diff
@@ -63,10 +63,26 @@ func New(realm string,
...
+		// OR there were mutations in between max. elapsed epochs (expected to be handled by passed callback):
```
cesarghali: Although there's no line limit in Go, we're trying to keep comments to 80 characters a line for better readability.

core/signer/signer.go
```diff
@@ -63,10 +63,26 @@ func New(realm string,
...
+		if err := sign(ctx, forceNewEpoch); err != nil {
```
cesarghali: Should this be a call to `CreateEpoch`? I can't find `sign` function.

core/signer/signer.go
cesarghali: Actually, do we need this function? `StartSigning` just calls `processEpochs`.

Liamsi: The only reason I only created `processEpochs` to make it possible to test the underlying logic without the need of creating a whole `Signer` (which depends on a lot of other running objects, like the DB, a mutator, transaction factory...). Other than making testing convinient we don't need, I guess.

core/signer/signer.go
Liamsi: Indeed, this is a call to `CreateEpoch`. `sign` is a "callback" passed to `processEpochs` (and `s.CreateEpoch` is passed as an argument).
Again, this only exists to make it testable without having to run a whole environment (like for instance in this [integration](https://github.com/google/keytransparency/blob/master/integration/testutil.go#L154-L213) package).

cesarghali: That makes sense, thanks for the explanation.

cmd/keytransparency-signer/backend.go
gdbelvin: It's only required for tests, but it's useful info to have for other failures too because it also gives the desired value.

core/signer/signer.go
```diff
@@ -44,13 +44,13 @@ type Signer struct {
...
 // New creates a new instance of the signer.
 func New(realm string,
-	mapID int64,
-	tmap trillian.TrillianMapClient,
-	logID int64,
-	sths appender.Remote,
-	mutator mutator.Mutator,
-	mutations mutator.Mutation,
-	factory transaction.Factory) *Signer {
+mapID int64,
```
gdbelvin: Make sure to run gofmt. Looks like there's a tab missing.

gdbelvin: Make sure to run gofmt. Looks like there's a tab missing.

core/signer/signer.go
```diff
@@ -63,10 +63,66 @@ func New(realm string,
 	}
 }
 
-// StartSigning advance epochs once per interval.
-func (s *Signer) StartSigning(ctx context.Context, interval time.Duration) {
-	for range time.NewTicker(interval).C {
-		if err := s.CreateEpoch(ctx); err != nil {
+// StartSigning advance epochs once per minInterval, if there were mutations,
+// and at least once per maxElapsed minIntervals.
+func (s *Signer) StartSigning(ctx context.Context, minInterval time.Duration, maxInterval time.Duration) {
```
gdbelvin: You can shorten the function signature by dropping repeated types
`ctx context.Context, minInterval, maxInterval time.Duration`

core/signer/signer.go
```diff
@@ -63,10 +63,66 @@ func New(realm string,
...
+	// TODO Fetch last time from previous map head (as stored in the map server) if it exists:
```
gdbelvin: TODO's should have the form: 
TODO(username):

core/signer/signer.go
```diff
@@ -63,10 +63,66 @@ func New(realm string,
...
+	last := time.Now()
+	processEpochs(ctx, last, minInterval, maxInterval, s.CreateEpoch, nil)
+}
+
+// processEpochs calls the given callback (usually Signer.CreateEpoch) every
+// minElapsed and every maxElapsed time.
```
gdbelvin: Collapse comments into solid blocks 80 chars wide. 
in vim: `gq`

core/signer/signer.go
cesarghali: Running `make presubmit` would take care of all of that and shows other possible errors :)

cmd/keytransparency-signer/backend.go
cesarghali: SG.

core/signer/signer.go
```diff
@@ -63,16 +63,65 @@ func New(realm string,
...
+		mapRoot := rootResp.GetMapRoot()
+		if mapRoot != nil {
+			last = time.Unix(0, mapRoot.TimestampNanos)
+		} else {
+			last = time.Now()
+		}
+	}
+	// Start issuing epochs:
+	enforceChan := processEpochs(last, minInterval, maxInterval)
+	for f := range enforceChan {
```
gdbelvin: `for f := range epochTicker(last, minInterval, maxInterval)` ?

core/signer/signer.go
```diff
@@ -63,16 +63,65 @@ func New(realm string,
...
 			glog.Fatalf("CreateEpoch failed: %v", err)
 		}
 	}
 }
 
-// newMutations returns a list of mutations to process and highest sequence number returned.
+// processEpochs returns and sends to a bool channel every time an epoch should
+// be created. If the boolean value is true this indicates that the epoch should
+// be created independent from if there were mutations or not.
+func processEpochs(last time.Time, minElapsed, maxElapsed time.Duration) <-chan bool {
```
gdbelvin: processesEpochs is a slightly confusing function name since it no longer processes epochs.  :-)
Consider epochTicker or another term that indicates that it outputs Ticks

core/signer/signer.go
```diff
@@ -63,16 +63,65 @@ func New(realm string,
...
+	go func() {
+		// Do not wait for the first minDuration to pass but directly resume from
+		// last
+		if (time.Since(last) + minElapsed) >= maxElapsed {
+			enforce <- true
+			last = time.Now()
+		}
+
+		for now := range tc {
+			if (time.Since(last) + minElapsed) >= maxElapsed {
```
gdbelvin: We should be consistent for what value of "now" to use. 
Consider using the `now` variable from the for loop rather than the embedded `time.Now()` inside `Since`

Liamsi: Good catch! Changed.

Liamsi:
Thanks for the review! I addressed your comments in https://github.com/google/keytransparency/pull/593/commits/2b9b2b432ba646f03beff446c78f86f264675a65.

PR #596, opened 2017-06-03
Add test vectors to crypto/commitment

gdbelvin:
Merge #595 first

Add a modifiable source of randomness to facilitate reliable tests.
Uses the new dev.Zero reader to generate test vectors.

### DIFF
--- a/core/crypto/commitments/commitments_test.go
+++ b/core/crypto/commitments/commitments_test.go
@@ -59,7 +59,11 @@ func TestVectors(t *testing.T) {
 		userID, appID, data string
 		want                []byte
 	}{
+		{"", "", "", dh("0698789c7beed09e93848e4df08be5c911de534d286abcbf69359debe4c62bc2")},
 		{"foo", "app", "bar", dh("064c8933f50f897e8b179065c6b3ec13e9d093337c6d403c77e3ed1701378ed6")},
+		{"foo1", "app", "bar", dh("77015921f7fe584e1b5866a32ab9f305715c4e0241581d41f66ee34b24cdb566")},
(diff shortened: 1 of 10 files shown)


core/crypto/commitments/commitments_test.go
```diff
@@ -45,3 +49,39 @@ func TestCommit(t *testing.T) {
 		}
 	}
 }
+
+func TestVectors(t *testing.T) {
+	// Use a constant key of zero to obtain consistent test vectors.
+	// Real commentment library MUST use random keys.
```
alsophian: nit: s/commentment/commitment/

gdbelvin: done. thanks

core/crypto/commitments/commitments_test.go
```diff
@@ -45,3 +49,39 @@ func TestCommit(t *testing.T) {
...
+func TestVectors(t *testing.T) {
+	// Use a constant key of zero to obtain consistent test vectors.
+	// Real commitment library MUST use random keys.
+	Rand = dev.Zeros
+	zeroKey := []byte{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0}
+	for _, tc := range []struct {
+		userID, appID, data string
+		want                []byte
+	}{
+		{"foo", "app", "bar", dh("064c8933f50f897e8b179065c6b3ec13e9d093337c6d403c77e3ed1701378ed6")},
```
cesarghali: We should expect the same `want` for all input, right? Can we add a couple other cases with different input and same `want`?

PR #598, opened 2017-06-05
Only create epochs if there is something to process

### LINKED ISSUE #509: Don't create epochs when there is nothing to process

Liamsi:
~~Give travis machines slightly more time to create another max-epoch (5 additional iterations to read from the channel should be enough)~~

resolves #509


gdbelvin:
It looks like the time based test isn't robust enough.  It might take some creativity, but it would be great to have a time independent test. A couple ideas:

1) Supply a fake clock for tests that we can advance predictably. 
2) Supply our own "tick" channel in the tests and count ticks.

core/signer/signer_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+// no offset (last epoch was created now or does not exist yet)
+const zero = 0 * time.Second
+
+func TestEpochCreation(t *testing.T) {
+	clock := fakeClock{now: parseTime(t, fakeNow)}
+	now := clock.Now()
+	for _, tc := range []struct {
+		wantForced int
+
```
cesarghali: Please remove this empty line.

core/signer/signer_test.go
```diff
@@ -0,0 +1,111 @@
...
+		fakeNow    time.Time
+		lastForced time.Time
+
```
cesarghali: Please remove this empty line.

core/signer/signer_test.go
```diff
@@ -0,0 +1,111 @@
...
+		nTicks int
+
```
cesarghali: Please remove this empty line.

core/signer/signer.go
```diff
@@ -63,16 +63,78 @@ func New(realm string,
...
+	return time.Now()
+}
+
+func (c sysClock) Since(t time.Time) time.Duration {
+	return time.Since(t)
+}
+
+// genEpochTicks returns and sends to a bool channel every time an epoch should
+// be created. If the boolean value is true this indicates that the epoch should
+// be created independent from if there were mutations or not.
```
cesarghali: s/independent from if there were mutations or not/regardless whether mutations exist?

core/signer/signer_test.go
```diff
@@ -0,0 +1,111 @@
...
+		tc <- last
+	}
+	return tc
+}
+
+// parseTime creates a time.Time from a time.RFC3339 formatted string.
+func parseTime(t *testing.T, ts string) time.Time {
+	ti, err := time.Parse(time.RFC3339, ts)
+	if err != nil {
+		t.Fatalf("parseTime(_, %v) expects time.RFC3339 formatted time strings as argument", ts)
```
cesarghali: Consider using `panic` here and not passing `t`. Since we're using hard-coded time strings here, this should never fail.

core/signer/signer_test.go
```diff
@@ -0,0 +1,111 @@
...
+import (
+	"testing"
+	"time"
+)
+
+const (
+	minDurationS = time.Second * 1
+	maxDurationH = time.Hour * 6
+	minInMax     = int(maxDurationH / minDurationS)
+	fakeNow      = "2015-10-21T04:29:00+00:00"
```
cesarghali: If `parseTime` uses panic (see comment below), you can directly use it here.

Liamsi: On one hand I'd prefer to keep these `const` (which would yield `const initializer parseTime("2015-10-21T04:29:00+00:00") is not a constant`). On the other hand it does increase the readability to initialize the vars as `time.Time` directly. Done in https://github.com/google/keytransparency/pull/598/commits/ce2f58852c917fe884c2c6111bfb61d83abdd382.

core/signer/signer.go
```diff
@@ -63,16 +63,78 @@ func New(realm string,
...
+func (s *Signer) StartSigning(ctx context.Context, minInterval, maxInterval time.Duration) {
+	// Fetch last time from previous map head (as stored in the map server) if it
+	// exists:
+	var last time.Time
+	rootResp, err := s.tmap.GetSignedMapRoot(ctx, &trillian.GetSignedMapRootRequest{
+		MapId: s.mapID,
+	})
+	if err != nil {
+		glog.Infof("GetSignedMapRoot failed: %v", err)
+		last = time.Now()
```
gdbelvin: If we failed to get the last root, we should be conservative and assume that we should create a new one right away, rather than waiting for the max period.

core/signer/signer.go
```diff
@@ -63,16 +63,78 @@ func New(realm string,
...
 	}
 }
 
-// newMutations returns a list of mutations to process and highest sequence number returned.
+type clock interface {
+	Now() time.Time
+	Since(time.Time) time.Duration
+}
+
+type sysClock struct{}
```
gdbelvin: even better, pull this clock out into a separate package. 

Then expose the sys clock a singleton.  In Go this looks like:
```go
package clock
// System is a realtime clock.
var System = sysClock{}
type sysClock struct{}
```
This exports a simple variable that we can then use as `clock.System`

core/signer/signer_test.go
```diff
@@ -0,0 +1,99 @@
...
+		got := 0
+		for i := 0; i < tc.nTicks; i++ {
+			force := <-enforce
+			if force {
+				got++
+			}
+		}
+
+		// Make sure we got the expected number of enforced epochs:
+		if got != tc.wantForced {
```
gdbelvin: Follow the standard go format:
`if got, want := forcedTicks, tc.wantForced; got != want {`

core/signer/signer.go
cesarghali: This is actually a good idea. Package `core/clock` can contain this. We can even put the fake clock there, e.g., `clock.Fake`. What do you think?

Liamsi: I would prefer to keep "real" packages clean from mocking/testing/fake objects if possible. But I also just found out that trillian already has such an [interface](https://github.com/google/trillian/blob/8b1388864d35cd200bf9dbf0278eb64472e5417b/util/timesource.go#L24) and it contains both a [SystemTimeSource](https://github.com/google/trillian/blob/8b1388864d35cd200bf9dbf0278eb64472e5417b/util/timesource.go#L30) as well as [FakeTimeSource](https://github.com/google/trillian/blob/8b1388864d35cd200bf9dbf0278eb64472e5417b/util/timesource.go#L39). Shouldn't we reuse this one instead?

core/signer/signer.go
```diff
@@ -66,56 +67,46 @@ func New(realm string,
...
+		rootResp, err = s.tmap.GetSignedMapRoot(ctx, &trillian.GetSignedMapRootRequest{
+			MapId: s.mapID,
+		})
+		if err != nil {
+			glog.Fatalf("GetSignedMapRoot failed after CreateEpoch: %v", err)
 		}
 	}
+	// Fetch last time from previous map head (as stored in the map server)
+	mapRoot := rootResp.GetMapRoot()
+	last := time.Unix(0, mapRoot.TimestampNanos)
```
gdbelvin: For protobuf objects, let's use a best practice of using the Get* operations. This will save us from a nil pointer dereference if we ever get one.

PR #599, opened 2017-06-05
Revert "Only create epochs if there is something to process or the maximum nu…"

cesarghali:
Reverts google/keytransparency#593


gdbelvin:
Looks like the tests which depend on time are too flakey for Travis. Let's come up with a time-independent testing strategy.

PR #602, opened 2017-06-09
Better auth error codes

gdbelvin:
Previously, the keyserver would return "Permission Denied" when an auth header was missing.  Now it returns "Unauthenticated". 

cc @AMarcedone

### DIFF
--- a/core/keyserver/keyserver.go
+++ b/core/keyserver/keyserver.go
@@ -226,9 +226,16 @@ func (s *Server) ListEntryHistory(ctx context.Context, in *tpb.ListEntryHistoryR
 // profile will be created.
 func (s *Server) UpdateEntry(ctx context.Context, in *tpb.UpdateEntryRequest) (*tpb.UpdateEntryResponse, error) {
 	// Validate proper authentication.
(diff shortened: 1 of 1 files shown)


core/keyserver/keyserver.go
```diff
@@ -226,9 +226,16 @@ func (s *Server) ListEntryHistory(ctx context.Context, in *tpb.ListEntryHistoryR
...
 	// Validate proper authentication.
-	if err := s.auth.ValidateCreds(ctx, in.UserId); err != nil {
-		glog.Warningf("Auth failed: %v", err)
+	switch err := s.auth.ValidateCreds(ctx, in.UserId); err {
+	case nil:
+		break // Authentication succeded.
+	case authentication.ErrWrongUser:
 		return nil, grpc.Errorf(codes.PermissionDenied, "Permission denied")
+	case authentication.ErrMissingAuth:
+		return nil, grpc.Errorf(codes.Unauthenticated, "Missing Auth Header")
```
Liamsi: s/Missing Auth Header/Missing authentication header 
or 
s/Missing Auth Header/Missing auth. header (?)

Issue #603, opened 2017-06-09
Add an optional flag to disable oauth authentication


AMarcedone:
When testing the client, it is annoying/slows me down to have to paste the oauth link on the browser to get a token and paste it in the client. Also it makes it hard to test modifications from multiple different users.

I propose to add this flag to the kt-server (frontend.go):
```go	
fakeAuth     = flag.Bool("insecure-fake-auth", false, "INSECURE! Do not user in production! Use a fake authentication instead of getting a google oauth token from clients.")
```
which would trigger the use of FakeAuth (from core/authentication/fake.go):
```go	
var auth authentication.Authenticator
	if *fakeAuth {
		glog.Warning("INSECURE! Using fake authentication.")
		auth = authentication.NewFake()
	} else {
		var err error
		auth, err = gauth.NewGoogleAuth()
		if err != nil {
			glog.Exitf("Failed to create authentication library instance: %v", err)
		}
	}
```
Similarly, I was thinking of adding a flag to the kt-client with which the user can pick an user-id to be used in the fake authentication:

```go	
RootCmd.PersistentFlags().String("fake-auth-userid","","userid to present to the server as identity for authentication. Only succeeds if fake auth is enabled on the server side.")
```

However, I am not very familiar with grpc so not sure how to send the user-id to the server as part of the context.

Any thoughts? Does it make sense or is it better not to add such a flag for security reasons?

gdbelvin:
Others have requested this feature as well. I think it's a good idea. 

On the outgoing call, here's how to explicitly set metadata:
https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md#sending-metadata-1 

It's probably best to adjust the fake auth to generate this for you.

AMarcedone:
Cool. I'll prepare a PR!

gdbelvin:
Closed via #605

PR #605, opened 2017-06-09
Add support for fake authentication

AMarcedone:
Adds support for fake authentication both to the client and the server (to be used for debug only). 

This allows the client to authenticate as user@domain.com by passing the additional parameter
`--fake-auth-userid=user@domain.com `
to the keytransparency-client binary (rather than having to get an actual oauth token). 
The feature should be enabled on the server side by starting the keytransparency-server binary with the ` --insecure-fake-auth ` flag.

### DIFF
--- a/cmd/keytransparency-client/cmd/post.go
+++ b/cmd/keytransparency-client/cmd/post.go
@@ -20,7 +20,6 @@ import (
 	"log"
 	"time"
 
-	"github.com/google/keytransparency/core/authentication"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	"golang.org/x/net/context"
@@ -86,12 +85,6 @@ User email MUST match the OAuth account used to authorize the update.
 		if err != nil {
 			return fmt.Errorf("updateKeys() failed: %v", err)
 		}
-
-		// Put fake authentication information in the context
-		if fakeUserId := viper.GetString("fake-auth-userid"); fakeUserId != "" {
-			ctx = authentication.InsertFakeAuthIntoContext(ctx,fakeUserId)
-		}
-
 		// TODO: fill signers and authorizedKeys.
 		if _, err := c.Update(ctx, userID, appID, profileData, signers, authorizedKeys); err != nil {
 			return fmt.Errorf("update failed: %v", err)
--- a/cmd/keytransparency-client/cmd/post.go
+++ b/cmd/keytransparency-client/cmd/post.go
@@ -20,6 +20,7 @@ import (
 	"log"
 	"time"
 
+	"github.com/google/keytransparency/core/authentication"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	"golang.org/x/net/context"
@@ -85,6 +86,12 @@ User email MUST match the OAuth account used to authorize the update.
 		if err != nil {
 			return fmt.Errorf("updateKeys() failed: %v", err)
 		}
+
+		// Put fake authentication information in the context
+		if fakeUserId := viper.GetString("fake-auth-userid"); fakeUserId != "" {
+			ctx = authentication.InsertFakeAuthIntoContext(ctx,fakeUserId)
+		}
+
 		// TODO: fill signers and authorizedKeys.
 		if _, err := c.Update(ctx, userID, appID, profileData, signers, authorizedKeys); err != nil {
 			return fmt.Errorf("update failed: %v", err)
--- a/cmd/keytransparency-client/cmd/root.go
+++ b/cmd/keytransparency-client/cmd/root.go
@@ -23,12 +23,14 @@ import (
 	"time"
 
 	"github.com/google/keytransparency/cmd/keytransparency-client/grpcc"
+	"github.com/google/keytransparency/core/authentication"
 	"github.com/google/keytransparency/core/client/kt"
 	"github.com/google/keytransparency/core/crypto/keymaster"
 	"github.com/google/keytransparency/core/crypto/signatures"
 	"github.com/google/keytransparency/core/crypto/vrf"
 	"github.com/google/keytransparency/core/crypto/vrf/p256"
-	"github.com/google/keytransparency/impl/google/authentication"
+	gauth "github.com/google/keytransparency/impl/google/authentication"
+	pb "github.com/google/keytransparency/impl/proto/keytransparency_v1_service"
 
 	"github.com/google/trillian/client"
 	"github.com/google/trillian/crypto/keys"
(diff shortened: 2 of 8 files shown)


AMarcedone:
Closes #603

cmd/keytransparency-client/cmd/post.go
```diff
@@ -20,6 +20,7 @@ import (
 	"log"
 	"time"
 
+	"github.com/google/keytransparency/core/authentication"
```
gdbelvin: For consistency, please add a new line here.  We're organizing our imports as follows:

```go
import(
    "golang packages"
   
     "keytransparency packages"

     "other packages"
)
```

cmd/keytransparency-client/cmd/root.go
```diff
@@ -217,22 +219,32 @@ func dial(ktURL, caFile, clientSecretFile string, serviceKeyFile string) (*grpc.
...
-		}
-		opts = append(opts, grpc.WithPerRPCCredentials(creds))
-	} else if serviceKeyFile != "" {
-		creds, err := getServiceCreds(serviceKeyFile)
-		if err != nil {
-			return nil, err
+	// Add authentication information for the grpc.
+	fakeUserId := viper.GetString("fake-auth-userid")
+	if fakeUserId != "" { // Using fake authentication.
+		// Fake authentication information are added in cmd/keytransparency-client/cmd/post.go
```
gdbelvin: It would be nice if the new fake auth implemented the [`credentials.PerRPCCredentials`](https://github.com/grpc/grpc-go/blob/master/credentials/credentials.go) interface so it could match the other methods of adding authenitcation to the connection.

cmd/keytransparency-client/cmd/root.go
```diff
@@ -217,22 +219,32 @@ func dial(ktURL, caFile, clientSecretFile string, serviceKeyFile string) (*grpc.
...
+		// TODO(amarcedone) Consider having the GetClient method (or this method) initialize
+		// the context so that we can set the metadata parameters for the fake authentication
+		// here rather than in post.go.
+	} else {
```
gdbelvin: This set of if/else's could be replaced with a `switch`, one case for each type of authentication.

cmd/keytransparency-server/frontend.go
```diff
@@ -140,9 +143,11 @@ func jsonLogger(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo
 		return handler(ctx, req)
 	}
 	glog.Infof("%v>%v", requestCounter, s)
+	*/
 
 	resp, err = handler(ctx, req)
 
+	/* TODO(amarcedone): logger not working. commenting out.
```
gdbelvin: Just don't use the --verbose flag.  Does this produce a compile time error?

cmd/keytransparency-server/frontend.go
```diff
@@ -183,9 +188,17 @@ func main() {
 	if err != nil {
 		glog.Exitf("Failed to load server credentials %v", err)
 	}
-	auth, err := authentication.NewGoogleAuth()
-	if err != nil {
-		glog.Exitf("Failed to create authentication library instance: %v", err)
+
+	var auth authentication.Authenticator
+	if *fakeAuth {
```
gdbelvin: A switch here would be more appropriate.  I imagine we'll support multiple types of authentication in the future. 

Perhaps the flag could also be modified to be `auth-type=google` by default and `auth-type=fake` for testing.

core/authentication/fake.go
```diff
@@ -31,24 +30,30 @@ func NewFake() *FakeAuth {
...
 // NewContext adds authentication details to a new background context.
 func (a *FakeAuth) NewContext(userID string) context.Context {
-	md := make(map[string][]string)
-	md["userid"] = []string{userID}
-	return metadata.NewOutgoingContext(context.TODO(), md)
+	return InsertFakeAuthIntoContext(context.TODO(), userID)
+}
+
+// Includes (fake) authentication information in the context to authenticate to the server as userID.
+func InsertFakeAuthIntoContext(ctx context.Context, userID string) context.Context {
```
gdbelvin: Try making `FakeAuth` implement the `credentials.PerRPCCredentials` interface. 
That way, it could work seamlessly with the other GRPC authentication libraries.

cmd/keytransparency-server/frontend.go
AMarcedone: Good idea =)

AMarcedone:
I updated the PR with the requested changes.

Please note that in this commit https://github.com/google/keytransparency/pull/605/commits/3a12d1e58b26477515e3e747764d53cf071a3796
I removed the authentication context because (strangely) the test was not using it (even though it seemed to claim so as there was one test case which was tagged "Incorrect auth").

cmd/keytransparency-client/cmd/root.go
AMarcedone: The description of the interface is not very clear, but I looked at the existing implementations and hacked something similar :)

core/authentication/fake.go
```diff
@@ -29,26 +33,50 @@ func NewFake() *FakeAuth {
 	return &FakeAuth{}
 }
 
-// NewContext adds authentication details to a new background context.
-func (a *FakeAuth) NewContext(userID string) context.Context {
-	md := make(map[string][]string)
-	md["userid"] = []string{userID}
-	return metadata.NewOutgoingContext(context.TODO(), md)
+func GetFakeCredential(userID string) credentials.PerRPCCredentials{
```
gdbelvin: For code clarity, would you mind moving this function below `FakeCredential`? 
We like to keep structs and their associated functions close together. Thanks!

core/authentication/fake.go
```diff
@@ -29,26 +33,50 @@ func NewFake() *FakeAuth {
...
+
+	if got, want := p[1], requiredUserID; got != want {
+		glog.V(2).Infof("FakeAuth: wrong user. got: %v, want %v", got, want)
 		return ErrWrongUser
 	}
+	glog.V(2).Infof("FakeAuth: fake authentication suceeded for user %+v", requiredUserID)
 	return nil
 }
+
+const fakeCredentialType string = "FakeCredential"
```
gdbelvin: Please move costs to the top of the file

AMarcedone:
Thanks a lot! I fixed an integration test which was failing and run gofmt on the files I touched :)

gdbelvin:
Thanks! LGTM and merged.

gdbelvin:
@jmillican This might be of interest to you

PR #607, opened 2017-06-12
Deployment to gce for continuous integration

### LINKED ISSUE #601: Continuous Integration Server
It would be awesome to have a server that tracked HEAD

A few options:
https://github.com/GoogleCloudPlatform/continuous-deployment-on-kubernetes

Docker hub seems to be down:
https://github.com/docker/hub-feedback/issues/590
https://docs.docker.com/docker-hub/github/#github-organizations

### LINKED ISSUE #615: max-period=1d is not an acceptable duration
In keytransparency-signer/Dockerfile, MAX_SIGN_PERIOD=1d which is not an acceptable duration
(https://golang.org/pkg/time/#ParseDuration). Suggest replacing with 24h.

Liamsi:
~~Work in progress PR created early to trigger travis (on the original repo, not my fork)
Will eventually resolve #601~~

Added kubernetes configuration and a deploy script which:

1. Builds all docker images
2. cleans up the default namespace in a particular gcloud project (delete all services, pods/deployments)
3. pushes all trillian images to the project's container registry
4. Starts a trillian deployment on gce
5. pushes all keytransparency docker images to gcloud
6. creates a log-/map-tree and sets the IDs in the keytransparency kubernetes file as environment vars
7. runs keytransparency in gce using kubernetes 

This script can be used locally or by travis. Travis will run the script every time after pull gets merged into master.

Resolves #601
Also resolves #615


.travis.yml
```diff
@@ -4,8 +4,23 @@ go:
...
-sudo: false
+sudo: required
 
+services:
+  - docker
+cache:
+  directories:
+    - "$HOME/google-cloud-sdk/"
+env:
+  - DOCKER_COMPOSE_VERSION="1.13.0" PATH=$PATH:${HOME}/google-cloud-sdk/bin CLOUDSDK_CORE_DISABLE_PROMPTS=1 PROJECT_NAME=transparent-keyserver
```
gdbelvin: Split the ENV variables onto separate lines if possible

.travis.yml
```diff
@@ -4,8 +4,23 @@ go:
   - 1.8
 
 dist: trusty
-sudo: false
+sudo: required
```
gdbelvin: From the travis docs, it looks like we can set sudo to false and just get the docker binaries. 

Docker is installed. We use the official Docker apt repository, so you can easily install another version with apt if needed. When sudo: required is specified, Docker is installed as a service. When sudo: false is specified, the Docker binaries are available for local use.

.travis.yml
```diff
@@ -4,8 +4,23 @@ go:
...
 
+services:
+  - docker
+cache:
+  directories:
+    - "$HOME/google-cloud-sdk/"
+env:
+  - DOCKER_COMPOSE_VERSION="1.13.0" PATH=$PATH:${HOME}/google-cloud-sdk/bin CLOUDSDK_CORE_DISABLE_PROMPTS=1 PROJECT_NAME=transparent-keyserver
+before_install:
+  - sudo rm /usr/local/bin/docker-compose
```
gdbelvin: Could you try using the travis apt addon to install the latest version?

.travis.yml
```diff
@@ -17,7 +32,20 @@ install:
 script:
   - make
   - make travis-presubmit
-  - ./coverage.sh 
+  - ./coverage.sh
 
 after_success:
   - bash <(curl -s https://codecov.io/bash)
+  # The last two commands are only needed for the deploy step (which will only
```
gdbelvin: If this is only needed in the deploy step, could it moved to the deploy step?

.travis.yml
```diff
@@ -17,7 +32,20 @@ install:
...
+  # run on master. The first fails on PRs from forks (but the build passes), the
+  # 2nd only does sth. if the cache is empty.
+  - openssl aes-256-cbc -K $encrypted_a1eb99cfc21e_key -iv $encrypted_a1eb99cfc21e_iv -in credentials.tar.gz.enc -out credentials.tar.gz -d
+  - if [ ! -d "$HOME/google-cloud-sdk/bin" ]; then rm -rf $HOME/google-cloud-sdk; curl https://sdk.cloud.google.com | bash; fi
+  # needed for envsubst command used in deploy script:
+  - sudo apt-get install gettext
```
gdbelvin: install via apt addon

cmd/keytransparency-server/Dockerfile
```diff
@@ -18,8 +18,9 @@ ENV MAP_ID=0 \
 ENV LOG_ID=0 \
     LOG_URL=localhost:8090 
 ENV GOOGLE_APPLICATION_CREDENTIALS=genfiles/service_key.json
-ENV VERBOSE=0
```
gdbelvin: We're deleting the code behind verbose, I think...  leave this as 0

.travis.yml
```diff
@@ -17,7 +32,20 @@ install:
...
   - make travis-presubmit
-  - ./coverage.sh 
+  - ./coverage.sh
 
 after_success:
   - bash <(curl -s https://codecov.io/bash)
+  # The last two commands are only needed for the deploy step (which will only
+  # run on master. The first fails on PRs from forks (but the build passes), the
+  # 2nd only does sth. if the cache is empty.
+  - openssl aes-256-cbc -K $encrypted_a1eb99cfc21e_key -iv $encrypted_a1eb99cfc21e_iv -in credentials.tar.gz.enc -out credentials.tar.gz -d
```
gdbelvin: Please pick a more descriptive name for the credential file

docker-compose.yml
```diff
@@ -45,13 +43,6 @@ services:
       DB_DATABASE: test
       DB_USER: test
       DB_PASSWORD: zaphod
-      GOOGLE_APPLICATION_CREDENTIALS: /genfiles/service_key.json
-      KEY: /kt/server.key
```
gdbelvin: leave these. It's better to have them explicitly set

kubernetes/keytransparency-deployment.yml
```diff
@@ -0,0 +1,98 @@
...
+  strategy:
+    type: Recreate
+  template:
+    metadata:
+      labels:
+        run: kt-server
+    spec:
+      containers:
+      - name: kt-server
+        image: us.gcr.io/transparent-keyserver/keytransparency-server
```
gdbelvin: this should be keytransparency or key-transparency

kubernetes/keytransparency-deployment.yml
```diff
@@ -0,0 +1,98 @@
...
+        ports:
+          - containerPort: 8080
+            name: json-grpc
+        env:
+            - name: LOG_URL
```
gdbelvin: can `name` and `value` be put on the same line?

.travis.yml
Liamsi: If you split them in several lines/items one build will be triggered per line: https://docs.travis-ci.com/user/environment-variables/#Defining-public-variables-in-.travis.yml
But it seems
```
env:
  global:
     # separate line per VAR
``` 
does what we want. Will do.

kubernetes/keytransparency-deployment.yml
Liamsi: This seems to be the default (as in the [documentation](https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/)). But I've tried: 
` - name: LOG_URL value: trillian-log:8090` 
and get an error: `error: error converting YAML to JSON: yaml: line 24: mapping values are not allowed in this context
`

.travis.yml
Liamsi: `docker build` needs the docker daemon to be running. Unfortunately, it looks like we can not do without `sudo: required`.

.travis.yml
Liamsi: Switched to using pip now which is less hacky (apt doesn't work for docker-compose).

.travis.yml
Liamsi: done

.travis.yml
Liamsi: Switched to `sed` as you suggested (thanks).

cmd/keytransparency-server/Dockerfile
Liamsi: done

docker-compose.yml
Liamsi: The trillian-log-signer doesn't need them. Copied from its Dockerfile:
```
 /go/bin/trillian_log_signer \
	--mysql_uri="${DB_USER}:${DB_PASSWORD}@tcp(${DB_HOST})/${DB_DATABASE}" \
	--http_endpoint="$HOST:$HTTP_PORT" \
	--sequencer_guard_window="$SEQUENCER_GUARD_WINDOW" \
	--force_master="$FORCE_MASTER" \
	--alsologtostderr
```

kubernetes/keytransparency-deployment.yml
Liamsi: changed

.travis.yml
Liamsi: Changed to: `client_secret-and-service_key.tar.gz`

Liamsi:
Comments addressed PTAL

.travis.yml
Liamsi: The only viable alternative I've found: Instead of running `docker-compose build` on travis (and therefore needing sudo to run the daemon), we could get rid of docker-compose, run only `docker build` instead and do this in gcloud instead of travis with `gcloud docker -- build <image-to-be-build>`. Se: https://cloud.google.com/sdk/gcloud/reference/docker
This could be faster, too.

kubernetes/keytransparency-deployment.yml
Liamsi: I've created a new service-key in this project but authenticating using this key does not grant the permissions to create a repository for the container registry there. Either one of the owners (@gdbevlin, @cesarghali)  need to create a key-file, download it and encrypt it after we merged this pull or you need to grant me access to do so.

.travis.yml
```diff
@@ -22,7 +32,23 @@ install:
...
+    - sudo apt-get update
+    - sudo apt-get -y -o Dpkg::Options::="--force-confnew" install docker-engine
+    - pip install docker-compose
+    - docker --version
+    - docker-compose --version
+    - if [ ! -d "$HOME/google-cloud-sdk/bin" ]; then rm -rf $HOME/google-cloud-sdk;
+      curl https://sdk.cloud.google.com | bash; fi
+    - openssl aes-256-cbc -K $encrypted_a1eb99cfc21e_key -iv $encrypted_a1eb99cfc21e_iv -in service_key.tar.gz.enc -out service_key.tar.gz -d
+    - tar -xzf service_key.tar.gz
+    - (cd scripts && ./deploy.sh)
```
gdbelvin: should this be `./scripts/deploy.sh` ?

.travis.yml
```diff
@@ -22,7 +32,23 @@ install:
...
+  provider: script
+  script:
+    - sudo apt-get update
+    - sudo apt-get -y -o Dpkg::Options::="--force-confnew" install docker-engine
+    - pip install docker-compose
+    - docker --version
+    - docker-compose --version
+    - if [ ! -d "$HOME/google-cloud-sdk/bin" ]; then rm -rf $HOME/google-cloud-sdk;
+      curl https://sdk.cloud.google.com | bash; fi
+    - openssl aes-256-cbc -K $encrypted_a1eb99cfc21e_key -iv $encrypted_a1eb99cfc21e_iv -in service_key.tar.gz.enc -out service_key.tar.gz -d
```
gdbelvin: Double check that `service_key.tar.gz.enc` is still correct.  It doesn't look like it matches with the `.enc` file in the PR.  

Consider naming this `travis_secrets.tar.gz.enc`

cmd/keytransparency-signer/Dockerfile
```diff
@@ -12,9 +12,11 @@ ENV LOG_ID=0 \
     LOG_KEY=../trillian/testdata/log-rpc-server.pubkey.pem
 
 ENV MIN_SIGN_PERIOD=5s \
-    MAX_SIGN_PERIOD=1d
+    MAX_SIGN_PERIOD=24h
```
gdbelvin: thanks

cmd/keytransparency-signer/Dockerfile
```diff
@@ -12,9 +12,11 @@ ENV LOG_ID=0 \
...
 
-ADD . /go/src/github.com/google/keytransparency
+ADD trillian/testdata/* /trillian/
+
```
gdbelvin: remove the newline

kubernetes/README.md
```diff
@@ -0,0 +1,8 @@
+# Usage
+
+To spin up a minimalistic deployment of keytransparency on 
+[Google Compute Engine](https://cloud.google.com/sdk/gcloud/), run the following 
+command to create, upload, and run all containers:
+```
+cd scripts && ./deploy.sh
```
gdbelvin: Is it possible to remove the cd?

kubernetes/mysql-trillian/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM mysql:5.7
```
gdbelvin: What do you think about moving these files to 
`deploy/mysql-trillian`
`deploy/kubernetes`

?

scripts/deploy.sh
```diff
@@ -0,0 +1,139 @@
+#!/usr/bin/env bash
+
+################################################################################
+# Following assumptions are made by this script:                               #
+# * gcloud, docker, and docker-compose is installed                            #
+# * it is called from $GOPATH/src/github.com/google/keytransparency/scripts    #
```
gdbelvin: Could we it's assumption to be called from the root directory
` $GOPATH/src/github.com/google/keytransparency` ?

.gitignore
```diff
@@ -10,3 +10,7 @@ infra*.etcd
 genfiles/*
 .keytransparency-scts.dat
 .keystore
+credentials.tar.gz
```
gdbelvin: Update filename

scripts/deploy.sh
```diff
@@ -0,0 +1,139 @@
...
+
+function buildDockerImgs()
+{
+  # Build all images defined in the docker-compose.yml:
+  (cd ../ && docker-compose build)
+  # Separately build the DB:
+  # Work around some git permission issues on linux:
+  chmod a+r ../../trillian/storage/mysql/storage.sql
+  # Create and push a fresh db image (with schema created)
+  docker build -t  us.gcr.io/${PROJECT_NAME}/db -f ../kubernetes/mysql-trillian/Dockerfile ../..
```
gdbelvin: See if you can get docker-compose to do this for you. 
- Set the tag
- it's ok to hardcode the project 
- See if you can publish the containers

scripts/deploy.sh
```diff
@@ -0,0 +1,139 @@
...
+}
+
+function pushTrillianImgs()
+{
+  gcloud docker -- push us.gcr.io/${PROJECT_NAME}/db
+  images=("trillian_log_server" "trillian_map_server" "keytransparency-server" \
+  "trillian_log_signer" "keytransparency-signer")
+  for DOCKER_IMAGE_NAME in "${images[@]}"
+  do
+    docker tag ${DOCKER_IMAGE_NAME} us.gcr.io/${PROJECT_NAME}/${DOCKER_IMAGE_NAME}
```
gdbelvin: If docker-compose tags things we might not need to do this here.

scripts/deploy.sh
```diff
@@ -0,0 +1,139 @@
...
+
+  pushKTImgs
+  waitForTrillian
+  createTreeAndSetIDs
+
+  # Deploy all keytransparency related services (server and signer):
+  kubectl apply -f ../kubernetes/keytransparency-deployment.yml
+}
+
+function initGcloud()
```
gdbelvin: Try doing this in before_deploy:
https://medium.com/google-cloud/continuous-delivery-in-a-microservice-infrastructure-with-google-container-engine-docker-and-fb9772e81da7

scripts/deploy.sh
```diff
@@ -0,0 +1,139 @@
...
+function pushTrillianImgs()
+{
+  gcloud docker -- push us.gcr.io/${PROJECT_NAME}/db
+  images=("trillian_log_server" "trillian_map_server" "keytransparency-server" \
+  "trillian_log_signer" "keytransparency-signer")
+  for DOCKER_IMAGE_NAME in "${images[@]}"
+  do
+    docker tag ${DOCKER_IMAGE_NAME} us.gcr.io/${PROJECT_NAME}/${DOCKER_IMAGE_NAME}
+    # Push the images as we refer to them in the kubernetes config files:
+    gcloud docker -- push us.gcr.io/${PROJECT_NAME}/${DOCKER_IMAGE_NAME}
```
gdbelvin: docker-compose push?
https://docs.docker.com/compose/reference/push/

scripts/deploy.sh
```diff
@@ -0,0 +1,139 @@
...
+  done
+}
+
+function pushKTImgs()
```
gdbelvin: Do we need to have a separate push command for KT? 
Is it possible to just push everything together?

scripts/deploy.sh
```diff
@@ -0,0 +1,139 @@
...
+  LOG_ID=""
+  MAP_ID=""
+  COUNTER=0
+  until [ -n "$LOG_ID" ] || [  $COUNTER -gt $MAX_RETRY ]; do
+    # RPC was not available yet, wait and retry:
+    sleep 10;
+    let COUNTER+=1
+    # get the currentl trillian-map pod:
+    MAPSRV=$(kubectl get pods --selector=run=trillian-map -o jsonpath={.items[*].metadata.name});
+    LOG_ID=$(echo 'go run $GOPATH/src/github.com/google/trillian/cmd/createtree/main.go --admin_server=localhost:8090 --pem_key_path=testdata/log-rpc-server.privkey.pem --pem_key_password="towel" --signature_algorithm=ECDSA --tree_type=LOG' | kubectl exec -i $MAPSRV -- /bin/sh )
```
gdbelvin: is localhost going to be a problem?

Liamsi: No, the command will be run directly on the pod (at the very end: ` kubectl exec -i $MAPSRV -- /bin/sh`
But we can change to `trillian-map` for consistency, if you prefer (which will resolve to the same machine as `localhost`).

scripts/deploy.sh
Liamsi: Doesn't look like docker-compose is supported by `gcloud`.

scripts/deploy.sh
Liamsi: As discussed: As we need to wait for the trillian services to be up and running, we can as well spent this time doing sth. we need to do anyways.

scripts/deploy.sh
Liamsi: All done besides publishing. `gcloud` doesn't seem to support `docker-compose`. What we could do though, is make docker login into gcloud: https://stackoverflow.com/a/29311300 
Do you think that's safe and worth trying?

scripts/deploy.sh
Liamsi: done (docker-compose now tags with us.gcr.io/key-transparency/<image-name>)

kubernetes/mysql-trillian/Dockerfile
Liamsi: done

scripts/deploy.sh
Liamsi: done

.gitignore
Liamsi: done

scripts/deploy.sh
Liamsi: done

kubernetes/README.md
Liamsi: done

cmd/keytransparency-signer/Dockerfile
Liamsi: done

.travis.yml
Liamsi: done

.travis.yml
Liamsi: done

Liamsi:
Adressed all comments, and tested the deploy script locally. PTAL

PR #611, opened 2017-06-16
Use gnu parallel for faster tests

gdbelvin:
This PR reduces build times from 30 min to 2 minutes. 
https://scribu.github.io/travis-stats/#google/keytransparency/master

### DIFF
--- a/coverage.sh
+++ b/coverage.sh
@@ -1,12 +1,6 @@
 #!/usr/bin/env bash
 
 set -e
-echo "" > coverage.txt
-
-for d in $(go list ./... | grep -v vendor); do
-    go test -race -coverprofile=profile.out -covermode=atomic $d
-    if [ -f profile.out ]; then
-        cat profile.out >> coverage.txt
-        rm profile.out
-    fi
-done
+mkdir -p out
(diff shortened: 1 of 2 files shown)


coverage.sh
```diff
@@ -1,12 +1,5 @@
...
-echo "" > coverage.txt
-
-for d in $(go list ./... | grep -v vendor); do
-    go test -race -coverprofile=profile.out -covermode=atomic $d
-    if [ -f profile.out ]; then
-        cat profile.out >> coverage.txt
-        rm profile.out
-    fi
-done
+go list ./... | grep -v vendor | parallel -k go test -race -coverprofile=out/{#} -covermode=atomic {}
```
pphaneuf: I think you may find that `xargs` with the help of appropriate `-P` and `-n` flags might do the trick, and be more portable?

gdbelvin: Thanks for the suggestion!

gdbelvin: On further research, parallel is actually better because it schedules jobs on cpus efficiently that fully utilizes the available resources without starving the host machine of cpu. 

It's also possible to preserve the `nosudo` flag by utilizing travis's apt addon so we can keep using the faster containerized build infrastructure.

PR #616, opened 2017-06-19
Bug fix: replace impossible "&&" with "||" in VRF test.

alsophian:
I found this bug while working on the JavaScript client.

### DIFF
--- a/core/crypto/vrf/p256/p256_test.go
+++ b/core/crypto/vrf/p256/p256_test.go
@@ -45,7 +45,7 @@ func TestH2(t *testing.T) {
 			t.Fatalf("Failed generating random message: %v", err)
 		}
 		x := H2(m)
-		if got := len(x.Bytes()); got < 1 && got > l {
+		if got := len(x.Bytes()); got < 1 || got > l {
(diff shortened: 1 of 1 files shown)


core/crypto/vrf/p256/p256_test.go
```diff
@@ -45,7 +45,7 @@ func TestH2(t *testing.T) {
 			t.Fatalf("Failed generating random message: %v", err)
 		}
 		x := H2(m)
-		if got := len(x.Bytes()); got < 1 && got > l {
+		if got := len(x.Bytes()); got < 1 || got > l {
```
Liamsi: Couldn't this be simplified to `if got := len(x.Bytes()); got != l {`?

gdbelvin: `H2` returns a number between 1 and N-1 which means `x.Bytes()` can also range from 1 to 32.

Liamsi: Ok. I see. But then the error message below should be different from:
`t.Errorf("len(h2(%v)) = %v, want %v", m, got, l)`
because this will always say `want 32` 
and not "a number between 1 and N-1"

PR #619, opened 2017-06-20
Another place where the name of the credentials file should be adapted

Liamsi:
Replace `tar -xzf service_key.tar.gz` with `tar -xzf travis_secrets.tar.gz`
to make travis pass (hopefully): https://travis-ci.org/google/keytransparency/builds/244962734#L5015-L5022


RJPercival:
It doesn't look like Travis did anything Docker-related...

Liamsi:
@RJPercival: Travis couldn't reach the point where it does anything docker related because of the error this pull fixed. The deploy script only ran locally so far. I suspect a few more lines that need to be changed here and there.

PR #631, opened 2017-06-20
Try gcloud update components kubectl without sudo

### LINKED ISSUE #630: travis still fails on deploy step
`kubectl` is not in the `PATH`:
https://travis-ci.org/google/keytransparency/builds/245102518#L5130
and something seems wrong with the `gcloud` installation. 
My guess is that part of it is in `$HOME/google-cloud-sdk/bin` and part of it seems to be in 
`/usr/lib/google-cloud-sdk`. See:
https://travis-ci.org/google/keytransparency/builds/245102518#L5866-L5885

Retry `gcloud --quiet components update kubectl` *without* sudo. Now that we `source $HOME/google-cloud-sdk/path.bash.inc` it might work without?

Liamsi:
See if `gcloud --quiet components update kubectl` works properly now that we `source $HOME/google-cloud-sdk/path.bash.inc`
Also, move all gcloud initialization to travis.yml.

resolves #630 (hopefully)

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -4,10 +4,13 @@
 # Following assumptions are made by this script:                               #
 # * gcloud, docker, and docker-compose is installed                            #
 # * it is called from $GOPATH/src/github.com/google/keytransparency            #
-# * there is a project called key-transparency on gce which has a cluster      #
-#   called "ci-cluster" within the "us-central1-a" compute zone                #
-# * there is a service key to authenticate with glcoud to above project in     #
-#   $GOPATH/src/github.com/google/keytransparency/service_key.json             #
+# * there is a project called key-transparency on gce  which has has gke       #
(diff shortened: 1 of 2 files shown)


scripts/deploy.sh
```diff
@@ -4,10 +4,13 @@
 # Following assumptions are made by this script:                               #
 # * gcloud, docker, and docker-compose is installed                            #
 # * it is called from $GOPATH/src/github.com/google/keytransparency            #
-# * there is a project called key-transparency on gce which has a cluster      #
-#   called "ci-cluster" within the "us-central1-a" compute zone                #
-# * there is a service key to authenticate with glcoud to above project in     #
-#   $GOPATH/src/github.com/google/keytransparency/service_key.json             #
+# * there is a project called key-transparency on gce  which has has gke       #
```
cesarghali: s/gce  which/gce which (one space).

PR #633, opened 2017-06-21
Fix order of gcloud commands (tested)

### LINKED ISSUE #632: Fix order of gcloud commands (travis)
After the last merge `gcloud` complains that `Accessing a Container Engine cluster requires the kubernetes commandline client [kubectl` (build [logs](https://travis-ci.org/google/keytransparency/builds/245176876)). 
Make sure `kubectl` gets installed before any depending `gcloud` command is run. 

This issue can be tested in a separate repo (to avoid more failing builds on master). Alternatively, there are docker images available which can be used to try commands in the exact same environment as travis.

Liamsi:
Resolves #632 (tested here: https://travis-ci.org/Liamsi/travis-kubernets-gcloud-test/builds/245320574)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,10 +52,10 @@ before_deploy:
   - gcloud auth activate-service-account --key-file service_key.json
   - rm -f service_key.json
   - gcloud --quiet version
+  - gcloud --quiet components update kubectl
   - gcloud config set project key-transparency
-  - gcloud container clusters get-credentials ci-cluster
   - gcloud config set compute/zone us-central1-a
-  - gcloud --quiet components update kubectl && gcloud info --show-log
+  - gcloud container clusters get-credentials ci-cluster
 
 deploy:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -52,10 +52,10 @@ before_deploy:
   - gcloud auth activate-service-account --key-file service_key.json
   - rm -f service_key.json
   - gcloud --quiet version
+  - gcloud --quiet components update kubectl
   - gcloud config set project key-transparency
-  - gcloud container clusters get-credentials ci-cluster
   - gcloud config set compute/zone us-central1-a
-  - gcloud --quiet components update kubectl && gcloud info --show-log
```
gdbelvin: does this fix the error related to 
```
gcloud config set compute/zone VALUE
```

Liamsi: Yes, the compute zone needs to be set before setting the cluster.

PR #641, opened 2017-06-21
Update credentials file to contain another service-key for the server

### LINKED ISSUE #640: Successful deployment: Error in kt-server log
The deployed kt-server can't find a proper `GOOGLE_APPLICATION_CREDENTIALS` credentials file.

Investigate if the recently added fake-auth (#605) would be a valid alternative to adding another service key for the server (with minimal permissions), or, if we can get with the service-key completely.

**Update:** We can get rid of the additional service-key for the server!

Liamsi:
include another service-key for the kt server

resolves #640


Liamsi:
Closing: This is resolved by #644

PR #644, opened 2017-06-22
Remove service-key requirement from kt-server

### LINKED ISSUE #617: make running docker-compose easier, ideally without user interaction
It's quite time consuming to copy & paste the LOG_ID and MAP_ID created by the create-tree command (at least if done over and over). 
I would propose wrap this in a script, or, at least export the environment vars and replace the [example IDs in the docker-compose](https://github.com/google/keytransparency/blob/2128c1d504b0b993eb7f6ec6ee4f97edbb00ae41/docker-compose.yml#L100-L102) file with environment variables. The user could then just launch docker-compose  up as described in the Readme without needing to copy & paste the ouput of the create tree command.

### LINKED ISSUE #640: Successful deployment: Error in kt-server log

Liamsi:
Use a default http client to query https://www.googleapis.com/oauth2/v2

Also, use environment vars for `LOG_ID` and `MAP_ID` in the `docker-compose.yml` file (makes local testing/debugging a easier/more convenient).

Resolves #640 
Resolves #617

### DIFF
--- a/.keytransparency.yaml
+++ b/.keytransparency.yaml
@@ -2,7 +2,8 @@ log-key: "../trillian/testdata/log-rpc-server.pubkey.pem"
 vrf:    "genfiles/vrf-pubkey.pem"
 kt-key: "genfiles/server.crt"
 kt-sig: "genfiles/p256-pubkey.pem"
-map-id: 3506304676031956248
-kt-url: "localhost:8080"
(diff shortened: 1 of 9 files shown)


Liamsi:
PTAL

PR #656, opened 2017-06-26
Update Readme

gdbelvin:


README.md
```diff
@@ -73,56 +62,44 @@ Set `$GOPATH` variable to point to your Go workspace directory and add `$GOPATH/
...
-  The tool will build the server binaries, generate keys, and configure the server.
-  Clients will need the following public keys in order to verify server responses:
+### Install 
+1. [OpenSSL](https://www.openssl.org/community/binaries.html)
+1. [Docker](https://docs.docker.com/engine/installation/) 
+   - Docker Engine 1.13.0+ `docker version -f '{{.Server.APIVersion}}'`
+   - Docker Compose 1.11.0+ `docker-compose --version`
+1. `go get -u github.com/google/keytransparency/...`
+1. `go get -u github.com/google/trillian/...`
+1. `./scripts/prepare_server.sh -f`
```
Liamsi: Should we add `jq` as a dependency?

README.md
```diff
@@ -73,56 +62,44 @@ Set `$GOPATH` variable to point to your Go workspace directory and add `$GOPATH/
...
-  ```
+## Running the server
 
-  The tool will build the server binaries, generate keys, and configure the server.
-  Clients will need the following public keys in order to verify server responses:
+### Install 
+1. [OpenSSL](https://www.openssl.org/community/binaries.html)
+1. [Docker](https://docs.docker.com/engine/installation/) 
+   - Docker Engine 1.13.0+ `docker version -f '{{.Server.APIVersion}}'`
+   - Docker Compose 1.11.0+ `docker-compose --version`
```
Liamsi: Should we add `jq` as a dependency? (It is installed on travis by default and the script which is referred to for creating the trees will tell the user that he needs to install it anyways.)

gdbelvin: yep

PR #669, opened 2017-06-28
Make sql files executable

gdbelvin:


deploy/mysql-trillian/Dockerfile
```diff
@@ -3,4 +3,7 @@ FROM mysql:5.7
 # expects the build context to be: $GOPATH/src/github.com/google/ (as both,
 # trillian and keytransparency, reside there and docker can find the storage.sql
 # file)
-COPY trillian/storage/mysql/storage.sql /docker-entrypoint-initdb.d/storage.sql
\ No newline at end of file
+COPY trillian/storage/mysql/storage.sql /docker-entrypoint-initdb.d/storage.sql
+RUN chmod -R 775 /docker-entrypoint-initdb.d
```
Liamsi: It seems on travis read permissions to everyone (`chmod a+r`) are sufficient. See [here](https://github.com/google/keytransparency/blob/f444979f2d4f9f54223a16f754ef4f7b17da570a/scripts/deploy.sh#L56).
On the other hand the `chmod` only happens inside the container. I would say that is OK!

PR #670, opened 2017-06-28
Autom. expose ports and an external IP for kt-server and Prometheus

### LINKED ISSUE #621: Publish Continuous Integration IP
Perhaps it would be useful to have the CI server available for people to play with. 
Any data put in it will be deleted with every new push to MASTER.

### LINKED ISSUE #658: failed to read: ../trillian/testdata/ct-http-server.pubkey.pem.
"error connecting: Failed to open public key <nil>: failed to read: ../trillian/testdata/ct-http-server.pubkey.pem. open ../trillian/testdata/ct-http-server.pubkey.pem: no such file or directory"

the above error happens when I execute the following command as suggested in the Using the Key Transparency Client seciton.
./keytransparency-client post user@domain.com app1 -d 'dGVzdA==' --config=./.keytransparency.yaml

also, from the instruction "3. Get an OAuth client ID and download the generated JSON file."  I can't get how and where to use this generated JSON file, a more detailed instruction will be helpful?

Liamsi:
Also, make `MIN_SIGN_PERIOD` and `MAX_SIGN_PERIOD` explicit in the kubernetes conf

resolves #621
also resolves #658 (log key not found)

### DIFF
--- a/deploy/kubernetes/keytransparency-deployment.yml.tmpl
+++ b/deploy/kubernetes/keytransparency-deployment.yml.tmpl
@@ -49,6 +49,7 @@ metadata:
     run: kt-server
 spec:
   type: LoadBalancer
+  loadBalancerIP: 35.184.134.53
   ports:
     - port: 8080
       targetPort: 8080
@@ -114,6 +115,7 @@ metadata:
     run: prometheus
 spec:
   type: LoadBalancer
+  loadBalancerIP: 35.184.145.242
   ports:
     - port: 9090
       targetPort: 9090
--- a/deploy/kubernetes/keytransparency-deployment.yml.tmpl
+++ b/deploy/kubernetes/keytransparency-deployment.yml.tmpl
@@ -44,9 +44,9 @@ spec:
 apiVersion: v1
 kind: Service
 metadata:
-  name: kt-server
+  name: kt-server-external
   labels:
-    run: kt-server
+    run: kt-server-external
 spec:
   type: LoadBalancer
   loadBalancerIP: 35.184.134.53
@@ -54,6 +54,18 @@ spec:
     - port: 8080
       targetPort: 8080
       name: json-grpc
+  selector:
+    run: kt-server
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: kt-server
+  labels:
+    run: kt-server
+spec:
+  type: NodePort
+  ports:
     - port: 8081
       targetPort: 8081
       name: metrics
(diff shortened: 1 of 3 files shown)


gdbelvin:
Please also update the default .keytransparency.yml file

Liamsi:
> Please update the .keytransparency.yml file to use the new test server.

That is a good idea. Currently, the external endpoint's IP is only known after the deployment (after merging this pull or by manually configuring an external endpoint). The last IP I've configured (and the manually edited kubernetes configuration) was overwritten with that last merge to master. I'll see if we can get a static IP, though. If yes, I'll set the IP in the script and in the yml file.

Liamsi:
Static IPs reserved and updated. PTAL

> I'm not sure about the test keys. We might have to make them available some other way.

Maybe the client should be able retrieve them via grpc (similar as in trillian). Also, the client coudl use the same mechanism for retrieving the `map-id` which it needs as well.

deploy/kubernetes/keytransparency-deployment.yml.tmpl
```diff
@@ -46,11 +48,15 @@ metadata:
...
     run: kt-server
 spec:
-  type: NodePort
+  type: LoadBalancer
+  loadBalancerIP: 35.184.134.53
   ports:
     - port: 8080
       targetPort: 8080
       name: json-grpc
+    - port: 8081
```
gdbelvin: don't expose this to the world :-)

scripts/prepare_client.sh
```diff
@@ -70,7 +70,7 @@ cd "${GOPATH}/src/github.com/google/keytransparency"
 go build ./cmd/keytransparency-client
 
 # Generate .keytransparency.yaml file.
-KTYAML="log-key: \"../trillian/testdata/ct-http-server.pubkey.pem\"
+KTYAML="log-key: \"../trillian/testdata/log-rpc-server.pubkey.pem\"
```
gdbelvin: Isn't this supposed to be `genfiles/trillian-log.pem`?

Please add TODO to fetch this info from #672

Liamsi:
>> I'm not sure about the test keys. We might have to make them available some other way.

> Maybe the client should be able retrieve them via grpc (similar as in trillian). Also, the client could use the same mechanism for retrieving the map-id which it needs as well.

See #672

deploy/kubernetes/keytransparency-deployment.yml.tmpl
Liamsi: Done: Now there is an external service for the kt-server (that exposes only the web interface to the world) and to "internal" (type: NodePort) services, such that prometheus can find the dns `kt-server` and `kt-signer`. NodePort ports do not autm. create a firewall rule that expose the port (while LoadBalancer type services do).

scripts/prepare_client.sh
Liamsi: Done

PR #683, opened 2017-07-05
Enhance Travis Presubmit Process

### LINKED ISSUE #682: Make travis fail if make presubmit does not look good
I noticed that `make presubmit` is sometimes not ran before PRs are submitted. We should follow Trillian and check that it doesn't find any issues as part of travis run.

cesarghali:
This PR forces travis to fail if any of the following has issues:

- `gofmt`.
- `golint`.
- `go vet`.
- `errcheck`.
- `gocyclo`.
- `ineffassign`.
- `misspell`.
- Missing license somewhere.

Closes #682.

### DIFF
--- a/scripts/travis-presubmit.sh
+++ b/scripts/travis-presubmit.sh
@@ -1,0 +1,102 @@
+#!/bin/bash
+#
+# Checks for lint errors, spelling, licensing and so on.
+set -eu
+
+check_deps() {
+  local failed=0
+  check_cmd golint github.com/golang/lint/golint || failed=10
+  check_cmd errcheck github.com/kisielk/errcheck || failed=11
+  check_cmd gocyclo github.com/fzipp/gocyclo || failed=12
+  check_cmd ineffassign github.com/gordonklaus/ineffassign || failed=13
+  check_cmd misspell github.com/client9/misspell/cmd/misspell || failed=14
+  return $failed
+}
+
+check_cmd() {
+  local cmd="$1"
+  local repo="$2"
+  if ! type -p "${cmd}" > /dev/null; then
+    echo "${cmd} not found, try to 'go get -u ${repo}'"
+    return 1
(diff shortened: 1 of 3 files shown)


scripts/travis-presubmit.sh
```diff
@@ -0,0 +1,102 @@
...
+    git status
+    git diff
+    fail=1
+  fi
+
+  printf "running golint...\n"
+  status=$(find . -iregex '.[^.]*.go' ! -name 'gen.go' ! -path "./vendor/*" -exec golint {} \;)
+  if [[ -n ${status} ]]; then
+    printf "golint found the following issues:\n${status}\n"
+    fail=1
```
Liamsi: Not sure how I feel failing on `golint`. I would expect it to complain quite often. And sometimes ignoring it is the right decision:

"The suggestions made by golint are exactly that: suggestions. Golint is not perfect, and has both false positives and false negatives. Do not treat its output as a gold standard. We will not be adding pragmas or other knobs to suppress specific warnings, so do not expect or require code to be completely "lint-free". In short, this tool is not, and will never be, trustworthy enough for its suggestions to be enforced automatically, for example as part of a build process. " - https://github.com/golang/lint#purpose

What could be a good idea is to enable some specific rules which never have false positives and are always useful. If you like this idea, I can lookup a list of rules.

cesarghali: Fair enough. I just stopped failing on `golint` which is the behavior we had before.

I like the idea of enabling certain rules but I think it deserves its own PR.

Liamsi: I've just learned that `golint` doesn't return a non-zero status if it found any issues. This means neither the code above nor the changed line in c3d66f8 would fail or produce any output. If you want at least to see `golint`'s output you can add a `set_exit_status` flag to its call.

PR #686, opened 2017-07-06
Expose GetMutations grpc

### LINKED ISSUE #685: Expose Querying GetMutations HTTP API
Querying `/v1/epochs/1` returns "Not Found" although epoch 0 and 1 have passed (and were signed)

Seems like the existing `MutationService` is not registered yet. 
- [ ] register service

Liamsi:
resolves #685


cmd/keytransparency-server/main.go
```diff
@@ -206,7 +210,11 @@ func main() {
 		grpc.StreamInterceptor(grpc_prometheus.StreamServerInterceptor),
 		grpc.UnaryInterceptor(grpc_prometheus.UnaryServerInterceptor),
 	)
+
+	// TODO(ismail): replace with a non-core version (cmutations)?
```
cesarghali: This is fine. We don't need to replace this with non-core version since this file is not imported.

core/mutation/mutation.go
```diff
@@ -28,6 +28,7 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 
+	spb "github.com/google/keytransparency/impl/proto/mutation_v1_service"
```
cesarghali: I don't think this will work. `core/mutation` cannot depend on `impl/proto/mutation_v1_service`.
If I remember correctly, the entire implementation of the streaming API should be done in `impl/proto/mutation_v1_service`. We might need to refactor `core/mutation/mutation.go` a little bit to reuse some of the core functionality.

core/mutation/mutation.go
```diff
@@ -131,6 +133,10 @@ func (s *Server) GetMutations(ctx context.Context, in *tpb.GetMutationsRequest)
 	}, nil
 }
 
+func (s *Server) GetMutationsStream(in *tpb.GetMutationsRequest, stream spb.MutationService_GetMutationsStreamServer) error {
```
cesarghali: I think we can remove this API from here. `MutationService_GetMutationsStreamServer` is in `impl/proto/mutation_v1_service` and cannot be a dependency here.

cmd/keytransparency-server/main.go
Liamsi: OK, thanks. I wasn't entirely sure. TODO will be removed with the next commit.

core/mutation/mutation.go
Liamsi: Sure, as discussed on hagouts, I was already aware of this. I've created this PR early because I wasn't sure why not even the non-stream version `GetMutations` is exposed by these changes. It should already, or not?

core/mutation/mutation.go
Liamsi: Same as above. Will remove with the following commit.

core/mutation/mutation.go
Liamsi: OK, `RegisterMutationServiceHandlerFromEndpoint` was missing. That's why it didn't work.

Liamsi:
This is ready for another review. PTAL

cmd/keytransparency-server/main.go
```diff
@@ -28,10 +28,14 @@ import (
 	"github.com/google/keytransparency/core/crypto/vrf/p256"
 	"github.com/google/keytransparency/core/keyserver"
 	"github.com/google/keytransparency/core/mapserver"
+	cmutation "github.com/google/keytransparency/core/mutation"
 	"github.com/google/keytransparency/core/mutator/entry"
 	ctxn "github.com/google/keytransparency/core/transaction"
+
 	gauth "github.com/google/keytransparency/impl/google/authentication"
+	"github.com/google/keytransparency/impl/mutation"
 	pb "github.com/google/keytransparency/impl/proto/keytransparency_v1_service"
```
gdbelvin: for consistency, please rename `pb` to `ktpb` (key transparency protobuf) no that it is no longer the only protobuf being included.

PR #698, opened 2017-07-10
Implement the Authorization Module

cesarghali:
This PR implements the authorization module for Key Transparency. This module allows users to access their own profiles and admins to access user profiles if they're authorized to do so.

Testing:
- go test impl/authorization/...
- make presubmit


impl/authorization/authorization_test.go
```diff
@@ -0,0 +1,257 @@
+package authorization
```
Liamsi: Lacking License header

impl/authorization/authorization.go
```diff
@@ -0,0 +1,75 @@
...
+			if p == sctx.Identity() && permitted(role, permission) {
+				return nil
+			}
+		}
+	}
+	return fmt.Errorf("%v is not authorized to perform %v on resource defined by <mapID=%v, appID=%v>", sctx.Identity(), permission, mapID, appID)
+}
+
+func resourceLabel(mapID, appID int64) string {
+	return fmt.Sprintf("%d|%d", mapID, appID)
```
gdbelvin: This works because mapID and appID don't contain the `|` character

impl/authorization/authorization.go
```diff
@@ -0,0 +1,75 @@
...
+	// Case 1.
+	if sctx.Identity() == userID {
+		return nil
+	}
+
+	// Case 2.
+	rLabel := resourceLabel(mapID, appID)
+	roles, ok := a.policy.GetResourceToRoles()[rLabel]
+	if !ok {
+		return fmt.Errorf("resource defined by <mapID=%v, appID=%v> does not exist", mapID, appID)
```
gdbelvin: does not have a defined policy?

Should this be access denied? If a resource doesn't have a policy it should failed closed.

core/proto/authorization/authorization.proto
```diff
@@ -43,7 +43,7 @@ message AuthorizationPolicy {
 
   // Roles contains a lot of role labels identifying each role.
   message Roles {
```
gdbelvin: consider renaming this RoleLabels

core/proto/authorization/authorization.proto
```diff
@@ -43,7 +43,7 @@ message AuthorizationPolicy {
...
-    repeated string label = 1;
+    repeated string labels = 1;
   }
 
   // label_to_role is a map of roles keyed by labels used in Roles.
```
gdbelvin: consider naming resource_to_roles - > resource_to_role_labels

gdbelvin: consider naming label_to_role, simply "roles"

impl/authorization/authorization.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	for _, l := range roles.GetLabels() {
+		role := a.policy.GetLabelToRole()[l]
+		for _, p := range role.GetPrincipals() {
```
gdbelvin: consider breaking this inner loop into a function `isPrincipalInRole`

impl/authorization/authorization.go
```diff
@@ -0,0 +1,75 @@
...
+		}
+	}
+	return fmt.Errorf("%v is not authorized to perform %v on resource defined by <mapID=%v, appID=%v>", sctx.Identity(), permission, mapID, appID)
+}
+
+func resourceLabel(mapID, appID int64) string {
+	return fmt.Sprintf("%d|%d", mapID, appID)
+}
+
+func permitted(role *authzpb.AuthorizationPolicy_Role, permission authzpb.Permission) bool {
```
gdbelvin: isPermissionInRole?

impl/authorization/authorization.go
cesarghali: They both `int64`. :)

impl/authorization/authorization_test.go
```diff
@@ -0,0 +1,257 @@
...
+	admin3   = "admin3@example.com"
+	admin4   = "admin4@example.com"
+	res1     = "1|1"
+	res2     = "1|2"
+	res3     = "1|3"
+	res4     = "1|4"
+)
+
+func setup() *authz {
+	roles1 := &authzpb.AuthorizationPolicy_Roles{
```
gdbelvin: This might be clearer as one big proto

impl/authorization/authorization.go
cesarghali: The error message is updated.

The gRPC handler will return an access denied error code when the call to this function fails.

impl/authorization/authorization_test.go
cesarghali: Good call. I didn't try to imagine it in one big proto, but it looks nicer now :)

cesarghali:
PR is updated. PTAL.

cesarghali:
@gdbelvin as for the default policy, something should be added in the next PRs.

impl/authorization/authorization.go
```diff
@@ -65,7 +63,16 @@ func resourceLabel(mapID, appID int64) string {
 	return fmt.Sprintf("%d|%d", mapID, appID)
 }
 
-func permitted(role *authzpb.AuthorizationPolicy_Role, permission authzpb.Permission) bool {
+func isPrincipalInRole(role *authzpb.AuthorizationPolicy_Role, identity string, permission authzpb.Permission) bool {
+	for _, p := range role.GetPrincipals() {
+		if p == identity && isPermisionInRole(role, permission) {
```
gdbelvin: Consider removing isPermissionInRole from this function, and having the caller test both 
`isPrincipalInRole() && isPermissionInRole()`

PR #700, opened 2017-07-11
Make sure URLs work with default parameters

### LINKED ISSUE #699: URLs exposed by services should work by default
Using GET: Even if mutations exist in between epochs (in the DB) `GetMutations` doesn't return them. The main problem is that the default `page_size` is zero. ~~There might be an additional one-off problem.~~

Liamsi:
Make sure the URLs
https://kt-server/v1/users/user@example.com/history
and 
https://kt-server/v1/epochs/1
work by default:
- add a default `page_size` to `GetMutations` API to return mutations by default
- add a default `start` to `ListEntryHistory` API to return entries by default (epochs start at 1)
- also make sure to fail earlier if `GetMutations` is explicitly called with invalid parameters (e.g. epoch<0)
- fix tests according to default `start` epoch equal to 1 (if not set) 
- for the client `ListHistory` epochs <1 (or =0) are forbidden (and fail early, too)
- for HTTP GET requests epoch = 0 is "allowed": we assume it's not set and set it to 1

resolves #699


core/keyserver/keyserver.go
```diff
@@ -39,6 +39,8 @@ const (
 	defaultPageSize = 16
 	// Maximum allowed requested page size to prevent DOS.
 	maxPageSize = 16
+  // If no epoch is provided default to epoch 1.
```
cesarghali: This doesn't look like the right indentation. Maybe it's only shown this way in github.

core/mutation/validate.go
```diff
@@ -0,0 +1,54 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package mutation
+
+import (
+	"errors"
+	tpb "github.com/google/keytransparency/core/proto/keytransparency_v1_types"
```
cesarghali: Please add an empty line between the two imports.

core/mutation/validate.go
```diff
@@ -0,0 +1,54 @@
...
+	// Maximum allowed requested page size to prevent DOS.
+	maxPageSize = 16
+)
+
+var (
+	// ErrInvalidStart occurs when the start epoch of ListEntryHistoryRequest
+	// is not valid (not in [1, currentEpoch]).
+	ErrInvalidStart = errors.New("invalid start epoch")
+	// ErrInvalidPagesize occurs when the page size is > 0.
+	ErrInvalidPagesize = errors.New("Invalid page size")
```
cesarghali: Please capitalize size, e.g., `ErrInvalidPageSize`. Same in the comment.

cesarghali:
Also, please check travis log. Seems like `make presubmit` needs to be ran locally.

Liamsi:
Adding @gdbelvin as a 2nd reviewer

core/keyserver/validate.go
```diff
@@ -44,12 +44,12 @@ var (
 	ErrNoCommitted = errors.New("missing commitment")
 	// ErrCommittedKeyLen occurs when the committed key is too small.
 	ErrCommittedKeyLen = errors.New("committed.key is too small")
-	// ErrWrongIndex occurs when the index in key value does not match the
-	// output of VRF.
-	ErrWrongIndex = errors.New("index does not match VRF")
```
cesarghali: Can you please keep this here? This will make the diff concise. Unless there's a reason for moving it.

Liamsi: Thanks! Makes sense. Done.

cmd/keytransparency-client/grpcc/grpc_client.go
```diff
@@ -139,7 +139,11 @@ func min(x, y int32) int32 {
 
 // ListHistory returns a list of profiles starting and ending at given epochs.
 // It also filters out all identical consecutive profiles.
+// Epochs start at 1.
 func (c *Client) ListHistory(ctx context.Context, userID, appID string, start, end int64, opts ...grpc.CallOption) (map[*trillian.SignedMapRoot][]byte, error) {
+	if start <= 0 {
```
gdbelvin: This should just be `start==0` because proto3 defines omitted values to have zero as their default

core/keyserver/validate.go
```diff
@@ -109,6 +109,10 @@ func validateListEntryHistoryRequest(in *tpb.ListEntryHistoryRequest, currentEpo
 	if in.Start < 0 || in.Start > currentEpoch {
 		return ErrInvalidStart
 	}
+	// Ensure a valid start epoch is provided if the Start parameter is not set.
+	if in.Start == 0 {
+		in.Start = defaultStartEpoch
```
gdbelvin: Also add a TODO and an issue, to zero index our epochs.

cmd/keytransparency-client/grpcc/grpc_client.go
Liamsi: This is the client with parameters the caller provides (could be any `int64`) and not by a recieved proto. Hence, I think it makes sense to keep it like this. We could change the params to unsigned `uint64` instead (than the check `==0` would be right).

core/mutation/validate.go
Liamsi: done

core/keyserver/validate.go
Liamsi: https://github.com/google/keytransparency/issues/703

cmd/keytransparency-client/grpcc/grpc_client.go
Liamsi: I added a test-case to document this isn't meant to be called with a start <0: https://github.com/google/keytransparency/pull/700/commits/c503ec6a33828321d1e566492cf9136b01fc3260#diff-f8b79d8d11b0c9af5cfa9f8d823cbb91L225

PR #702, opened 2017-07-13
Plug in the Authorization Module in keyserver.go

cesarghali:
This PR plugs in the authorization module implemented in #698 in `keyserver.go`. This requires changing the authentication module to produce an `*authentication.SecurityContext` object that is passed to the `IsAuthorized` API.


cmd/keytransparency-server/main.go
```diff
@@ -45,7 +46,7 @@ import (
 	"github.com/google/trillian"
 
 	"github.com/golang/glog"
-	"github.com/grpc-ecosystem/go-grpc-prometheus"
+	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
```
gdbelvin: it would be nice to not rename this import if possible

cmd/keytransparency-server/main.go
```diff
@@ -169,6 +170,7 @@ func main() {
 	default:
 		glog.Exitf("Invalid auth-type parameter: %v.", *authType)
 	}
+	authz := authorization.New()
```
gdbelvin: `New()` Uses an empty policy proto right?  aka, default authorization

core/keyserver/keyserver.go
```diff
@@ -229,17 +234,22 @@ func (s *Server) ListEntryHistory(ctx context.Context, in *tpb.ListEntryHistoryR
...
-	case authentication.ErrWrongUser:
-		return nil, grpc.Errorf(codes.PermissionDenied, "Permission denied")
 	case authentication.ErrMissingAuth:
 		return nil, grpc.Errorf(codes.Unauthenticated, "Missing authentication header")
 	default:
 		glog.Warningf("Auth failed: %v", err)
 		return nil, grpc.Errorf(codes.Unauthenticated, "Unauthenticated")
 	}
+	// Validate proper authorization.
+	// TODO: use the correct appID.
```
gdbelvin: AppID is present in the user request `in.AppId`

cmd/keytransparency-server/main.go
cesarghali: Yes, that's correct.

core/keyserver/keyserver.go
cesarghali: Oh, I missed it somehow, thanks for the tip.

PR #705, opened 2017-07-13
Regenerate protos

gdbelvin:
This PR, along with google/trillian#584 gets us back on track for being able to generate protos correctly. 

With regard to the grpc-gateway `runtime.AnnotateContext` issue:
- etcd, which trillian uses, vendors grpc-gateway to it's latest released version v1.2.2
- Trillian generate `pb.gw.go` file accordingly
- Which means KT must do the same. 
- It's a good idea to use release versions rather than HEAD.


vendor/vendor.json
```diff
@@ -1,6 +1,105 @@
 {
 	"comment": "",
 	"ignore": "test",
-	"package": [],
+	"package": [
```
cesarghali: My understanding is that we want to ignore changes to this file, right? These are `govendor` specific changes.

core/proto/keytransparency_v1_types/gen.go
```diff
@@ -12,6 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:generate protoc -I=. -I=$GOPATH/src/ -I=$GOPATH/src/github.com/google/trillian/ --go_out=:. keytransparency_v1_types.proto
+//go:generate protoc -I=. -I=$GOPATH/src/ -I=$GOPATH/src/github.com/google/trillian/ -I=$GOPATH/src/github.com/googleapis/googleapis --go_out=:. keytransparency_v1_types.proto
```
cesarghali: Should we do like Trillian and use the `googleapis` package in vendored `grpc-gateway`?
https://github.com/google/trillian/blob/master/gen.go#L18

vendor/vendor.json
gdbelvin: This file is useful for documenting what we did, but you're right, `go get` ignores this file and just looks at the contents of the vendor directory.

core/proto/keytransparency_v1_types/gen.go
gdbelvin: hmm, Good question. I didn't notice that.. 
Hmm, I'm thinking no, because grpc-gateway isn't actually vendoring it (ie. it's not in the vendor directory), so the files that the compiler will actually be using are the native ones. 
What do you think?

vendor/vendor.json
cesarghali: Good idea!

core/proto/keytransparency_v1_types/gen.go
cesarghali: Fair enough. Let's keep it the way it is then.

Issue #715, opened 2017-07-31
Remove protobuf references from verification interfaces. 


gdbelvin:
`gomobile bind` doesn't support passing protobuf objects as function arguments. 
This set of PRs will convert the verification API interface to use basic types. 

Supports #711

gdbelvin:
Duplicate of #713

Issue #713, opened 2017-07-28
Remove dependency on google protobuf from verification functions


gdbelvin:
go mobile bind does not like the protobuf generated code. 
To get our functions to run on mobile platforms, let's create a verification interface that depends only on native types. 

Supports #711

PR #716, opened 2017-07-31
Remove proto from commitments package

### LINKED ISSUE #717: Verifying a nil commitment silently passes
https://github.com/google/keytransparency/blob/master/core/client/kt/verify.go#L70:
```
// VerifyCommitment verifies that the commitment in `in` is correct for userID.
func (Verifier) VerifyCommitment(userID, appID string, in *tpb.GetEntryResponse) error {
	if in.Committed != nil {
		entry := new(tpb.Entry)
		if err := proto.Unmarshal(in.GetLeafProof().GetLeaf().GetLeafValue(), entry); err != nil {
			return err
		}
		if err := commitments.Verify(userID, appID, entry.Commitment, in.Committed); err != nil {
			return err
		}
	}
	return nil
}
```
If `in.committed == nil` the verification silently passes (no error is thrown). Is this intended? If so, why?
I would expect an error to be raised when Committed == nil

gdbelvin:
In order to make mobile bindings possible, we need to remove protobuf
references from verification interfaces.

Part of #715  #713
Closes #717


core/crypto/commitments/commitments.go
```diff
@@ -53,26 +51,25 @@ var (
...
-	Write(ctx context.Context, commitment []byte, committed *tpb.Committed) error
+	Write(ctx context.Context, commitment, data, nonce []byte) error
 	// Read looks up a cryptograpic commitment and returns associated data.
-	Read(ctx context.Context, commitment []byte) (*tpb.Committed, error)
+	Read(ctx context.Context, commitment []byte) (data, nonce []byte, err error)
 }
 
 // Commit makes a cryptographic commitment under a specific userID to data.
-func Commit(userID, appID string, data []byte) ([]byte, *tpb.Committed, error) {
+func Commit(userID, appID string, data []byte) (commitment, nonce []byte, err error) {
```
AMarcedone: Despite being better than before (as it does not depend on protobuf), this function is still not gobind friendly, as it returns 3 outputs. To be "visible" from Java, functions need to return only one output (plus an optional second one which needs to be of type error).

gdbelvin:
Would it be easier to return a struct, or should those be avoided as well?

AMarcedone:
A struct would work (as long as its members are of gobind friendly types such as []byte).
I am not sure what is best from a design perspective, but my original idea was to leave the interface as it currently is and add some simple gobind friendly wrappers that internally call the original proto based interfaces. I think it is worth discussing this face to face tomorrow.

AMarcedone:
Here is a concrete example of the interface I was thinking about (the B prefix is meant to indicate the goBind friendly version of a function)
https://github.com/AMarcedone/keytransparency/commit/6579ef03e5d478f158360ccf58c8d857f6b265a0

core/crypto/commitments/commitments.go
gdbelvin: To fix this, I've created a `GenCommitmentKey` function.

core/keyserver/validate_test.go
```diff
@@ -81,19 +81,23 @@ func TestValidateUpdateEntryRequest(t *testing.T) {
 	appID := "app"
 	vrfPriv, _ := p256.GenerateKey()
 	index, _ := vrfPriv.Evaluate(vrf.UniqueID(userID, appID))
-	commitment, committed, _ := commitments.Commit(userID, appID, profileData)
+	nonce, err := commitments.GenCommitmentKey()
```
cesarghali: Consider using a fixed nonce to make the test repeatable.

impl/sql/commitments/sql_test.go
```diff
@@ -39,10 +39,11 @@ func TestWriteRead(t *testing.T) {
 
 	// Create test data.
 	pdata := []byte("key")
-	commitmentC, committedC, err := commitments.Commit("foo", "app", pdata)
+	nonce, err := commitments.GenCommitmentKey()
```
cesarghali: Same here.

core/client/kt/verify.go
```diff
@@ -88,9 +74,22 @@ func (Verifier) VerifyCommitment(userID, appID string, in *tpb.GetEntryResponse)
...
+
+	// If this is not a proof of absence, verify the connection between
+	// profileData and the commitment in the merkle tree leaf.
+	if in.GetCommitted() != nil {
+		commitment := entry.GetCommitment()
+		data := in.GetCommitted().GetData()
+		nonce := in.GetCommitted().GetKey()
+		if err := commitments.Verify(userID, appID, commitment, data, nonce); err != nil {
+			Vlog.Printf("✗ Commitment verification failed.")
+			return fmt.Errorf("VerifyCommitment(): %v", err)
```

core/client/kt/verify_test.go
```diff
@@ -1,73 +0,0 @@
-// Copyright 2016 Google Inc. All Rights Reserved.
```
gdbelvin: Good idea.

core/crypto/commitments/commitments.go
```diff
@@ -50,36 +46,21 @@ var (
...
-type Committer interface {
-	// Write saves a cryptographic commitment and associated data.
-	Write(ctx context.Context, commitment []byte, committed *tpb.Committed) error
-	// Read looks up a cryptograpic commitment and returns associated data.
-	Read(ctx context.Context, commitment []byte) (*tpb.Committed, error)
-}
-
-// Commit makes a cryptographic commitment under a specific userID to data.
-func Commit(userID, appID string, data []byte) ([]byte, *tpb.Committed, error) {
+// GenCommitmentKey generates a commitment key for use in Commit.  This key
```
gdbelvin: done

core/keyserver/validate_test.go
gdbelvin: done

impl/sql/commitments/sql_test.go
gdbelvin: done

gdbelvin:
Added tests and cleaned up. PTAL.

core/client/kt/verify_test.go
```diff
@@ -17,56 +17,106 @@ package kt
...
-			LeafProof: &trillian.MapLeafInclusion{
-				Leaf: &trillian.MapLeaf{
-					LeafValue: tc.entryData,
+		{
+			desc:    "empty",
+			userID:  "noalice",
+			appID:   "app",
+			trusted: &trillian.SignedLogRoot{},
+			in: &keytransparency_v1_types.GetEntryResponse{
+				VrfProof:  []byte{0x8, 0xca, 0x1, 0xaa, 0x1, 0x74, 0x70, 0xf4, 0xbf, 0x96, 0x69, 0xff, 0x73, 0x5d, 0xc7, 0x9d, 0x2b, 0x51, 0xb0, 0xa5, 0x42, 0xb8, 0x8c, 0x42, 0x34, 0xfc, 0xa1, 0xa2, 0xc0, 0x80, 0xdf, 0x76, 0x33, 0xfc, 0x7d, 0xee, 0x4a, 0xf3, 0x18, 0xea, 0x30, 0xc4, 0xa4, 0x6f, 0x31, 0xc9, 0x20, 0x73, 0x42, 0x84, 0xd9, 0x71, 0x39, 0x52, 0xb2, 0x8f, 0x58, 0x52, 0x4, 0x53, 0x87, 0x8, 0x3e, 0x81, 0x4, 0xb, 0x13, 0x89, 0xd7, 0xc6, 0x63, 0x22, 0x39, 0x18, 0x73, 0x72, 0xfa, 0x32, 0xf6, 0xeb, 0x3, 0x8, 0x5d, 0x7, 0x4e, 0x2, 0x3a, 0xc6, 0x7f, 0x89, 0xe8, 0x44, 0x27, 0xcb, 0x73, 0xdc, 0xf2, 0x2f, 0xcc, 0xcd, 0x90, 0x6e, 0x97, 0xcb, 0x22, 0xff, 0x6e, 0xdb, 0x74, 0x22, 0xbf, 0x28, 0x27, 0x9b, 0x9e, 0x26, 0x1a, 0xe4, 0xc6, 0x16, 0x59, 0x4f, 0x7d, 0xcc, 0xb9, 0x8e, 0x7d, 0x41, 0xf7},
```
cesarghali: Up to you: What do you think about expressing this as a hexa string and then convert to byte before using it? It'll be shorter. Same applies below.

gdbelvin: That's true, but makes the test vectors harder to regenerate with `log.Printf("%#v")`

cesarghali: :+1:

PR #721, opened 2017-08-01
Use MapID as reported in SignedMapRoot

gdbelvin:
MapID is part of SignedMapRoot and part of leaf and empty hash
computation. We can remove a bit of state from the client without a
reduction in security by relying on the value as reported by the server.

### DIFF
--- a/cmd/keytransparency-client/grpcc/grpc_client.go
+++ b/cmd/keytransparency-client/grpcc/grpc_client.go
@@ -91,7 +91,7 @@ type Client struct {
 }
 
 // New creates a new client.
-func New(mapID int64,
+func New(
 	client spb.KeyTransparencyServiceClient,
 	vrf vrf.PublicKey,
 	verifier crypto.PublicKey,
@@ -99,7 +99,7 @@ func New(mapID int64,
 	return &Client{
 		cli:        client,
 		vrf:        vrf,
-		kt:         kt.New(vrf, tv.New(mapID, sparse.CONIKSHasher), verifier, log),
+		kt:         kt.New(vrf, tv.New(sparse.CONIKSHasher), verifier, log),
 		log:        log,
 		mutator:    entry.New(),
 		RetryCount: 1,
--- a/core/client/kt/verify.go
+++ b/core/client/kt/verify.go
@@ -25,15 +25,16 @@ import (
 	"github.com/google/keytransparency/core/crypto/commitments"
 	"github.com/google/keytransparency/core/crypto/vrf"
 	"github.com/google/keytransparency/core/tree/sparse"
-	tv "github.com/google/keytransparency/core/tree/sparse/verifier"
 
 	"github.com/golang/protobuf/proto"
+	"github.com/google/trillian"
 	"github.com/google/trillian/client"
-	tcrypto "github.com/google/trillian/crypto"
 	"golang.org/x/net/context"
 
+	tv "github.com/google/keytransparency/core/tree/sparse/verifier"
+	tcrypto "github.com/google/trillian/crypto"
+
 	tpb "github.com/google/keytransparency/core/proto/keytransparency_v1_types"
-	"github.com/google/trillian"
 )
 
 var (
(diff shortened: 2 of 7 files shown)


gdbelvin:
That's right.  An append-only log makes sure that if the server ever forks, it must continue to fork in perpetuity.  This makes it much easier for a gossip protocol to detect the inconsistency.

cmd/keytransparency-client/grpcc/grpc_client.go
```diff
@@ -91,15 +91,15 @@ type Client struct {
 }
 
 // New creates a new client.
-func New(mapID int64,
+func New(
 	client spb.KeyTransparencyServiceClient,
```
cesarghali: Please move this to the line above it.

core/client/kt/verify.go
```diff
@@ -25,15 +25,16 @@ import (
 	"github.com/google/keytransparency/core/crypto/commitments"
 	"github.com/google/keytransparency/core/crypto/vrf"
 	"github.com/google/keytransparency/core/tree/sparse"
-	tv "github.com/google/keytransparency/core/tree/sparse/verifier"
```
cesarghali: Please keep this here, AFAIK we're putting all Key Transparency imports in the same group regardless of they're renamed.

core/client/kt/verify.go
```diff
@@ -25,15 +25,16 @@ import (
...
 
 	"github.com/golang/protobuf/proto"
+	"github.com/google/trillian"
 	"github.com/google/trillian/client"
-	tcrypto "github.com/google/trillian/crypto"
```
cesarghali: Same here.

core/client/kt/verify.go
cesarghali: The grouping that we're using is:

- Standard packages.
- KT packages.
- Others.
- Proto.

I'm completely fine with a new scheme but we should apply it to all files.

core/client/kt/verify.go
```diff
@@ -106,7 +107,8 @@ func (v *Verifier) VerifyGetEntryResponse(ctx context.Context, userID, appID str
 		return ErrNilProof
 	}
 
-	if err := v.tree.VerifyProof(leafProof.Inclusion, index[:], leafProof.Leaf.LeafValue, sparse.FromBytes(in.GetSmr().RootHash)); err != nil {
+	mapID := in.GetSmr().GetMapId()
```
cesarghali: No need to define `mapID` here. Its value can be inlined in the next line.

core/tree/sparse/verifier/verifier.go
```diff
@@ -38,27 +38,25 @@ var (
...
 	return &Verifier{
-		mapID:  mapID,
 		hasher: hasher,
 	}
 }
 
 // VerifyProof verifies a tree proof of a given leaf at a given index based on
 // the provided root and neighbor list
-func (v *Verifier) VerifyProof(neighbors [][]byte, index, leaf []byte, root sparse.Hash) error {
+func (v *Verifier) VerifyProof(treeID int64, neighbors [][]byte, index, leaf []byte, root sparse.Hash) error {
```
cesarghali: Please use `mapID` instead of `treeID` to keep the terms consistent.

core/tree/sparse/verifier/verifier.go
```diff
@@ -73,7 +71,7 @@ func (v *Verifier) VerifyProof(neighbors [][]byte, index, leaf []byte, root spar
 
 // calculateRoot calculates the root of the tree branch defined by leaf and
 // neighbors.
-func (v *Verifier) calculateRoot(neighbors [][]byte, bindex string, leaf []byte) (sparse.Hash, error) {
+func (v *Verifier) calculateRoot(treeID int64, neighbors [][]byte, bindex string, leaf []byte) (sparse.Hash, error) {
```
cesarghali: Same here.

cesarghali:
I just realized this got merged. I can apply those changes in a "small fixes" PR.

PR #722, opened 2017-08-01
Add API to publish verification data

### LINKED ISSUE #672: Server should publish verification data
There should be an API for requesting the data needed to verify a particular KT server. 
This data should never change for a particular stable instance of KT and clients should have this information, or a hash of this information baked in, but it is a useful convenience feature and supports discovery of new KT servers.

Liamsi:
- add admin client to key-server
- add `Public()` method to VRF private key
- add `GetDomainInfo` to proto and key-server

resolves #672


Liamsi:
I'm fixing the integration test and will merge if travis passes. 

cc @AMarcedone that is the PR Gary was referring to during the hangout. You should be able to read the key (der encoded) by querying the API with the corresponding URL (see #672)

AMarcedone:
Awesome @Liamsi! 
Looking forward to have this merged so I can start testing grpc from Android. In the meantime I will try to deploy my own version.

Issue #711, opened 2017-07-20
Java Client


gdbelvin:
Resources
- https://godoc.org/golang.org/x/mobile/cmd/gobind

AMarcedone:
I am starting to look into using gobind. Maybe you tagged the wrong Antonio?

gdbelvin:
Yep -- my bad. Thanks for taking this on!

AMarcedone:
Update: I finally got an android studio project which leverages gobind to work (had to guess the right combination of gradle plugin versions that work together, as this is confusing and poorly documented).

However, the result is not really useful due to gobind's type restrictions: https://godoc.org/golang.org/x/mobile/cmd/gobind#hdr-Type_restrictions

I will briefly look into other similar alternatives and probably start a new Java implementation from scratch.

gdbelvin:
It sounds like things might work if we created a verification API surface that was sufficiently simple? 
Could you start with the [Commitments](https://github.com/google/keytransparency/tree/master/core/crypto/commitments) package?

AMarcedone:
I will try looking into that. My biggest concerns are how to handle logging (not sure how compatible the systems are) and the filesystem interface (I imagine I will have to use a Java interface for the android filesystem to read public key and cert files and then pass them as strings or byte arrays to the go part of the code).
Will keep you posted.

AMarcedone:
Update: I wrote an interface for Commitments which correctly runs on my Android phone.

Next, I will look into extending the interface to other packages, as well as how to handle logging from go to android.

One issue I run into is that I cannot return protobuf generated structs to Java or use them directly as arguments in functional calls because gobind does not like the protobuf go code. I can serialize the protobufs to byte arrays to cross language boundaries. Not ideal, but it works.

gdbelvin:
amazing!
I can put together some PRs to remove the dependency on protobuf in the
verification functions.
Logging -- I wonder if you could hook into the go logging interface and
pipe it out to the right Java logger?

rmhrisk:
AWESOME

AMarcedone:
Hey @gdbelvin, thanks a lot for the PRs. I will review them later today. 
In general, from now on I think it will be easier if I modify the go interfaces as I need to hook them up with the Java code I am writing. I will put you as a reviewer, if you do not mind :)
Stay tuned :)

gdbelvin:
Sounds good.  :-)

AMarcedone:
Update: Progress is good. 
[...]

PR #725, opened 2017-08-03
Subject Alternative Names in cert generation

### LINKED ISSUE #724: Deployed kt server's ssl cert has the wrong ip
The ssl server deployed from travis has a (self signed) cert for 0.0.0.0 and not its actual ip (currently 35.184.134.53), therefore the connection from the client fails (using a browser allows to ignore the issue).

In addition, it would be great to expose the ssl cert as a file so it is easy to download and pass to the client as an argument (but I am now looking for some openssl commands to automatically download it)

Liamsi:
Quickfix: These changes are necessary to enable clients (e.g. in #711) to communicate with the deployed server.
closes #724


gdbelvin:
What's the status of this PR?

Liamsi:
This PR includes a few lines of changes such that @AMarcedone's grpc (Java) client can communicate with the deployed server without warning about a wrong IP. It should create a cert that can be used from inside docker-composer, kubernetes, and externally by using the server's IP. 

I think the scripts need further cleanup but it would be good if this got merged so that the deployed server can be contacted not only from inside the kubernetes cluster.

scripts/gen_server_keys.sh
```diff
@@ -18,21 +18,24 @@ COMMONNAME=""
...
     SUBJECT="${SUBJECT}/CN=${COMMONNAME}"
 fi
 
-SANEXT="[SAN]\nbasicConstraints=CA:TRUE\nsubjectAltName=DNS:localhost"
+# TODO(ismail): make the IPs configurable as well
+ALTNAMES="[alt_names]\nDNS.1=${SAN_DNS}\nDNS.2=localhost\nIP.1=0.0.0.0\nIP.2=35.184.134.53"
+SANEXT="[SAN]\nbasicConstraints=CA:TRUE\nsubjectAltName=@alt_names\n\n${ALTNAMES}"
 if [[ -n "${ADDRESS}" ]]; then
-    SANEXT="${SANEXT},IP.2:${ADDRESS}"
+    SANEXT="${SANEXT}"
```
AMarcedone: The address option here is ignored, which might be confusing for users. I suggest fixing it before merging.

scripts/gen_server_keys.sh
```diff
@@ -44,6 +47,8 @@ openssl genrsa -des3 -passout pass:x -out server.pass.key 2048
 openssl rsa -passin pass:x -in server.pass.key -out server.key
 chmod 600 server.key
 rm server.pass.key
+# The following command does not work on a Mac
```
AMarcedone: It works on mine. Both files exists, do not remember doing anything specific to fix it.

AMarcedone:
The deployed server works for me atm (so the cert is well formed and accepted).

Liamsi:
> The deployed server works for me atm (so the cert is well formed and accepted).

Thanks for testing that and thanks for the review. I removed the comment about the script not working on Macs, made the san extension mandatory, and made the deployed servers address configurable. PTAL

AMarcedone:
LGTM! Thanks for the changes and the PR in general.

PR #727, opened 2017-08-03
Use the Trillian testing environment 

### LINKED ISSUE #229: Delete Stale Code
This is a followup to issue #176.
- [x] Looks like no one is using `tree/sparse/memhist` package. Should we remove it?
- [x] From a previous marshaling/unmarshaling experience, it’s better to use `uint64` to represent time in ctmap.proto. If we use `uint64`, we can, then, remove `proto/protobuf/timestamp.proto`.
- [x] Check if `proto/ct` contains any used code. If not, delete it.
- [x] Memory tree in `tree/sparse/memtree`.
- [x] `tree/sparse/doc.go`.

gdbelvin:
This PR transitions Key Transparency from the KT mapserver implementation onto the Trillian MapEnv which runs a proper trillian map server which more closely replicates KT behavior while using a Trillian backend.

This PR also deletes the now unused code.  Closes #229


cmd/keytransparency-signer/main.go
```diff
@@ -108,20 +82,11 @@ func main() {
...
-			glog.Exitf("grpc.Dial(%v): %v", *mapURL, err)
-		}
-		tmap = trillian.NewTrillianMapClient(mconn)
-	} else {
-		var err error
-		tmap, err = newMapServer(context.Background(), sqldb, factory)
-		if err != nil {
-			glog.Exitf("newMapServer: %v", err)
-		}
+	mconn, err := grpc.Dial(*mapURL, grpc.WithInsecure())
```
Liamsi: This is not directly related to this pull but is the grpc-connection to the map-server always unencrypted?

cmd/keytransparency-server/main.go
```diff
@@ -192,22 +175,13 @@ func main() {
...
-		}
-		tmap = trillian.NewTrillianMapClient(mconn)
-		tadmin = trillian.NewTrillianAdminClient(mconn)
-	} else {
-		// Create an in-process readonly mapserver.
-		tmap, err = newReadonlyMapServer(context.Background(), *mapID, sqldb, factory)
-		if err != nil {
-			glog.Exitf("newReadonlyMapServer(): %v", err)
-		}
+
```
Liamsi: Remove this new extra line?

integration/client_test.go
```diff
@@ -263,7 +263,7 @@ func (e *Env) setupHistory(ctx context.Context, userID string, signers []signatu
 	// Note that profile 5 is submitted twice by the user to test that
 	// filtering case.
 	for i, p := range [][]byte{
-		nil, nil, cp(1), cp(2), nil, nil, cp(3), nil,
+		nil, cp(1), cp(2), nil, nil, cp(3), nil,
```
Liamsi: What is the reason for removing this `nil`? 

Also, if you keep it like this, update the comment above to:
```golang
// following profiles:
// [nil, 1, 2, 2, 2, 3, 3, 4, 5, 5, 5, 5, 5, 5, 6, 6, 5, 7, 7].
```

core/keyserver/keyserver_test.go
```diff
@@ -1,267 +0,0 @@
...
-		{0, 1, 2, []int{1}, codes.OK},                                                            // start epoch is not set (will default to 1).
-		{1, 10, 11, []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}, codes.OK},                              // 10 entries per page.
-		{4, 10, 14, []int{4, 5, 6, 7, 8, 9, 10, 11, 12, 13}, codes.OK},                           // start epoch is not 1.
-		{1, 0, 17, []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16}, codes.OK},       // zero page size.
-		{20, 10, 0, []int{20, 21, 22, 23, 24}, codes.OK},                                         // end of list.
-		{24, 10, 0, []int{24}, codes.OK},                                                         // requesting the very last entry.
-		{1, 1000000, 17, []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16}, codes.OK}, // DOS prevention.
-		{40, 10, 0, []int{}, codes.InvalidArgument},                                              // start epoch is beyond current epoch.
-		{-1, 1, 1, []int{0}, codes.InvalidArgument},                                              // start epoch is less than 0.
```
Liamsi: Shouldn't this test be moved to `integration` instead of deleting it? If yes, I'm happy if we merge this pull an open an issue for that (I can give it a shot).

Liamsi: Ah OK. A similar test from the client perspective is already here: https://github.com/google/keytransparency/pull/727/files/0ad94ff088be2e59b6832e5e8b258080bf6e0dc4#diff-f8b79d8d11b0c9af5cfa9f8d823cbb91R204

You can ignore this comment.

cmd/keytransparency-signer/main.go
gdbelvin: Yes, Trillian is supposed to be a backend connection, not exposed to the internet so at the moment it's unencrypted.

integration/client_test.go
gdbelvin: The trillian map server starts at revision 1 rather than revision 0. 
This change shifts the test data by one epoch to accomodate that.

core/keyserver/keyserver_test.go
gdbelvin: Yep.  This functionality is more rigorously tested in `integration/client_test.go`

integration/client_test.go
Liamsi: Thanks for the explanation! Could you update the comment to reflect this change as well?

gdbelvin: done

cmd/keytransparency-server/main.go
gdbelvin: done

Liamsi:
LGTM

PR #728, opened 2017-08-04
Use the CONIKS hasher

gdbelvin:
The moment of truth! - Where we finally switch back to using the CONIKS
hashing algorithm successfully :-)

related to google/trillian#780


gdbelvin:
This is currently failing because the verifier algorithm is not handling empty proofs that pass through empty branches properly. Fix coming soon.

Liamsi:
Just saw your comment. Awesome!

gdbelvin:
It WORKS! 💯

PR #729, opened 2017-08-04
Remove dead code

### LINKED ISSUE #714: Remove outdated Merkle tree implementation
key-transparency includes a bunch of code whose tasks now could and should be handled by trillian.

includes removing: 
- core/tree/sparse (Merkle tree implementation)
- signing key from signer
- ...

gdbelvin:
- core/tree is now subsumed by Trillian
- keytransparency-signer no longer needs a private key. Trillian signs.

Closes #714


docker-compose.yml
```diff
@@ -124,8 +124,6 @@ services:
       DB_USER: test
       DB_PASSWORD: zaphod
       LOG_KEY: /kt/trillian-log.pem
-      SIGN_KEY: /kt/p256-key.pem
-      SIGN_KEY_PW: towel
```
Liamsi: Can you also update the kubernetes [config](https://github.com/google/keytransparency/blob/master/deploy/kubernetes/keytransparency-deployment.yml.tmpl#L123-L126)? Otherwise this will break the deployed server.

cesarghali: :+1:

gdbelvin: done

PR #732, opened 2017-08-08
Simplify presubmit with gometalinter

### LINKED ISSUE #707: Cleaner presubmit script
Cleanup presubmit script and use `govendor` for `fmt, build, install, test, vet, generate` with `+local`.

gdbelvin:
This replaces all existing presubmit logic with `gometalinter`. 
Additional presubmit checks can be turned on in the future. 

The only intended deviation is the removal of `errcheck` from presubmit.  The reason is performance: `gometalinter` creates a new process for each package, yet `errcheck` compiles all dependencies, producing a 60x slowdown and numerous timeouts. 

The custom license rule is blocked on alecthomas/gometalinter#332

Closes #707


metalinter.json
```diff
@@ -0,0 +1,13 @@
+{
+  "Linters": {
+	      "errcheck": "errcheck -abspath -ignore 'Close|Write|Serve' {path}:PATH:LINE:COL:MESSAGE",
```
cesarghali: errcheck currently ignores `'Close|Write|Serve,os:Remove'`. Not sure if `os:Remove` is used.

metalinter.json
```diff
@@ -0,0 +1,13 @@
...
+	      "license": "grep -rL 'Apache License' {path}:^(?P<path>.*?\\.go.*)"
+             },
+  "Severity": {"license": "error"},
+  "Enable": ["gofmt", "golint", "vet", "ineffassign", "misspell", "gocyclo"],
+  "Vendor": true,
+  "VendoredLinters": true,
```
cesarghali: I'm assuming `Vendor` means use code in vendor directory, right? Does `VendoredLinters` mean that the linters are in the vendor directory? Just out of curiosity.

Makefile
```diff
@@ -21,37 +21,27 @@
 
 # TODO: Makefile will be deleted once the repo is public. Check issue #411.
 
-main: client
-	go build ./cmd/keytransparency-server
-	go build ./cmd/keytransparency-signer
+main: 
+	go build ./cmd/keytransparency-server ./cmd/keytransparency-signer ./cmd/keytransparency-client
```
gdbelvin: Hmm.  I think the useful error is what file had the problem, not necessarily which binary. 
The idea is that we'll get a speedup from only compiling the common dependencies only once.

metalinter.json
gdbelvin: `gometalinter` splits this string on the first ":". Is there a way we can use a different string for that? 

Also, I've currently turned off "errcheck" because it's pretty inefficient when called from metalinter.  I've kept this line to preserve our checks for when metalinter fixes this performance issue and we can turn `errcheck` back on.

metalinter.json
gdbelvin: - `Vendor` - means "this project is using vendoring" - ignore packages in the vendor folder.
- `VendoredLinters` means use the linters that were vendored by `gometalinter`. This is a bit of insurance against upstream changes in the linters themselves.

Makefile
cesarghali: Sounds good.

metalinter.json
cesarghali: I think this works for now. Once we enable errcheck and we start seeing os.Remove related errors, we can look for a fix.

metalinter.json
cesarghali: :+1:

PR #733, opened 2017-08-08
Turn on more presubmit checks

gdbelvin:
Based on #732 

This PR turns on the following linter checks along with simple code fixups. 

- deadcode
- varcheck detects unused global variables
- interfacer - not turned on due to larger code changes needed, but did implement a fixup
- megacheck - detects use of deprecated interfaces. 
- gas - security checks

### DIFF
--- a/core/signer/signer.go
+++ b/core/signer/signer.go
@@ -122,7 +122,7 @@ func (s *Signer) StartSigning(ctx context.Context, minInterval, maxInterval time
 	last := time.Unix(0, mapRoot.GetTimestampNanos())
 	// Start issuing epochs:
 	clock := util.SystemTimeSource{}
-	tc := time.Tick(minInterval)
+	tc := time.NewTicker(minInterval).C
 	for f := range genEpochTicks(clock, last, tc, minInterval, maxInterval) {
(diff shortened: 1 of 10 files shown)


Liamsi:
Does this depend on #732 (or is #732 a part of this pull)?

gdbelvin:
This PR currently contains #732, which should be merged first.

core/signer/signer.go
```diff
@@ -122,7 +122,7 @@ func (s *Signer) StartSigning(ctx context.Context, minInterval, maxInterval time
 	last := time.Unix(0, mapRoot.GetTimestampNanos())
 	// Start issuing epochs:
 	clock := util.SystemTimeSource{}
-	tc := time.Tick(minInterval)
+	tc := time.NewTicker(minInterval).C
```
Liamsi: Did one of the linter checks recommend this change? `time.Tick` is just a wrapper around `time.NewTicker().C`

gdbelvin: Indeed it did. I forget which linter, but the it was complaining that `Tick` leaked the underlying object and explicitly recommended using `NewTicker`

Issue #712, opened 2017-07-26
Update Dockerfiles


Liamsi:
Update all Dockerfiles to match/follow @witoff's changes after https://github.com/google/trillian/pull/667 is merged.

Also, remove libtool dependency as soon as https://github.com/google/trillian/pull/734 is merged (related to https://github.com/google/keytransparency/issues/687).

PR #738, opened 2017-08-09
Simplify command line client flags

### LINKED ISSUE #659: prepare_client.sh is out of date
The `prepare_client.sh` script is out of date. It will overwrite the `keytransparency.yml` client configuration with values that won't work from scratch (e.g. the `log-key`).

gdbelvin:
With the transition onto Trillian, several of the previous flags are now obsolete. 
This PR cleans up and clarifies command line flags.  

This PR also fixes a few things in the client and cleans up docker-compose. 

References:
#712
Closes #659


scripts/prepare_client.sh
```diff
@@ -14,70 +14,45 @@
...
+if [[ -n "${SIGKEY}" ]]; then
+    printf "log-key: %s\n" "${LOGKEY}" >> .keytransparency.yaml
 fi
 
-read -p "Key Transparency URL and port, i.e. url:port, (default=${KTURLDEFAULT}): " KTURL
-if [[ -z "${KTURL}" ]]; then
-    KTURL="${KTURLDEFAULT}"
+read -p "Trillian Map verification key: " MAPKEY
+if [[ -n "${MAPKEY}" ]]; then
+    printf "map-key: %s\n" "${LOGKEY}" >> .keytransparency.yaml
```
Liamsi: Change to  `MAPKEY`?

.keytransparency.yaml
```diff
@@ -1,8 +1,4 @@
-log-key: "genfiles/trillian-log.pem"
-vrf:    "genfiles/vrf-pubkey.pem"
-kt-key: "genfiles/server.crt"
-kt-sig: "genfiles/p256-pubkey.pem"
-domain: "example.com"
-kt-url: "35.184.134.53:8080"
-client-secret: "client_secret.json"
-service-key: ""
+
```
Liamsi: Remove this new line.

.keytransparency.yaml
```diff
@@ -1,8 +1,4 @@
...
+kt-url: localhost:8080
+client-secret: 
+service-key:
```
Liamsi: This file looks like it does not work as it is. Also, and it gets completely overwritten below (in prepare_client.sh). 
Is it still useful? Or should we delete it completely?

scripts/prepare_client.sh
gdbelvin: thanks

.keytransparency.yaml
gdbelvin: done

.keytransparency.yaml
gdbelvin: This empty file allows the command line client to use the defaults specified in its command line flags, which do work - and it removes another place where we need to manage defaults.

gdbelvin:
Updated PTAL

.keytransparency.yaml
Liamsi: Thanks for the explanation. LGTM

PR #739, opened 2017-08-09
Update mutator interface to accept proto objects directly

Liamsi:
This PR changes 
```golang
type Mutation interface {
    Mutate(value, mutation []byte ([]byte, error)
}
``` 
to
```golang
type Mutation interface {
    Mutate(value, mutation proto.Message) ([]byte, error)
}
```

This PR also removes `Replacer` which isn't used anywhere.


core/mutator/mutator.go
```diff
@@ -47,9 +47,9 @@ var (
 
 // Mutator verifies mutations and transforms values in the map.
 type Mutator interface {
-	// CheckMutation verifies that this is a valid mutation for this item and
+	// Mutate verifies that this is a valid mutation for this item and
 	// applies mutation to value.
-	Mutate(value, mutation []byte) ([]byte, error)
```
gdbelvin: To make this more generic, should we consider making the mutation a generic protobuf?

Liamsi: like `proto.Any`?

gdbelvin: like `proto.Message`

Liamsi: Done

Liamsi:
We currently `objecthash.ObjectHash(nil)` if the previous entry does not exist. I've added a little helper function `FromLeafValue` to have a central place to consistently convert what we have in `LeafValue`.
PTAL

core/client/kt/requests.go
```diff
@@ -75,7 +77,7 @@ func (v *Verifier) CreateUpdateEntryRequest(
 	if err != nil {
 		return nil, err
 	}
-	previous := objecthash.ObjectHash(getResp.GetLeafProof().GetLeaf().GetLeafValue())
+	previous := objecthash.ObjectHash(prevEntry)
```
gdbelvin: Please add a TODO to change this hash to plain SHA256. 
The previous entry is stored as `[]byte` and therefore we should hash it directly.

core/mutator/entry/entry_test.go
```diff
@@ -71,18 +72,21 @@ func createEntry(commitment []byte, pkeys []string) ([]byte, error) {
...
-		return nil, fmt.Errorf("Marshal(%v)=%v", entry, err)
-	}
-	return entryData, nil
+	}, nil
 }
 
-func prepareMutation(key []byte, entryData []byte, previous []byte, signers []signatures.Signer) ([]byte, error) {
+func prepareMutation(key []byte, newEntry *tpb.Entry, previous []byte, signers []signatures.Signer) (*tpb.SignedKV, error) {
+	var entryData []byte
+	if newEntry != nil {
```
gdbelvin: It should not be valid to set `newEntry` with `nil`. Please return an error in this case.

core/mutator/entry/entry_test.go
```diff
@@ -183,13 +182,35 @@ func TestCheckMutation(t *testing.T) {
...
-			t.Errorf("Mutate(%v, %v)=%v, want %v", tc.oldValue, mutation, got, tc.err)
+func TestFromLeafValue(t *testing.T) {
+	entry := &tpb.Entry{Commitment: []byte{1, 2}}
+	entryB, _ := proto.Marshal(entry)
+	for i, tc := range []struct {
+		leafVal []byte
+		want    *tpb.Entry
+		wantErr bool
+	}{
+		{make([]byte, 0), nil, false},            // empty leaf -> nil, no error
```
gdbelvin: Add a test case for both `nil` and `[]byte{}` and make sure they produce different values

core/client/kt/requests.go
Liamsi: Thanks, I've added a TODO. 
I'm not sure what the second part of the comment means: `prevEntry` is of type  `Entry` and not []byte. Should I add a comment about this as well?

Liamsi:
Thanks for your review! I hope I addressed all your comments (besides the 2nd part of https://github.com/google/keytransparency/pull/739#discussion_r132683130). PTAL

core/mutator/entry/entry.go
```diff
@@ -41,25 +42,41 @@ func New() *Entry {
...
-	update := new(tpb.SignedKV)
-	if err := proto.Unmarshal(mutation, update); err != nil {
-		return nil, err
-	}
-
+func (*Entry) Mutate(oldValueM, updateM proto.Message) ([]byte, error) {
 	// Ensure that the mutation size is within bounds.
-	if proto.Size(update) > mutator.MaxMutationSize {
-		glog.Warningf("mutation (%v bytes) is larger than the maximum accepted size (%v bytes).", proto.Size(update), mutator.MaxMutationSize)
+	if proto.Size(updateM) > mutator.MaxMutationSize {
```
gdbelvin: go style typically discourages putting type letters in variable names.

core/mutator/entry/entry.go
```diff
@@ -41,25 +42,41 @@ func New() *Entry {
...
+		var entryData []byte
+		if oldEntry != nil {
+			var err error
+			if entryData, err = proto.Marshal(oldEntry); err != nil {
+				return nil, fmt.Errorf("Marshal(%v)=%v", oldEntry, err)
+			}
+		}
 		// Check if this mutation is a replay.
-		if bytes.Equal(oldValue, update.GetKeyValue().Value) {
+		if bytes.Equal(entryData, update.GetKeyValue().Value) {
```
gdbelvin: `proto.Marshal` is not required to produce consistent byte output.  Proto's binary encoding for example, could reorder tags and values.  Could this be upgraded to a `proto.Equal`?

Liamsi:
Thanks for the review! I've updated to `proto.Equal` and addressed the other comments. PTAL

core/mutator/entry/entry.go
```diff
@@ -41,51 +42,75 @@ func New() *Entry {
...
 	}
 
+	updated, ok := update.(*tpb.SignedKV)
+	if !ok {
+		glog.Warning("received proto.Message is not of type *tpb.SignedKV.")
+		return nil, fmt.Errorf("updateM.(*tpb.SignedKV): _, %v", ok)
+	}
+	var oldEntry *tpb.Entry
+	if oldValue != nil {
+		oldEntry, ok = oldValue.(*tpb.Entry)
```
gdbelvin: `oldEntry` shadows the outer variable.  It's safer to declare a new variable name, and then set set `outer = inner`

Issue #746, opened 2017-08-11
CI: Don't reset database with every pull request


gdbelvin:
Let's leave the continuous integration database up so we can start to accumulate interesting history.

Liamsi:
I opened https://github.com/google/keytransparency/pull/747 for this. 

Not sure this is enough for accumulating interesting history though: The (map- log-signing) keys still get regenerated with every PR (and old ones get deleted).

PR #749, opened 2017-08-13
CI: Store generated keys in secret and only regenerate if deleted (manually)

### LINKED ISSUE #748: CI: Don't regenarate keys with every pull request
Same reasoning as in https://github.com/google/keytransparency/issues/746 but for the VRF key (and maybe the TLS key while we're at it).

One simple way to achieve this is to generate the keys once and use [secrets](https://kubernetes.io/docs/concepts/configuration/secret/). The deploy script should create a new secret only if it doesn't exist yet (this means keys won't be updated with each PR).

Liamsi:
This changes the CI environment to only generate keys once (instead of regenerating with every PR).
Additionally, this PR changes the deploy step to rely on `kubectl apply`. It does not call `kubectl delete` anymore.

This PR has the same reasoning as #746 behind it (accumulate interesting history).

This builds on top of #747 (which should be merged first)

Resolves #748


Liamsi:
As discussed, I've removed `kubectl delete` from the deploy script. To make it explicit that kubernetes should pull a fresh image (and restart the pods), I added `imagePullPolicy: Always` to kt and trillian configs. The [default](https://github.com/kubernetes/kubernetes/blob/master/pkg/api/v1/defaults.go#L81-L92) [behavior](https://kubernetes.io/docs/concepts/configuration/overview/#container-images) is to pull images only if they are not present yet.

PR #752, opened 2017-08-14
Client/domaininfo

gdbelvin:
Configure the client using the info published at `v1/domain/info` 

This saves users from needing to do a lot of complicated configuration. 

Changes:
- Adds an `--autoconfig` flag, defaulting to `true`, which reads configuration information from `v1/domain/info`
- Adds an `--insecure` flag in case users want to connect to a server with a self-signed certificate.
- uses `crypto.PublicKey` as the return type for `vrf.Public()`


cmd/keytransparency-client/cmd/root.go
```diff
@@ -77,13 +79,18 @@ func init() {
...
-	RootCmd.PersistentFlags().String("kt-url", "", "URL of Key Transparency server")
+	RootCmd.PersistentFlags().String("kt-url", "35.184.134.53:8080", "URL of Key Transparency server")
 	RootCmd.PersistentFlags().String("kt-cert", "genfiles/server.crt", "Path to public key for Key Transparency")
+	RootCmd.PersistentFlags().Bool("autoconfig", true, "Fetch config info from the server's /v1/domain/info")
+	RootCmd.PersistentFlags().Bool("insecure", false, "Skip TLS checks")
+
 	RootCmd.PersistentFlags().String("vrf", "genfiles/vrf-pubkey.pem", "path to vrf public key")
 
 	RootCmd.PersistentFlags().String("log-key", "genfiles/trillian-log.pem", "Path to public key PEM for Trillian Log server")
 	RootCmd.PersistentFlags().String("map-key", "genfiles/trillian-map.pem", "Path to public key PEM for Trillian Map server")
```
Liamsi: Suggestion: Maybe add a comment stating that these flags will be set automatically if `--autoconfig=true`? Or, even add this to the viper description string. Otherwise, it might be unclear what will be set automatically and what not.

(If `--autoconfig=true`, which is default, it doesn't matter what you set for the `vrf`, `log-key`, and `map-key`)

cmd/keytransparency-client/cmd/root.go
```diff
@@ -77,13 +79,18 @@ func init() {
...
 
+	RootCmd.PersistentFlags().String("client-secret", "", "Path to client_secret.json file for user creds")
+	RootCmd.PersistentFlags().String("service-key", "", "Path to service_key.json file for anonymous creds")
 	RootCmd.PersistentFlags().String("fake-auth-userid", "", "userid to present to the server as identity for authentication. Only succeeds if fake auth is enabled on the server side.")
```
Liamsi: Suggestion: Document the "priority" here as well? If `fake-auth-userid` is set, `client-secret` and `service-key` will be ignored, etc. I wonder if viper provides and automatic way to warn you that it doesn't make sense to set all of these 3.

cmd/keytransparency-client/cmd/root.go
```diff
@@ -169,48 +166,67 @@ func getServiceCreds(serviceKeyFile string) (credentials.PerRPCCredentials, erro
...
+
+	switch {
+	case insecure: // Impatient insecure.
+		return credentials.NewTLS(&tls.Config{
+			InsecureSkipVerify: true,
+		}), nil
+
+	case ktCert != "": // Custom CA Cert.
+		return credentials.NewClientTLSFromFile(ktCert, host)
+
```
Liamsi: Remove newline.

cmd/keytransparency-client/cmd/root.go
```diff
@@ -169,48 +166,67 @@ func getServiceCreds(serviceKeyFile string) (credentials.PerRPCCredentials, erro
...
+	if err != nil {
+		return nil, err
+	}
+
+	switch {
+	case insecure: // Impatient insecure.
+		return credentials.NewTLS(&tls.Config{
+			InsecureSkipVerify: true,
+		}), nil
+
```
Liamsi: remove superfluous newline?

gdbelvin: the newlines visually break up the cases

Issue #745, opened 2017-08-11
Change hash function back pointer to be SHA256


gdbelvin:
Since the map stores leaf objects as `[]byte` it doesn't make sense to use ObjectHash. We can just use a regular hash function.

Liamsi:
Thanks for chiming in @benlaurie :-) 

Currently, the trillian map represents/stores leaf data in [`[]byte`](https://github.com/google/trillian/blob/e11b1d5427f8e8505bf8713b8e434f125449d74e/trillian_map_api.proto#L31) (independent from what you store there). My understanding is a back pointer to a leaf should be a hash of these raw bytes in the leaf (for which we could use sha256). 
Are you suggesting we should change from bytes to some other non-marshaled, general representation of the data on the trillian side (and hash this for a back pointer)?

gdbelvin:
Thanks for your input @benlaurie!  

To my thinking this seems to be a question of semantics?  The Trillian map stores leaf data as `[]byte` as @Liamsi mentioned.  We have a couple choices for making a hash chain for each leaf revision:
1. Hash of the previous revision's `[]byte` representation.  
   - Easy to understand, hard to mess up.
   - Forces the hash to change if there's a change in representation. eg. redacting a field, re-ordering proto elements, etc. 
2. Hash of the previous revision's ObjectHash. 
   - Forces object serialization to JSON or at least a decode/encode step through JSON
       - Introduces the possibility of accidentally dropping unknown fields. 
   - Hash wouldn't change on field reordering, redaction etc. 

To my intuition, the option of using the raw bytes seems stricter and safer.

gdbelvin:
Fixed some timeago

Issue #387, opened 2016-10-21
Efficient account history


gdbelvin:
The current naive implementation requires clients to download all epochs when verifying their account history. This is too resource intensive to be practical.  This proposal makes this practical. 
- Include account history in the account leaf node. 
- Support generic monitoring by third parties.
- Add monitoring logic to verify accurate account histories.

PR #758, opened 2017-08-15
Store EntryUpdate in MapLeaf

### LINKED ISSUE #757: Store EntryUpdate in map leaf
In order to verify the correctness of a leaf update, we need to store the policy and it's authorizing signatures in the leaf node.

Liamsi:
And and use sha256 instead of objecthash for back pointers

~~Note to myself: clarify if we want to store the map index in the leaf as well. Then we could simply reuse `SignedKV`~~  Yes, the signature is on the the whole struct, which includes the key/index.

This is part of #387 

resolves #757 #745


cmd/keytransparency-client/grpcc/grpc_client.go
```diff
@@ -26,6 +26,10 @@ import (
 	"log"
 	"time"
 
+	"github.com/golang/protobuf/proto"
```
gdbelvin: we are putting 3rd party imports below the key transparency and trillian imports

cmd/keytransparency-client/grpcc/grpc_client.go
```diff
@@ -277,7 +275,11 @@ func (c *Client) Retry(ctx context.Context, req *tpb.UpdateEntryRequest) error {
 	}
 
 	// Check if the response is a replay.
-	if got, want := updateResp.GetProof().GetLeafProof().Leaf.LeafValue, req.GetEntryUpdate().GetUpdate().GetKeyValue().GetValue(); !bytes.Equal(got, want) {
+	serverKV := new(tpb.SignedKV)
+	if err := proto.Unmarshal(updateResp.GetProof().GetLeafProof().Leaf.LeafValue, serverKV); err != nil {
```
gdbelvin: - GetLeaf().GetLeafValue
- use of bytes is more strict and better here.

core/client/kt/verify.go
```diff
@@ -75,8 +75,8 @@ func New(vrf vrf.PublicKey,
 func (v *Verifier) VerifyGetEntryResponse(ctx context.Context, userID, appID string,
 	trusted *trillian.SignedLogRoot, in *tpb.GetEntryResponse) error {
 	// Unpack the merkle tree leaf value.
-	entry := new(tpb.Entry)
-	if err := proto.Unmarshal(in.GetLeafProof().GetLeaf().GetLeafValue(), entry); err != nil {
+	entry, err := entry.FromLeafValue(in.GetLeafProof().GetLeaf().GetLeafValue())
```
gdbelvin: this is no longer an entry?  should we call it something else?

core/keyserver/keyserver.go
```diff
@@ -16,6 +16,15 @@
 package keyserver
 
 import (
+	"github.com/golang/glog"
+	"github.com/golang/protobuf/proto"
```
gdbelvin: move below key transparency imports?

gdbelvin:
Sorry for the whiparound here. After an offline discussion we decided:
1. We should keep mutations and leaf data separate for the monitor
2. The back pointer needs to be moved into the leaf data. 
3. We also need to move signatures into leaf data.

PR #759, opened 2017-08-17
Use log tree size as current revision

### LINKED ISSUE #753: Bug: Log Inclusion proofs sometimes fail
Fetching the log inclusion proof sometimes fails during an update. Here's an example: 

```
E0815 12:10:56.542003       6 keyserver.go:184] tlog.GetInclusionProof(5806818483210672871, 8, 7): rpc error: code = InvalidArgument desc = GetInclusionProofRequest.LeafIndex: 7 >= TreeSize: 7, want <
```

I suspect that this occurs doing the brief period of time between a map update succeeding but before the new map head has been sequenced by the log. 

Perhaps we should read the log first to obtain the latest signed map root revision?

gdbelvin:
Fixes #753

By using the log as the authoritative source for the latest revision, we
avoid the scenario where by the map has produced a new revision which
has not been sequenced (yet) by the log.

During this window of a time, a request for the latest revision produces
a log inclusion proof error.


core/keyserver/keyserver.go
```diff
@@ -97,17 +97,34 @@ func (s *Server) GetEntry(ctx context.Context, in *tpb.GetEntryRequest) (*tpb.Ge
...
+		})
+	if err != nil {
+		glog.Errorf("tlog.GetLatestSignedLogRoot(%v): %v", s.logID, err)
+		return nil, grpc.Errorf(codes.Internal, "Cannot fetch SignedLogRoot")
+	}
+	// Use the log as the athoritative source of the latest revision.
+	if revision < 0 {
+		maxLogIndex := logRoot.GetSignedLogRoot().GetTreeSize() - 1
+		// Convert log index to map revision.
+		revision = maxLogIndex + 1
```
Liamsi: map revision and log index are not one off anymore

gdbelvin:
Updated. PTAL

PR #760, opened 2017-08-17
Use a simple grpc connection to Trillian backends

gdbelvin:
Previously, the signer used a verifying trillian log client in the
signer. This:
- Required a separate build step to acquire the verifying key
- Was tied to a confusing and only partially implemented multi-tennant
  configuration system.

This PR simplifies the situation to make a pure grpc call to the
trillian log backend, simplifying debugging, and makes things consistent
consistent with the way we are querying the trillian map server, and is
also consistent with the way the Certificate Transparency frontends
treat their own trillian backends.

This PR also removes the dead code associated with these defunct interfaces.


cmd/keytransparency-signer/main.go
```diff
@@ -86,17 +79,12 @@ func main() {
 	tmap := trillian.NewTrillianMapClient(mconn)
 
 	// Connection to append only log
-	tlog, err := config.LogClient(*logID, *logURL, *logPubKey)
+	lconn, err := grpc.Dial(*logURL, grpc.WithInsecure())
```
cesarghali: Is this on the same machine? If yes, will it always be on the same machine? If not, why did we switch to an insecure connection?

core/signer/signer.go
```diff
@@ -310,9 +307,20 @@ func (s *Signer) CreateEpoch(ctx context.Context, forceNewEpoch bool) error {
...
 	// Put SignedMapHead in an append only log.
-	if err := s.sths.Write(ctx, s.logID, revision, setResp.GetMapRoot()); err != nil {
-		// TODO(gdbelvin): If the log doesn't do this, we need to generate an emergency alert.
-		return fmt.Errorf("sths.Write(%v, %v): %v", s.logID, revision, err)
+	smrJSON, err := json.Marshal(setResp.GetMapRoot())
+	if err != nil {
+		return err
+	}
+	// TODO(gbelvin): Add leaf at a specific index. trillian#423
+	// TODO(gdbelvin): If the log doesn't do this, we need to generate an emergency alert.
```
cesarghali: nit: gbelvin? :)

cmd/keytransparency-signer/main.go
gdbelvin: The trillian log and map are not always on the same machine.

Using `grpc.WithInsecure()` is not a change.  It was previously hidden behind `config`.  This makes it a bit more obvious and consistent.

core/signer/signer.go
gdbelvin: on github, it's gdbelvin, confusing I know.

PR #761, opened 2017-08-17
Fix/756 Fix ListHistory command

### LINKED ISSUE #703: Consistent index start
Provide a function (or several) in a central place that takes care of indices/revisions/numbers/epochs such that epochs can start at 0 without error'ing and without making the dev think about it in every single call.

### LINKED ISSUE #756: ListHistory fails
Due to epochs expecting to start at 1 and the default start epoch in the client [being 0](https://github.com/google/keytransparency/blob/5a851ef25bd0036b4b16c2ee383c8de7ca56ae8c/cmd/keytransparency-client/cmd/history.go#L32) `ListHistory` fails with:
`2017/08/15 15:24:39 ListHistory failed: start=0, want > 0`

If the default start epoch is changed to 1 the command still fails with:
`Error: ListHistory failed: incomplete account history`. 

This is due to `end` epoch being automatically being set to the [latest](https://github.com/google/keytransparency/blob/5a851ef25bd0036b4b16c2ee383c8de7ca56ae8c/cmd/keytransparency-client/cmd/history.go#L64) for this entry (which is OK) and a combination of how the  [pagesize](https://github.com/google/keytransparency/blob/5a851ef25bd0036b4b16c2ee383c8de7ca56ae8c/cmd/keytransparency-client/grpcc/grpc_client.go#L185) is set and [this line](https://github.com/google/keytransparency/blob/5a851ef25bd0036b4b16c2ee383c8de7ca56ae8c/cmd/keytransparency-client/grpcc/grpc_client.go#L210)

gdbelvin:
Cleans up a few off by on errors and poor stopping condition logic in the list history command.

Fixes #756 
Rebased on #760 and #762. Merge them first.

### DIFF
--- a/core/signer/signer.go
+++ b/core/signer/signer.go
@@ -113,6 +113,7 @@ func (s *Signer) Initialize(ctx context.Context) error {
 	// add the empty map root to the log.
 	if logRoot.GetSignedLogRoot().GetTreeSize() == 0 &&
 		mapRoot.GetMapRoot().GetMapRevision() == 0 {
+		glog.Infof("Initializing Trillian Log with empty map root")
 		if err := queueLogLeaf(ctx, s.tlog, s.logID, mapRoot.GetMapRoot()); err != nil {
 			return err
 		}
--- a/cmd/keytransparency-client/cmd/history.go
+++ b/cmd/keytransparency-client/cmd/history.go
@@ -98,6 +98,6 @@ func (m mapHeads) Less(i, j int) bool { return m[i].MapRevision < m[j].MapRevisi
 func init() {
 	RootCmd.AddCommand(histCmd)
 
-	histCmd.PersistentFlags().Int64Var(&start, "start", 0, "Start epoch")
+	histCmd.PersistentFlags().Int64Var(&start, "start", 1, "Start epoch")
 	histCmd.PersistentFlags().Int64Var(&end, "end", 0, "End epoch")
 }
--- a/cmd/keytransparency-client/grpcc/grpc_client.go
+++ b/cmd/keytransparency-client/grpcc/grpc_client.go
@@ -170,14 +170,15 @@ func min(x, y int32) int32 {
 
 // ListHistory returns a list of profiles starting and ending at given epochs.
 // It also filters out all identical consecutive profiles.
-// Epochs start at 1.
 func (c *Client) ListHistory(ctx context.Context, userID, appID string, start, end int64, opts ...grpc.CallOption) (map[*trillian.SignedMapRoot][]byte, error) {
-	if start <= 0 {
-		return nil, fmt.Errorf("start=%v, want > 0", start)
+	if start < 0 {
+		return nil, fmt.Errorf("start=%v, want >= 0", start)
 	}
 	var currentProfile []byte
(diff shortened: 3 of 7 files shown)


core/signer/signer.go
```diff
@@ -63,40 +64,80 @@ func init() {
...
+		tlog:      tlog,
 		mutator:   mutator,
 		mutations: mutations,
 		factory:   factory,
 	}
 }
 
+// Initialize inserts the object hash of an empty struct into the log if it is empty.
+// This keeps the log leaves in-sync with the map which starts off with an
+// empty log root at map revision 0.
```
Liamsi: That is a cool and simple idea!

core/signer/signer.go
```diff
@@ -63,40 +64,80 @@ func init() {
...
+	if err != nil {
+		return fmt.Errorf("GetSignedMapRoot(%v): %v", s.mapID, err)
+	}
+
+	// If the tree is empty and the map is empty,
+	// add the empty map root to the log.
+	if logRoot.GetSignedLogRoot().GetTreeSize() == 0 &&
+		mapRoot.GetMapRoot().GetMapRevision() == 0 {
+		smrJSON, err := json.Marshal(mapRoot.GetMapRoot())
+		idHash := sha256.Sum256(smrJSON)
```
Liamsi: What is the reason behind not calling ObjectHash() on the map root?

core/signer/signer.go
```diff
@@ -310,9 +351,22 @@ func (s *Signer) CreateEpoch(ctx context.Context, forceNewEpoch bool) error {
...
+	// TODO(gdbelvin): If the log doesn't do this, we need to generate an emergency alert.
+	if _, err := s.tlog.QueueLeaf(ctx, &trillian.QueueLeafRequest{
+		LogId: s.logID,
+		Leaf: &trillian.LogLeaf{
+			LeafValue:        smrJSON,
+			LeafIdentityHash: idHash[:],
+		},
+	}); err != nil {
+		return fmt.Errorf("trillianLog.QueueLeaf(logID: %v, leaf: %v): %v",
+			s.logID, smrJSON, err)
```
Liamsi: This is exactly the same code as in https://github.com/google/keytransparency/pull/761/files#diff-98fddaa13ed78860afbe32b4e92dc431R116 (only the two TODOs are different).

What do you think about refactoring this into a private helper `writeToLog(tlog *LogClient, smr *trillian.SignedMapRoot)`

README.md
```diff
@@ -96,7 +96,6 @@ source scripts/configure_trillian.sh && createLog && createMap
 ```
 
 3. Run Key Transparency
-- `docker-compose build kt-signer`
```
Liamsi: This looks like we might need to update the deploy script. I'll take care of that in case this is necessary.

core/signer/signer.go
gdbelvin: The log server sets the leaf hash for us. It's only duplicate work on our end to do it here.

core/signer/signer.go
gdbelvin: Good call.  Eventually we will replace this code with a proper Verifying Log Client that auto configures itself

README.md
gdbelvin: yep

PR #762, opened 2017-08-17
Fix/703 Initialize log with empty SignedMapRoot

### LINKED ISSUE #703: Consistent index start

gdbelvin:
This brings the relationship between MapRevisions and Log indexes back into parity.

Previously log index = map revision -1
Now, log index = map revision

Fixes #703
Rebased on #760. Merge that first.


Liamsi:
How does this relate to https://github.com/google/keytransparency/pull/761?

gdbelvin:
I made a bit of a confusing swap with the PRs.  This is actually the one that you just LGTM'ed
#761 contains an additional fix the the ListHistory command.

PR #764, opened 2017-08-18
Rename: keytransparency-signer to keytransparency-sequencer

gdbelvin:
After the migration to Trillian, the keytransparency-signer doesn't actually _sign_ anything anymore. 
Therefore it makes sense to give it a new name for it's real job: sequencing and batching mutations into Map Revisions.


.gitignore
```diff
@@ -3,7 +3,7 @@
 *~
 
 /keytransparency-server
-/keytransparency-signer
+/keytransparency-sequencer
```
Liamsi: Due to this change the `keytransparency-signer` made it into the repo.

gdbelvin:
Updated PTAL

Issue #588, opened 2017-05-23
Batch APIs


gdbelvin:
APIs:
- [x] Batch GetUser
  Supports looking up {self, others} in one go. 
- [x] Batch UpdateEntry
   Assumes that user authentication has occurred upstream. Authentication and authorization may be implemented as grpc interceptors.

gdbelvin:
New use-case: a single user updating multiple trees with a single API end point.

jtoohill:
Is a batch API needed for that? IIUC there's no transactionality or log consistency required across trees.

What I would like a batch API for is looking up multiple leaves in a given tree in a single request, which should return entries all in the same epoch (and only requiring a single consistency proof).

gdbelvin:
You're right. That's not an actual use case. I'll add a batch API for lookups to the top.

PR #778, opened 2017-08-22
Define API for batch operations

gdbelvin:
- BatchGetEntries will return full proofs for a list of users
- BatchUpdateEntries accepts a list of UpdateEntryRequests
- BatchSetEntries does the work a client would do, creating
  UpdateEntryRequests, submitting them, and waiting for them to be
  integrated before returning.

#588


impl/proto/keytransparency_v1_service/keytransparency_v1_service.proto
```diff
@@ -70,3 +70,30 @@ service KeyTransparencyService {
     option (google.api.http) = { get: "/v1/domain/info" };
   }
 }
+
+// The KeyTransparencyBackendService API provides batch access to the directory of public keys. 
+service KeyTransparencyBackendService {
+  // BatchGetEntries returns full proofs for a set of user_ids.
+  rpc BatchGetEntries(keytransparency.v1.types.BatchGetEntriesRequest) returns (keytransparency.v1.types.BatchGetEntriesResponse) {
```
cesarghali: I like this API :)

impl/proto/keytransparency_v1_service/keytransparency_v1_service.proto
```diff
@@ -70,3 +70,30 @@ service KeyTransparencyService {
...
+  rpc BatchUpdateEntries(keytransparency.v1.types.BatchUpdateEntriesRequest) returns (keytransparency.v1.types.BatchUpdateEntriesResponse) {
+    option (google.api.http) = {
+      post: "/v1/users"
+      body: ""
+    };
+  }
+
+  // BatchSetEntries uses an authorized_public key to perform a set request on multiple 
+  // entries at once.  
+  rpc BatchSetEntries(keytransparency.v1.types.BatchSetEntriesRequest) returns (keytransparency.v1.types.BatchSetEntriesResponse) {
```
cesarghali: What's the difference between Set and Update apis? I can see in the description that the latter accepts and the former does the work, but not why doing it all at once?

gdbelvin: Perhaps I need better names, but the idea is that `BatchUpdate` is just a batch version of `Update`, meaning that it is pretty dumb and just puts things in the mutation table. 

`Set` is really more of a server side client that has been equipped with a server `authorized_key`  Perhaps this needs a better name like `AccountResetService` or something like that.  In order to make integrations with other backend services easier, I thought it would be best to make the API interface for updating accounts as simple as possible.  This in particular is an API that needs to be authorized.  The other two are just higher volume versions of the earlier methods.

core/proto/keytransparency_v1_types/keytransparency_v1_types.proto
```diff
@@ -251,3 +251,79 @@ message GetDomainInfoResponse {
...
+  // log_inclusion proves that smr is part of log_root at index=srm.MapRevision.
+  repeated bytes log_inclusion = 7;
+}
+
+message BatchUpdateEntriesRequest {
+  // mutations contains the set of mutations to put in the queue. 
+  repeated UpdateEntryRequest mutations = 1;
+} 
+
+message BatchUpdateEntriesResponse {
```
cesarghali: Don't we need to return the proofs?

core/proto/keytransparency_v1_types/keytransparency_v1_types.proto
```diff
@@ -251,3 +251,79 @@ message GetDomainInfoResponse {
   // Vrf contains the VRF public key.
   keyspb.PublicKey vrf = 3;
 }
+
+message BatchGetEntriesRequest {
```
cesarghali: Who are the users of this API?

impl/proto/keytransparency_v1_service/keytransparency_v1_service.proto
cesarghali: I agree with the name change. Reset would be more descriptive.

gdbelvin:
Simplified. PTAL

core/proto/keytransparency_v1_types/keytransparency_v1_types.proto
gdbelvin: This is a backend / trusting API.  If the API responds with a non-error message, it succeeded.

core/proto/keytransparency_v1_types/keytransparency_v1_types.proto
```diff
@@ -251,3 +251,26 @@ message GetDomainInfoResponse {
...
+  // This must be a key that this server has the private key for.
+  string key_id = 3;
+}
+
+
+// BatchSetEntriesResponse returns a list of users for which the set operation
+// was unsuccessful.
+message BatchSetEntriesResponse {
+  // errors is a map from user_ids to errors, if there was an error for that user.
+  map<string, string> errors = 1;
```
cesarghali: Should we return proofs here? This API should be like Update but in batches:
https://github.com/google/keytransparency/blob/master/core/proto/keytransparency_v1_types/keytransparency_v1_types.proto#L200

In its current form, the batch updater is trusting the server to do that right thing. This might not always be the case.

impl/proto/keytransparency_v1_service/keytransparency_v1_service.proto
```diff
@@ -70,3 +70,15 @@ service KeyTransparencyService {
     option (google.api.http) = { get: "/v1/domain/info" };
   }
 }
+
+// The KeyTransparencyAdminService API provides batch access to the directory of public keys. 
+service KeyTransparencyAdminService {
+  // BatchSetEntries uses an authorized_public key to perform a set request on multiple entries at once.  
+  rpc BatchSetEntries(keytransparency.v1.types.BatchSetEntriesRequest) returns (keytransparency.v1.types.BatchSetEntriesResponse) {
```
cesarghali: How do you feel about calling this `BatchUpdateEntries`. Set gives the impression that this API should be called once at the beginning, but update hints that it should be called every time there's a entry update.

core/proto/keytransparency_v1_types/keytransparency_v1_types.proto
gdbelvin: Yes, in it's current form the caller of the batch API trusts it to do the right thing.
Since this is a backend integration point, I thought this would be a reasonable trust model.

impl/proto/keytransparency_v1_service/keytransparency_v1_service.proto
gdbelvin: `BatchUpdateEntries` sounds good.

impl/proto/keytransparency_v1_service/keytransparency_v1_service.proto
```diff
@@ -70,3 +70,15 @@ service KeyTransparencyService {
     option (google.api.http) = { get: "/v1/domain/info" };
   }
 }
+
+// The KeyTransparencyAdminService API provides batch access to the directory of public keys.
```
cesarghali: nit: this might be longer than 80 characters, didn't measure it though.

impl/proto/keytransparency_v1_service/keytransparency_v1_service.proto
```diff
@@ -70,3 +70,15 @@ service KeyTransparencyService {
     option (google.api.http) = { get: "/v1/domain/info" };
   }
 }
+
+// The KeyTransparencyAdminService API provides batch access to the directory of public keys. 
+service KeyTransparencyAdminService {
+  // BatchSetEntries uses an authorized_public key to perform a set request on multiple entries at once.
```
cesarghali: Same here.

gdbelvin: While these are longer than 80 chars, they match the length of the line below them, so I thought it would be easier on the eye to keep them on one line

cesarghali: That's completely fine. There's no line limit but I thought we're keeping comments at 80 characters :)
If you think it looks good in your editor, I'm fine keeping it as is.

PR #786, opened 2017-08-23
Store SMHs in BFTKV

eozturk1:
Store SMHS in BFTKV using the HTTP API.

Please note that The BFTKV API Server that is connected to has the address http://docker.for.mac.localhost:6001/

docker.for.mac.localhost is used to access an address in the host machine.

### DIFF
--- a/core/sequencer/sequencer.go
+++ b/core/sequencer/sequencer.go
@@ -346,6 +346,15 @@ func (s *Sequencer) CreateEpoch(ctx context.Context, forceNewEpoch bool) error {
 		// TODO(gdbelvin): If the log doesn't do this, we need to generate an emergency alert.
 		return err
 	}
+	
+	// Store signed log root in BFTKV
+	logRoot, err := s.tlog.GetLatestSignedLogRoot(ctx, &trillian.GetLatestSignedLogRootRequest{
+		LogId: s.logID,
+	})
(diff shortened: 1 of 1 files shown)


core/sequencer/sequencer.go
```diff
@@ -350,9 +352,28 @@ func (s *Sequencer) CreateEpoch(ctx context.Context, forceNewEpoch bool) error {
 	mapUpdateHist.Observe(mapSetEnd.Sub(mapSetStart).Seconds())
 	createEpochHist.Observe(time.Since(start).Seconds())
 	glog.Infof("CreatedEpoch: rev: %v, root: %x", revision, setResp.GetMapRoot().GetRootHash())
+
+	// store smr in BFTKV
+	writeToBFTKV(string(s.logID) + "|" + string(revision), string(setResp.GetMapRoot().GetRootHash()))
```
gdbelvin: Please send the log roots rather than the map roots. Thanks!

eozturk1:
I signed it!

core/sequencer/sequencer.go
eozturk1: Monitors seem to be storing map roots. Wouldn't it be easier if BFTKV stores the map roots and monitors compare their version of the map roots with what is stored in BFTKV?

PR #788, opened 2017-08-23
Remote Signer implementation

gdbelvin:
This PR implements a authorized account update service.

The service holds a private key that it uses to sign `UpdateEntryRequests`. 
If authorized by each individual account, this service can make updates to their account on their behalf. Eg. This service could update accounts after performing SMS or email based verification. 

The service is structured in the following way: 
1. The service receives a batch account update request 
2. The service verifies that this request is authorized for this app and each user. 
3. The service creates a UpdateEntryRequests
4. The service queues the update requests
5. Return
--- 
The sequencer then:
1. Reads mutations out of the queue
2. Tests to see if they are valid, authorized mutations for any particular account.
3. Applies the mutations to the map and commits by sending the new map head to the log. 
4. Accounts are updated. 

#588

### DIFF
--- a/cmd/keytransparency-sequencer/adminapi.go
+++ b/cmd/keytransparency-sequencer/adminapi.go
@@ -37,7 +37,7 @@ import (
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/reflection"
 
-	ktpb "github.com/google/keytransparency/impl/proto/keytransparency_v1_service"
+	ktpb "github.com/google/keytransparency/core/proto/keytransparency_v1_grpc"
 	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
 )
 
--- a/cmd/keytransparency-sequencer/adminapi.go
+++ b/cmd/keytransparency-sequencer/adminapi.go
@@ -1,0 +1,141 @@
+// Copyright 2017 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"database/sql"
+	"flag"
+	"fmt"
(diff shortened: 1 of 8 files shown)


cmd/keytransparency-sequencer/adminapi.go
```diff
@@ -0,0 +1,142 @@
...
+	gwmux := runtime.NewServeMux()
+	if err := ktpb.RegisterKeyTransparencyAdminServiceHandlerFromEndpoint(ctx, gwmux, addr, dopts); err != nil {
+		return nil, err
+	}
+
+	return gwmux, nil
+}
+
+func adminAPI(
+	sqldb *sql.DB,
```
cesarghali: Please put at least one parameter on the first line. It's a bit hard to read this way.

core/mutator/entry/mutation.go
```diff
@@ -28,10 +28,10 @@ import (
 type Mutation struct {
 	userID, appID string
 	index         []byte
-	data, nonce   []byte
+	Data, Nonce   []byte
```
cesarghali: I prefer to keep these unexported and add getters.

core/mutator/entry/mutation.go
```diff
@@ -28,10 +28,10 @@ import (
...
 
 	prevEntry *tpb.Entry
-	entry     *tpb.Entry
+	Entry     *tpb.Entry
```
cesarghali: Sane here.

core/adminservice/server.go
```diff
@@ -0,0 +1,171 @@
...
+	getResp, err := s.tmap.GetLeaves(ctx, &trillian.GetMapLeavesRequest{
+		MapId:    s.mapID,
+		Index:    indexes,
+		Revision: -1,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// Create Update Requests.
```
cesarghali: Please move what's inside the look to its own unexported function.

cmd/keytransparency-sequencer/adminapi.go
gdbelvin: done

gdbelvin:
I've pulled out a bit of the refactor contained this PR into it's own pull request. Please review #817 first.

The rest of this PR has been cleaned up and is ready for a second pass.

core/adminservice/server.go
gdbelvin: done

core/mutator/entry/mutation.go
gdbelvin: done

core/mutator/entry/mutation.go
gdbelvin: done

PR #792, opened 2017-08-24
Create a mutation manipulation object / API

gdbelvin:
Previously the client code had a giant function for creating mutations. 
I realized that the process of creating a mutation is closely related to verifying a mutation. 

This PR creates
 - A mutation creator `entry.Mutation`
 - A mutation verifier / function`entry.Mutator`  Which verifies and applies a mutation to a previous entry in the tree. 

#588

### DIFF
--- a/cmd/keytransparency-client/grpcc/grpc_client.go
+++ b/cmd/keytransparency-client/grpcc/grpc_client.go
@@ -26,7 +26,7 @@ import (
 	"log"
 	"time"
 
-	ktclient "github.com/google/keytransparency/core/client/kt"
+	"github.com/google/keytransparency/core/client/kt"
 	"github.com/google/keytransparency/core/crypto/signatures"
 	"github.com/google/keytransparency/core/crypto/vrf"
 	"github.com/google/keytransparency/core/crypto/vrf/p256"
@@ -79,7 +79,7 @@ var (
 type Client struct {
 	cli        spb.KeyTransparencyServiceClient
 	vrf        vrf.PublicKey
-	kt         *ktclient.Verifier
+	kt         *kt.Verifier
 	mutator    mutator.Mutator
 	RetryCount int
 	RetryDelay time.Duration
@@ -131,7 +131,7 @@ func New(cc *grpc.ClientConn,
 	return &Client{
 		cli:        spb.NewKeyTransparencyServiceClient(cc),
 		vrf:        vrf,
-		kt:         ktclient.New(vrf, mapHasher, mapPubKey, logVerifier),
+		kt:         kt.New(vrf, mapHasher, mapPubKey, logVerifier),
 		mutator:    entry.New(),
 		RetryCount: 1,
 		RetryDelay: 3 * time.Second,
@@ -240,7 +240,7 @@ func (c *Client) Update(ctx context.Context, userID, appID string, profileData [
 		return nil, fmt.Errorf("VerifyGetEntryResponse(): %v", err)
 	}
 
-	req, err := ktclient.CreateUpdateEntryRequest(&c.trusted, getResp, c.vrf, userID, appID, profileData, signers, authorizedKeys)
+	req, err := kt.CreateUpdateEntryRequest(&c.trusted, getResp, c.vrf, userID, appID, profileData, signers, authorizedKeys)
 	if err != nil {
 		return nil, fmt.Errorf("CreateUpdateEntryRequest: %v", err)
 	}
--- a/cmd/keytransparency-client/grpcc/grpc_client.go
+++ b/cmd/keytransparency-client/grpcc/grpc_client.go
@@ -26,7 +26,7 @@ import (
 	"log"
 	"time"
 
-	"github.com/google/keytransparency/core/client/kt"
+	ktclient "github.com/google/keytransparency/core/client/kt"
 	"github.com/google/keytransparency/core/crypto/signatures"
 	"github.com/google/keytransparency/core/crypto/vrf"
 	"github.com/google/keytransparency/core/crypto/vrf/p256"
@@ -79,7 +79,7 @@ var (
 type Client struct {
 	cli        spb.KeyTransparencyServiceClient
 	vrf        vrf.PublicKey
-	kt         *kt.Verifier
+	kt         *ktclient.Verifier
 	mutator    mutator.Mutator
 	RetryCount int
 	RetryDelay time.Duration
@@ -131,7 +131,7 @@ func New(cc *grpc.ClientConn,
 	return &Client{
 		cli:        spb.NewKeyTransparencyServiceClient(cc),
 		vrf:        vrf,
-		kt:         kt.New(vrf, mapHasher, mapPubKey, logVerifier),
+		kt:         ktclient.New(vrf, mapHasher, mapPubKey, logVerifier),
 		mutator:    entry.New(),
 		RetryCount: 1,
 		RetryDelay: 3 * time.Second,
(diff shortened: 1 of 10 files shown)


cmd/keytransparency-client/grpcc/grpc_client.go
```diff
@@ -26,7 +26,7 @@ import (
 	"log"
 	"time"
 
-	"github.com/google/keytransparency/core/client/kt"
+	ktclient "github.com/google/keytransparency/core/client/kt"
```
cesarghali: Please keep the same package name if it's not conflicting with other imports.

cmd/keytransparency-client/grpcc/grpc_client.go
```diff
@@ -79,7 +79,7 @@ var (
 type Client struct {
 	cli        spb.KeyTransparencyServiceClient
 	vrf        vrf.PublicKey
-	kt         *kt.Verifier
+	kt         *ktclient.Verifier
```
cesarghali: We can call this variable `ktclient` :)

core/client/kt/requests.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2016 Google Inc. All Rights Reserved.
+// Copyright 2017 Google Inc. All Rights Reserved.
```
cesarghali: Almost all our copyright headers are dated 2016.

core/client/kt/requests.go
```diff
@@ -17,22 +17,17 @@ package kt
...
-	"github.com/google/keytransparency/core/crypto/commitments"
 	"github.com/google/keytransparency/core/crypto/signatures"
 	"github.com/google/keytransparency/core/crypto/vrf"
-
-	"github.com/benlaurie/objecthash/go/objecthash"
-	"github.com/golang/protobuf/proto"
-
 	"github.com/google/keytransparency/core/mutator/entry"
-	tpb "github.com/google/keytransparency/core/proto/keytransparency_v1_types"
 	"github.com/google/trillian"
```
cesarghali: Please add an empty line before this import.

core/mutator/entry/mutation.go
```diff
@@ -0,0 +1,134 @@
...
+		entry: &tpb.Entry{
+			AuthorizedKeys: prevEntry.GetAuthorizedKeys(),
+			Previous:       hash[:],
+			Commitment:     prevEntry.GetCommitment(),
+		},
+	}, nil
+}
+
+// SetCommitment updates entry to be a commitment to data.
+func (m *Mutation) SetCommitment(data []byte) error {
```
cesarghali: Can we combine this function with `NewMutation`? `NewMutation` can take data and return a new mutation object with the commitment set.

core/mutator/entry/mutation.go
```diff
@@ -0,0 +1,134 @@
...
+func (m *Mutation) ReplaceAuthorizedKeys(pubkeys []*tpb.PublicKey) error {
+	if got, want := len(pubkeys), 1; got < want {
+		return mutator.ErrMissingKey
+	}
+	m.entry.AuthorizedKeys = pubkeys
+	return nil
+}
+
+// GenerateMutation produces the mutation.
+func (m *Mutation) GenerateMutation(signers []signatures.Signer) (*tpb.UpdateEntryRequest, error) {
```
cesarghali: Maybe call this `GenerateUpdateEntryRequest` or something like that? The names `GenerateMutation` and `NewMutations` are a bit ambiguous.

core/mutator/entry/mutation.go
```diff
@@ -0,0 +1,134 @@
+// Copyright 2017 Google Inc. All Rights Reserved.
```
cesarghali: Can you please add tests for this?

core/mutator/entry/mutation.go
```diff
@@ -0,0 +1,134 @@
...
+	userID, appID string
+	index         []byte
+	data, nonce   []byte
+
+	prevEntry *tpb.Entry
+	entry     *tpb.Entry
+}
+
+// NewMutation creates a mutation object from a previous value which can be modified.
+func NewMutation(oldValue, index []byte, userID, appID string) (*Mutation, error) {
```
cesarghali: Can we combine all the functions in one? If I understood this correctly, to get a new `UpdateEntryRequest` object the caller needs:
* Call `NewMutation`.
* Call `SetCommitment`.
* Call `GenerateMutation`.

Why not have a single function that takes whatever it needs and returns `UpdateEntryRequest`? Do we need the partial functions elsewhere?

cesarghali: "A single function" can be made of multiple but only one of them is exposed.

core/mutator/entry/mutator.go
```diff
@@ -0,0 +1,138 @@
+// Copyright 2017 Google Inc. All Rights Reserved.
```
cesarghali: If this is branched out of `core/mutator/entry/entry.go`, can you please use `git mv`? It might make it easier to review.

cmd/keytransparency-client/grpcc/grpc_client.go
gdbelvin: fixed

cmd/keytransparency-client/grpcc/grpc_client.go
gdbelvin: done

core/client/kt/requests.go
gdbelvin: I'm just updating the files that I touch

core/client/kt/requests.go
gdbelvin: done

core/mutator/entry/mutation.go
gdbelvin: We used to have a single long function - and still do in `CreateEntryUpdateRequest`. 
However, smaller functions are easier to visually inspect and test, and I also need the flexibility that the functions give for the AdminAPI.

core/mutator/entry/mutation.go
gdbelvin: What if the caller just wants to update`AuthorizedKeys`?

core/mutator/entry/mutation.go
gdbelvin: We are indeed struggling with naming consistency at the moment... 
So far we have:
- `Mutator` - which is the function that applies mutations
- `Mutation` - the mutation itself 
- `Mutations` - the API for retrieving mutations.  (perhaps this should be renamed?)

gdbelvin: Should we call this `SerializeAndSign`?

core/mutator/entry/mutator.go
gdbelvin: I split the file.. not sure how git mv is supposed to handle this?  
I could have tried git cp?

gdbelvin: Nothing in this file has changed

core/mutator/entry/mutation.go
gdbelvin: Added tests for the function with an if statement. 

All other functions are simple wrappers around other tested functions.

gdbelvin:
Updated PTAL

core/client/kt/requests.go
cesarghali: SG will do the same then.

core/mutator/entry/mutation.go
cesarghali: Sounds good. Can you please add a comment on the full sequence to generate `UpdateEntryRequest`. This will make it clear how the API should be used.

core/mutator/entry/mutation.go
cesarghali: :+1:

core/mutator/entry/mutator.go
cesarghali: If nothing has changes, then don't worry about it :)

cesarghali:
LGTM with one nit to add a comment.

core/mutator/entry/mutation.go
gdbelvin: done

PR #812, opened 2017-08-31
Use BFTKV Client to store and compare map roots

eozturk1:
* Use BFTKV to store map roots (append only) in the Sequencer and
compare them in the Monitor

* Add the Dockerfile used for running a BFTKV cluster

* Add testdata/u01 that contains a test keyring
(pubring.gpg and secring.gpg) to join the BFTKV cluster


Liamsi:
Cool. Thanks!

It looks like quite some changes which already made it to master show still up as changes/commits. To make it easier to review this PR  try rebasing on master.

eozturk1:
Extra commits that show up will be resolved once the bftkv branch is synced with master. I think this will also resolve googlebot's message.

Issue #770, opened 2017-08-18
Refactor Protos: Renaming Mutations & include signatures in Entry


Liamsi:
Remove `SignedKV` and `Mutation` and add the index and signatures to `Entry`. 

The signatures in the `Entry` are there for clients/monitors to being able to check and potentially publish who changed the Entry. The index is also needed in Entry as the signature is on the map-index, too. 

```proto
// Entry is a signed change to the map. 
// It is stored as a leaf in the map . At the same it represents a mutation.
message Entry {
  // index is the map entry key.
  bytes index = 1;
  // commitment is a cryptographic commitment to arbitrary data.
  bytes commitment = 2;
  // authorized_keys is the set of keys allowed to sign updates for this entry.
  repeated PublicKey authorized_keys = 3;  
  // signatures on key_value. Must be signed by keys from both previous and
  // current epochs. The first proves ownership of new epoch key, and the
  // second proves that the correct owner is making this change.
  map<string, sigpb.DigitallySigned> signatures = 4;
  // back-pointer to previous leaf value
  bytes previous = 3;
}

// MutationProof contains the actual mutation and the inclusion proof of the
// corresponding leaf. It is the data structure monitors receive to validate and 
// apply a mutation.
message MutationProof {
  // update contains the actual mutation information.
  // TODO: figure out if we want to keep this generic and embed proto.Any here instead. 
  Entry mutation = 1;
  // proof contains a leaf and an inclusion proof in the map of the previous
  // epoch. This is used by Storage-less monitors.
  trillian.MapLeafInclusion proof = 2;
}
```

Entry is the thing we will store in the map. `MutationProof` is the thing monitors receive.

This depends on #769 and is related to #758, #387

PR #816, opened 2017-09-27
Rename Mutation proto to MutationProof

gdbelvin:
This PR renames the `Mutation` message to `MutationProof` in an effort to maintain naming consistency.  

The update to a particular leaf is now called `Mutation` (previously `Update`). 
The proof of the leaf's former value is now called `LeafProof` (previously `Proof`).  
And the message that contains both of these is now called `MutationProof` (previously `Mutation`). 

The `MutationProof` message contains enough information for an external party to independently compute the new value for a given leaf via `Mutate(old_leaf_value, mutation)`

#770


core/proto/keytransparency_v1_types/keytransparency_v1_types.proto
```diff
@@ -96,14 +96,17 @@ message SignedKV {
...
 
-// Mutation contains the actual mutation and the inclusion proof of the
-// corresponding leaf.
-message Mutation {
-  // update contains the actual mutation information.
-  SignedKV update = 1;
-  // proof contains a leaf and an inclusion proof in the map of the previous
-  // epoch. This is used by Storage-less monitors.
-  trillian.MapLeafInclusion proof = 2;
+// MutationProof contains the information nessesary to compute the new leaf value.
```
phad: typo: necessary

core/proto/keytransparency_v1_types/keytransparency_v1_types.proto
```diff
@@ -96,14 +96,17 @@ message SignedKV {
...
+// It contains the a) old leaf value with it's inclusion proof and b) the mutation.
```
phad: the a)   ->  a) the

core/proto/keytransparency_v1_types/keytransparency_v1_types.proto
```diff
@@ -96,14 +96,17 @@ message SignedKV {
...
+// The new leaf value is computed via:
+//       Mutate(leaf_value, mutation)
+message MutationProof {
+  // mutation contains the information needed to modify the old leaf.
+  // The format of a mutation is specific to the particular Mutate function being used.
+  // TODO: figure out if we want to keep this generic and embed proto.Any here instead.
```
phad: TODO(gdbelvin):

phad: Can you think of a situation where SignedKV couldn't fully express the mutation?

core/proto/keytransparency_v1_types/keytransparency_v1_types.proto
```diff
@@ -96,14 +96,17 @@ message SignedKV {
...
+  SignedKV mutation = 1;
+  // leaf_proof contains the leaf and its inclusion proof for a particular map revision.
```
phad: No longer 'the previous epoch'?

core/proto/keytransparency_v1_types/keytransparency_v1_types.proto
gdbelvin: thanks

core/proto/keytransparency_v1_types/keytransparency_v1_types.proto
gdbelvin: done

core/proto/keytransparency_v1_types/keytransparency_v1_types.proto
gdbelvin: Some applications might not want signatures in their leaf nodes. In general, it should be up to the application to decide what the format of their leaf nodes should look like.  At the moment though, we only have KT as the application.

core/proto/keytransparency_v1_types/keytransparency_v1_types.proto
gdbelvin: I'm trying to express the idea from Ben's paper, where a new values of a leaf are computed as newvalue = F(oldvalue, mutation).  I thought the previous epoch bit was a bit confusing because we didn't have a clear definition of what epoch mutations live at. Instead, I thought it better to think of the mutation as a step function moving us forward in time.  

The monitor uses this information to verify each map revision incrementally from the beginning.

PR #818, opened 2017-09-28
Unbreak KT since change to Trillian's Map Metadata

phad:
See https://github.com/google/trillian/pull/838

Trillian has withdrawn the MapperMetadata protocol message in favor of
using google.protobuf.Any instead.  Map personalities can store any
protocol message they want in Signed Map Head metadata but must
marshal/unmarshal to Any when doing so.

This commit does the marshalling / unmarshalling in Key Transparency.

See:
- https://github.com/google/trillian/pull/838
- https://github.com/google/trillian/commit/80d7516


phad:
Since this won't build in Travis without #819 I'm going to close #818 and #819 in favor of #820 which combines them.

PR #819, opened 2017-09-28
Remove some vendored code that caused conflicts

phad:
There were two types of conflict:
- googleapis and protobuf/ptypes : Go compile time errors like:
core/internal/common.go:35: cannot use r.GetMetadata() (type *"github.com/golang/protobuf/ptypes/any".Any) as type *"github.com/google/keytransparency/vendor/github.com/golang/protobuf/ptypes/any".Any in argument to ptypes.UnmarshalAny

- prometheus and grpc-ecosystem/go-grpc-prometheus : Runtime errors along the lines of 
proto: duplicate proto type registered: grpc.gateway.runtime.StreamError
panic: proto: duplicate enum registered: io.prometheus.client.MetricType

This PR is required in order for PR #818 to actually build (Travis reports it fails with the Any import conflict above)


phad:
Since this won't build in Travis without #818 I'm going to close #818 and #819 in favor of #820 which combines them.

PR #821, opened 2017-09-29
Use the keyspb.PublicKey from Trillian

gdbelvin:
Using OneOf to distinguish key types is actually duplicating the OID's that are already present in the key DER.  Additionally, gRPC seemed to have troubles serializing proto3's one of types.  Switching to Trillian's `PublicKey`  helps us delete a bunch of code and gain some simplicity in the process. 

As a side note, KT still has a signature library that should be migrated over to the Trillian version at some point as well.


core/mutator/entry/mutation.go
```diff
@@ -15,13 +15,17 @@
...
-	"github.com/golang/protobuf/proto"
 	"github.com/google/keytransparency/core/crypto/commitments"
 	"github.com/google/keytransparency/core/crypto/signatures"
 	"github.com/google/keytransparency/core/mutator"
-	tpb "github.com/google/keytransparency/core/proto/keytransparency_v1_types"
+
+	"github.com/google/trillian/crypto/keyspb"
 	"github.com/google/trillian/crypto/sigpb"
+
+	"github.com/benlaurie/objecthash/go/objecthash"
```
phad: did you mean to move this import?
is it the result of gofmt?

core/mutator/entry/mutator.go
```diff
@@ -18,13 +18,16 @@ import (
...
-	"github.com/benlaurie/objecthash/go/objecthash"
-	"github.com/golang/glog"
-	"github.com/golang/protobuf/proto"
 	"github.com/google/keytransparency/core/crypto/signatures"
 	"github.com/google/keytransparency/core/mutator"
+
+	"github.com/google/trillian/crypto/keyspb"
 	"github.com/google/trillian/crypto/sigpb"
 
+	"github.com/benlaurie/objecthash/go/objecthash"
```
phad: Is this gofmt reordering of imports?

gdbelvin:
I'm running goimport, but manually grouping the imports so they look nice.

core/mutator/entry/mutation.go
gdbelvin: I cleaned up the import order while I was here

core/mutator/entry/mutator.go
gdbelvin: The groups are manual. goimports keeps things in groups once I create the groups

PR #823, opened 2017-10-02
Simplify mutation protos 

gdbelvin:
This PR makes incremental progress towards #770.

This PR converts
```proto
message SignedKV {
  // key_value is a serialized KeyValue.
  KeyValue key_value = 1;
  // signatures on key_value. Must be signed by keys from both previous and
  // current epochs. The first proves ownership of new epoch key, and the
  // second proves that the correct owner is making this change.
  map<string, sigpb.DigitallySigned> signatures = 2;
}
```
to
```proto
message SignedKV {
  // index is the location of this leaf in the sparse merkle tree.
  bytes index = 3;
  // value is an Entry.
  // TODO(gbelvin): migrate the elements from Entry into SignedKV.
  Entry value = 5;

  // signatures on key_value. Must be signed by keys from both previous and
  // current epochs. The first proves ownership of new epoch key, and the
  // second proves that the correct owner is making this change.
  map<string, sigpb.DigitallySigned> signatures = 2;
}
```

The next PR in this series will change the value stored in the Trillian map from `Entry` to `SignedKV`


core/proto/keytransparency_v1_types/keytransparency_v1_types.proto
```diff
@@ -64,19 +65,13 @@ message Entry {
...
-  // TODO(ismail): make proto.Any
-  // value contains the map entry value.
-  bytes value = 2;
-}
-
 // SignedKV is a signed change to a map entry.
 message SignedKV {
-  // key_value is a serialized KeyValue.
-  KeyValue key_value = 1;
+  // index is the location of this leaf in the sparse merkle tree.
```
phad: You should probably make it clear that tag 1 is not to be re-used by adding

reserved 1;  // key_value

core/proto/keytransparency_v1_types/keytransparency_v1_types.proto
```diff
@@ -41,9 +41,8 @@ message Committed {
 
 // EntryUpdate contains the user entry update(s).
 message EntryUpdate {
-  // update authorizes the change to entry.
-  // TODO(gdbelvin): rename to mutation.
-  SignedKV update = 2;
+  // mutation authorizes the change to entry.
+  SignedKV mutation = 2;
```
phad: You might want to consider using a new tag ID for this.
Although field renaming is compatible on the wire if you have any old text-serialized protos to be compatible with you'll break when the new code reads them.

cmd/keytransparency-client/grpcc/grpc_client.go
```diff
@@ -26,6 +26,7 @@ import (
 	"log"
 	"time"
 
+	"github.com/gogo/protobuf/proto"
```
phad: What's special about gogo's proto?
I'm pretty sure the github.com/golang/protobuf/proto has Equal() that should do what you want?

core/monitor/verify.go
```diff
@@ -141,12 +141,21 @@ func (m *Monitor) verifyMutations(muts []*ktpb.MutationProof, oldRoot, expectedN
...
 			errList = append(errList, ErrInvalidMutation)
 		}
 		newLeafnID := storage.NewNodeIDFromPrefixSuffix(index, storage.Suffix{}, m.mapHasher.BitLen())
+		newLeaf, err := entry.ToLeafValue(newValue)
+		if err != nil {
+			glog.Infof("Failed to serialize: %v", err)
+			errList = append(errList, err)
+		}
+
+		// BUG(gdbelvin): Proto serializations are not idempotent.
```
phad: Is this bug already present?  Or was it previously not an issue?

core/mutator/entry/mutator_test.go
```diff
@@ -93,7 +93,12 @@ func TestCheckMutation(t *testing.T) {
...
 		// Prepare mutations.
-		mutation, err := prepareMutation(tc.key, tc.newEntry, tc.previous, tc.signers)
+		m := &Mutation{
+			index: tc.key,
+			entry: tc.newEntry,
+		}
+		m.entry.Previous = tc.previous
+		mutation, err := m.sign(tc.signers)
 		if err != nil {
 			t.Fatalf("prepareMutation(%v, %v, %v)=%v", tc.key, tc.newEntry, tc.previous, err)
```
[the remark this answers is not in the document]
gdbelvin: Indeed it is.  I'll update it.

core/proto/keytransparency_v1_types/keytransparency_v1_types.proto
gdbelvin: This rename is merely for developer clarity. It maintains the same functionality. 
We don't have any text serialized protos that I know of.

cmd/keytransparency-client/grpcc/grpc_client.go
gdbelvin: goimport automatically includes this.  I'll delete it off my machine. We don't want to use it.

core/monitor/verify.go
gdbelvin: This bug was already present. I'm just flagging it here so I don't forget it when I come back to this code base.

phad: OK

phad: OK sgtm

PR #826, opened 2017-10-02
travis: add 1.9.x instead 1.8 go version

dvrkps:


gdbelvin:
Please replace `1.9.x` with `1.9` since `x` is not a valid go release number.

dvrkps:
@gdbelvin 1.9.x match all 1.9 minor releases.

PR #829, opened 2017-10-03
Store SignedKV in Trillian Map

gdbelvin:
As planned in #770.

This change puts the signatures used to authorize key changes in the
Trillian map. This allows the monitor to verify that updates were
properly authorized according to each leaf's authorized_keys policy.

### DIFF
--- a/core/mutator/entry/mutation_test.go
+++ b/core/mutator/entry/mutation_test.go
@@ -18,6 +18,10 @@ import (
 	"testing"
 
 	"github.com/google/trillian/crypto/keyspb"
+
+	"github.com/google/keytransparency/core/crypto/dev"
+	"github.com/google/keytransparency/core/crypto/signatures"
+	"github.com/google/keytransparency/core/crypto/signatures/factory"
 )
 
 func TestReplaceAuthorizedKeys(t *testing.T) {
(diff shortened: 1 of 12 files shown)


core/mutator/entry/mutation_test.go
```diff
@@ -42,3 +46,65 @@ func TestReplaceAuthorizedKeys(t *testing.T) {
...
+	for _, tc := range []struct {
+		old     []byte
+		pubKeys []*keyspb.PublicKey
+		signers []signatures.Signer
+		data    []byte
+		wantErr bool
+	}{
+		{
+			old:     nil,
+			pubKeys: []*keyspb.PublicKey{mustPublicKey(testPubKey1)},
```
phad: maybe this could be:
`pubKeys: mustPublicKeys([]string{testPubKey1}),`

feels marginally shorter?

core/mutator/entry/mutation_test.go
```diff
@@ -42,3 +46,65 @@ func TestReplaceAuthorizedKeys(t *testing.T) {
...
+		oldValue, err := FromLeafValue(tc.old)
+		if err != nil {
+			t.Errorf("FromLeafValue(%v): %v", tc.old, err)
+			continue
+		}
+		f := New()
+		if _, err := f.Mutate(oldValue, update.GetEntryUpdate().GetMutation()); err != nil {
+			t.Errorf("Mutate(%v): %v", update.GetEntryUpdate().GetMutation(), err)
+		}
+
```
phad: ding empty line here and 2 below

core/mutator/entry/mutation_test.go
```diff
@@ -42,3 +46,65 @@ func TestReplaceAuthorizedKeys(t *testing.T) {
...
+		}
+		if err := m.SetCommitment(tc.data); err != nil {
+			t.Errorf("SetCommitment(%v): %v", tc.data, err)
+			continue
+		}
+		err = m.ReplaceAuthorizedKeys(tc.pubKeys)
+		if got, want := err != nil, tc.wantErr; got != want {
+			t.Errorf("ReplaceAuthorizedKeys(%v): %v, wantErr: %v", tc.pubKeys, got, want)
+		}
+		update, err := m.SerializeAndSign(tc.signers)
```
phad: should the test continue here if err from ReplaceAuthorizedKeys was non-nil?

core/mutator/entry/mutation_test.go
```diff
@@ -42,3 +46,65 @@ func TestReplaceAuthorizedKeys(t *testing.T) {
...
+		if got, want := err != nil, tc.wantErr; got != want {
```
phad: This is the same wantErr as the one used to check the err from ReplaceAuthorizedKeys.
That seems to make it hard to test the case where ReplaceAuthorizedKeys returns nil but SerializeAndSign fails.

core/mutator/entry/mutator.go
```diff
@@ -90,14 +89,14 @@ func (*Mutator) Mutate(oldValue, update proto.Message) (proto.Message, error) {
 
 	kv := *updated
 	kv.Signatures = nil
-	if err := verifyKeys(oldEntry.GetAuthorizedKeys(),
+	if err := verifyKeys(oldEntry.GetValue().GetAuthorizedKeys(),
 		newEntry.GetAuthorizedKeys(),
 		kv,
 		updated.GetSignatures()); err != nil {
```
phad: Not your diff, but this `; err != nil{` check is really lost here.  Consider a line break just before the `)` immediately preceding it.

core/mutator/entry/mutator_test.go
```diff
@@ -30,82 +30,155 @@ func TestCheckMutation(t *testing.T) {
...
+		m, err := tc.mutation.sign(tc.signers)
 		if err != nil {
 			t.Errorf("mutation.sign(%v): %v", tc.signers, err)
 			continue
 		}
 
-		if _, got := New().Mutate(tc.oldEntry, mutation); got != tc.err {
-			t.Errorf("%d Mutate(%v, %v): %v, want %v", i, tc.oldEntry, mutation, got, tc.err)
+		if _, got := New().Mutate(tc.old, m); got != tc.err {
+			t.Errorf("%d Mutate(): %v, want %v", i, got, tc.err)
```
phad: Not really your diff, but having a `desc string` field in the test table gives more easily understood output than a test case index when something goes wrong.

core/mutator/entry/mutation_test.go
gdbelvin: Yeah, thanks!

core/mutator/entry/mutation_test.go
gdbelvin: yep.  My oversight.

core/mutator/entry/mutation_test.go
gdbelvin: this just just be `if err != nil`

core/mutator/entry/mutation_test.go
gdbelvin: done

core/mutator/entry/mutator.go
gdbelvin: Good idea.

core/mutator/entry/mutator_test.go
gdbelvin: done

gdbelvin:
Included in #830

PR #830, opened 2017-10-03
Collapse SignedKV and Entry into the same proto

### LINKED ISSUE #770: Refactor Protos: Renaming Mutations & include signatures in Entry

gdbelvin:
Closes #770

This PR completes the simplification of the proto file. 
The new `Entry` is both the leaf value in the Trillian Map and the mutation.  
`Entry` no longer is a mess of multi-level protos.


core/mutator/entry/mutator_test.go
```diff
@@ -69,7 +70,8 @@ func TestCheckMutation(t *testing.T) {
 				},
 			},
 			signers: signersFromPEMs(t, [][]byte{[]byte(testPrivKey1)}),
-		}, // Very first mutation, working case
+			desc:    "Very first mutation, working case",
```
phad: super nit:  have `desc` be the first field in each case.

PR #834, opened 2017-10-06
Revert type and service proto merge

gdbelvin:
Internally, gRPC services have separate package names than the types that they use, requiring separate import lines for services and types.  It turns out that the service / type split was entirely appropriate after all.  

This PR undoes the service / type proto merge, then cherry-pick's the commits that occurred afterwards.  Then finally, it moves all the protos to the core directory so they can be imported together.


phad:
Can you explain why this PR was needed?

PR #841, opened 2017-10-16
Readability code cleanup

gdbelvin:
More upstream Go readability code cleanup

### DIFF
--- a/core/client/multi/writer.go
+++ b/core/client/multi/writer.go
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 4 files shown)


core/client/multi/writer.go
```diff
@@ -12,32 +12,34 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package multiWriter implements a pub/sub system that meets the io.Writer interface.
-package multiWriter
+// Package multi contains utilities for multiplexing io operations.
```
phad: Good naming here!

PR #842, opened 2017-10-19
Generate protos from a common directory

gdbelvin:
This PR allows us to
- Have multiple proto files per package
- Have grpc reflection w.r.t. grpc_cli work correctly

protoc uses the path passed to protoc as the prefix for symbol
resolution. If protoc is invoked from different directories, the
resulting FileDescriptors will have different symbols, either causing
the symbols to duplicate, or be unresolvable.

The solution in this commit is to generate all the protos for this
project from the same directory like we did for Trillian.

This also has the side benefit of producing a single directory for
reverse symbol lookkup when using grpc reflection with the grpc_cli
tool.


core/proto/sequencer_v1_grpc/sequencer_v1_grpc.proto
```diff
@@ -14,9 +14,11 @@
 
 syntax = "proto3";
 
+option go_package = "github.com/google/keytransparency/core/proto/sequencer_v1_grpc";
+
 package sequencer.v1.grpc;
 
-import "github.com/google/keytransparency/core/proto/keytransparency_v1_proto/keytransparency_v1_proto.proto";
+import "proto/keytransparency_v1_proto/keytransparency_v1_proto.proto";
```
phad: sort order (should be below google/api/...)

core/proto/mutation_v1_grpc/mutation_v1_grpc.proto
```diff
@@ -14,13 +14,15 @@
...
+option go_package = "github.com/google/keytransparency/core/proto/mutation_v1_grpc";
+
 // Mutation Service
 //
 // The Key Transparency mutation server service consists of APIs to fetch
 // mutations.
 package mutation.v1.grpc;
 
-import "github.com/google/keytransparency/core/proto/keytransparency_v1_proto/keytransparency_v1_proto.proto";
+import "proto/keytransparency_v1_proto/keytransparency_v1_proto.proto";
```
phad: sort order (should be below google/api/...)

gdbelvin: done

core/proto/sequencer_v1_grpc/sequencer_v1_grpc.proto
gdbelvin: done

gdbelvin:
Adjusted the import order

PR #845, opened 2017-10-24
Admin server for adding / deleting domains

gdbelvin:
This PR creates a basic Key Transparency administration API.  

This is part of supporting multi-tenancy #534


core/adminserver/admin_server.go
```diff
@@ -0,0 +1,220 @@
...
+)
+
+type server struct {
+	storage adminstorage.Storage
+	client  trillian.TrillianAdminClient
+	keygen  keys.ProtoGenerator
+}
+
+// New returns a KeyTransparencyAdminService implementation.
+func New(
```
cesarghali: Please combine the next few lines into one, or only two lines. I don't think this is very readable. Thanks.

core/crypto/vrf/p256/p256.go
```diff
@@ -37,9 +37,10 @@ import (
 	"fmt"
 	"math/big"
 
-	"github.com/golang/protobuf/proto"
 	"github.com/google/keytransparency/core/crypto/vrf"
 	"github.com/google/trillian/crypto/keys"
+
```
cesarghali: I think we should come up with a strategy to group imports :)

core/adminserver/admin_server.go
gdbelvin: Done.  Also renamed `trillian` to `tpb` to make the line shorter.

core/crypto/vrf/p256/p256.go
gdbelvin: This follows the following import strategy:
1. golang imports
2. KT and Trillian imports
3. Third party imports
4. Underscore and renamed imports

PR #846, opened 2017-10-25
VRF: NewFromWrappedKey

gdbelvin:
Trillian supports opaque private key objects which can then be converted
back into private keys by `keys.NewSigner`.  Since VRF requires a
specific key type (ecdsa), not just a `crypto.Signer`, I've added a function to
perform the appropriate checks and type converstion.

### DIFF
--- a/core/crypto/vrf/p256/p256.go
+++ b/core/crypto/vrf/p256/p256.go
@@ -217,7 +217,8 @@ func (pk *PublicKey) ProofToHash(m, proof []byte) (index [32]byte, err error) {
 	return sha256.Sum256(vrf), nil
 }
 
-// NewFromWrappedKey creates a signer object from an encrypted private key.
+// NewFromWrappedKey creates a VRF signer object from an encrypted private key.
+// The opque private key must resolve to an `ecdsa.PrivateKey` in order to work.
 func NewFromWrappedKey(ctx context.Context, wrapped proto.Message) (vrf.PrivateKey, error) {
 	// Unwrap.
 	signer, err := keys.NewSigner(ctx, wrapped)
--- a/core/crypto/vrf/p256/p256.go
+++ b/core/crypto/vrf/p256/p256.go
@@ -218,7 +218,7 @@ func (pk *PublicKey) ProofToHash(m, proof []byte) (index [32]byte, err error) {
 }
 
 // NewFromWrappedKey creates a VRF signer object from an encrypted private key.
-// The opque private key must resolve to an `ecdsa.PrivateKey` in order to work.
+// The opaque private key must resolve to an `ecdsa.PrivateKey` in order to work.
 func NewFromWrappedKey(ctx context.Context, wrapped proto.Message) (vrf.PrivateKey, error) {
 	// Unwrap.
 	signer, err := keys.NewSigner(ctx, wrapped)
--- a/core/crypto/vrf/p256/p256.go
+++ b/core/crypto/vrf/p256/p256.go
@@ -22,6 +22,7 @@ package p256
 
 import (
 	"bytes"
+	"context"
 	"crypto"
 	"crypto/ecdsa"
 	"crypto/elliptic"
@@ -33,9 +34,12 @@ import (
 	"encoding/binary"
 	"encoding/pem"
 	"errors"
+	"fmt"
 	"math/big"
 
+	"github.com/golang/protobuf/proto"
 	"github.com/google/keytransparency/core/crypto/vrf"
+	"github.com/google/trillian/crypto/keys"
 )
 
 var (
@@ -211,6 +215,22 @@ func (pk *PublicKey) ProofToHash(m, proof []byte) (index [32]byte, err error) {
 		return nilIndex, ErrInvalidVRF
 	}
 	return sha256.Sum256(vrf), nil
+}
+
+// NewFromWrappedKey creates a signer object from an encrypted private key.
+func NewFromWrappedKey(ctx context.Context, wrapped proto.Message) (vrf.PrivateKey, error) {
+	// Unwrap.
+	signer, err := keys.NewSigner(ctx, wrapped)
+	if err != nil {
+		return nil, err
+	}
+
+	switch key := signer.(type) {
+	case *ecdsa.PrivateKey:
+		return NewVRFSigner(key)
+	default:
+		return nil, fmt.Errorf("NewSigner().type: %T, want ecdsa.PrivateKey", key)
+	}
 }
 
 // NewVRFSigner creates a signer object from a private key.
--- a/core/crypto/vrf/p256/p256_test.go
+++ b/core/crypto/vrf/p256/p256_test.go
@@ -77,9 +77,10 @@ func TestH2(t *testing.T) {
 func TestNewFromWrappedKey(t *testing.T) {
 	ctx := context.Background()
 	for _, tc := range []struct {
-		desc   string
-		spec   *keyspb.Specification
-		keygen keys.ProtoGenerator
+		desc               string
+		wantFromWrappedErr bool
+		spec               *keyspb.Specification
+		keygen             keys.ProtoGenerator
 	}{
 		{
 			desc: "DER with ECDSA spec",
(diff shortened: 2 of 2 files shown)


core/crypto/vrf/p256/p256_test.go
```diff
@@ -66,6 +74,60 @@ func TestH2(t *testing.T) {
 	}
 }
 
+func TestFromWrappedKey(t *testing.T) {
```
RJPercival: s/TestFrom/TestNewFrom/

core/crypto/vrf/p256/p256_test.go
```diff
@@ -66,6 +74,60 @@ func TestH2(t *testing.T) {
...
+		if err != nil {
+			t.Errorf("NewProtoFromSpec failed: %v", err)
+			continue
+		}
+		vrfPriv, err := NewFromWrappedKey(ctx, wrapped)
+		if err != nil {
+			t.Errorf("NewFromWrappedKey failed: %v", err)
+			continue
+		}
+		vrfPublicPB, err := der.ToPublicProto(vrfPriv.Public())
```
RJPercival: Could use `der.MarshalPublicKey(crfPriv.Public())` to avoid unnecessarily wrapping the DER in a proto.

core/crypto/vrf/p256/p256_test.go
```diff
@@ -66,6 +74,60 @@ func TestH2(t *testing.T) {
...
+				Params: &keyspb.Specification_EcdsaParams{
+					EcdsaParams: &keyspb.Specification_ECDSA{
+						Curve: keyspb.Specification_ECDSA_P256,
+					},
+				},
+			},
+			keygen: func(ctx context.Context, spec *keyspb.Specification) (proto.Message, error) {
+				return der.NewProtoFromSpec(spec)
+			},
+		},
```
RJPercival: I'm not convinced it's worth preemptively making something a tabular test when it only has one test case.

core/crypto/vrf/p256/p256_test.go
```diff
@@ -66,6 +74,60 @@ func TestH2(t *testing.T) {
...
+	} {
+		// Generate VRF key.
+		wrapped, err := tc.keygen(ctx, tc.spec)
+		if err != nil {
+			t.Errorf("NewProtoFromSpec failed: %v", err)
```
RJPercival: There's no guarantee that `tc.keygen` will call `NewProtoFromSpec`, so this error message could be misleading in the future.

RJPercival: If you're going to leave this as a tabular test, either all of the error messages should be prefixed with `tc.desc` or the body of the test should be run in a subtest (I recommend the latter).

core/crypto/vrf/p256/p256_test.go
```diff
@@ -66,6 +74,60 @@ func TestH2(t *testing.T) {
...
+			continue
+		}
+		vrfPub, err := NewVRFVerifierFromRawKey(vrfPublicPB.Der)
+		if err != nil {
+			t.Errorf("NewVRFVerifierFromRawKey failed: %v", err)
+			continue
+		}
+		// Test that the public and private components match.
+		m1 := []byte("foobar")
+		index1, proof1 := vrfPriv.Evaluate(m1)
```
RJPercival: s/index1/privIndex/
s/proof1/proof/

core/crypto/vrf/p256/p256_test.go
```diff
@@ -66,6 +74,60 @@ func TestH2(t *testing.T) {
...
+		index, err := vrfPub.ProofToHash(m1, proof1)
+		if err != nil {
+			t.Errorf("ProofToHash(): %v", err)
+			continue
+		}
+		if got, want := index, index1; got != want {
+			t.Errorf("ProofToInex(%s, %x): %x, want %x", m1, proof1, got, want)
```
RJPercival: s/ProofToInex/ProofToHash/

core/crypto/vrf/p256/p256_test.go
```diff
@@ -66,6 +74,60 @@ func TestH2(t *testing.T) {
...
+		}
+		vrfPub, err := NewVRFVerifierFromRawKey(vrfPublicPB.Der)
+		if err != nil {
+			t.Errorf("NewVRFVerifierFromRawKey failed: %v", err)
+			continue
+		}
+		// Test that the public and private components match.
+		m1 := []byte("foobar")
+		index1, proof1 := vrfPriv.Evaluate(m1)
+		index, err := vrfPub.ProofToHash(m1, proof1)
```
RJPercival: s/index/pubIndex/

core/crypto/vrf/p256/p256_test.go
```diff
@@ -66,6 +74,60 @@ func TestH2(t *testing.T) {
...
+			t.Errorf("ToPublicProto failed: %v", err)
+			continue
+		}
+		vrfPub, err := NewVRFVerifierFromRawKey(vrfPublicPB.Der)
+		if err != nil {
+			t.Errorf("NewVRFVerifierFromRawKey failed: %v", err)
+			continue
+		}
+		// Test that the public and private components match.
+		m1 := []byte("foobar")
```
RJPercival: Why "m1"?

core/crypto/vrf/p256/p256.go
```diff
@@ -213,6 +217,22 @@ func (pk *PublicKey) ProofToHash(m, proof []byte) (index [32]byte, err error) {
...
 
+// NewFromWrappedKey creates a signer object from an encrypted private key.
+func NewFromWrappedKey(ctx context.Context, wrapped proto.Message) (vrf.PrivateKey, error) {
+	// Unwrap.
+	signer, err := keys.NewSigner(ctx, wrapped)
+	if err != nil {
+		return nil, err
+	}
+
+	switch key := signer.(type) {
```
RJPercival: Instead of testing the type of the signer, I recommend testing the type of the public key instead (see [SignatureAlgorithm(crypto.PublicKey)](https://github.com/google/trillian/blob/f415147c1e56ab0b7289a5e0b16dd6531bf4cf90/crypto/signatures.go#L27)). It's possible that some key proto handlers will not return a raw key, in which case this code will fail. For example, imagine a key proto handler that communicates with a remote key store to perform signing operations. The public key is much more likely to be a raw key though.

core/crypto/vrf/p256/p256_test.go
gdbelvin: I think the Go style guides strongly prefer table driven tests, even for one test case. 
It makes it much easier to add new tests.

RJPercival: https://github.com/golang/go/wiki/TableDrivenTests uses the wording "If you ever find yourself using copy and paste when writing a test, think about [...] refactoring into a table-driven test". With only one test case, you're not eliminating any duplicated code - in fact you're adding more than necessary. If more test cases are required in the future, it's easy enough to refactor into table-driven tests then (as suggested by the wording above).

core/crypto/vrf/p256/p256.go
gdbelvin: How would you recommend this code handle a raw key? At the end of the day I need an `ecdsa.PrivateKey` in order for the rest of this library to function correctly.  Is there a conversion function I might need to call? Would a `switch` be needed?

RJPercival: Ah I see, you need access to `key.D`.  In that case, you can leave this as it is. For the kind of signer I talked about, it quite possibly won't allow access to the raw key (e.g. a HSM with non-extractable key) so you'd just have to return an error in that case.

RJPercival: It might be worth adding a little more documentation to this func to state that `wrapped` must identify a raw ECDSA private key, or something to that effect (I'm not sure what the best terms are to use).

gdbelvin: ok, that makes sense.  Yes, this particular library is doing some pretty new / not (yet) standardized crypto on the Elliptic Curve itself.  Much fancier than `crypto.Signer` :-)  I'll add a note.

core/crypto/vrf/p256/p256_test.go
gdbelvin: done

core/crypto/vrf/p256/p256_test.go
gdbelvin: do you mind if I leave the test as is?  I suspect we'll add tests in the future, and it's a bit more consistent with the other tests.  I acknowledge that it isn't 100% necessary in this case.

RJPercival: Sure, if you have an idea of other test cases that'll be added then that makes sense. I was just pushing back on using table-driven tests by default for everything. In the future, I'd recommend only making it table-driven once it needs to be, potentially saving yourself the effort of addressing comments like the ones below.

core/crypto/vrf/p256/p256_test.go
gdbelvin: done

core/crypto/vrf/p256/p256_test.go
gdbelvin: done

core/crypto/vrf/p256/p256_test.go
gdbelvin: copypasta.  Fixed.

core/crypto/vrf/p256/p256_test.go
gdbelvin: done

gdbelvin: Index isn't private, so I've labeled these `indexA` and `IndexB`

core/crypto/vrf/p256/p256_test.go
gdbelvin: done

core/crypto/vrf/p256/p256_test.go
gdbelvin: done

gdbelvin:
Made the requested changes. PTAL.

gdbelvin:
Friendly ping

core/crypto/vrf/p256/p256.go
```diff
@@ -213,6 +217,23 @@ func (pk *PublicKey) ProofToHash(m, proof []byte) (index [32]byte, err error) {
 	return sha256.Sum256(vrf), nil
 }
 
+// NewFromWrappedKey creates a VRF signer object from an encrypted private key.
+// The opque private key must resolve to an `ecdsa.PrivateKey` in order to work.
```
RJPercival: s/opque/opaque/

core/crypto/vrf/p256/p256_test.go
```diff
@@ -66,6 +74,65 @@ func TestH2(t *testing.T) {
...
+				return der.NewProtoFromSpec(spec)
+			},
+		},
+	} {
+		t.Run(tc.desc, func(t *testing.T) {
+			// Generate VRF key.
+			wrapped, err := tc.keygen(ctx, tc.spec)
+			if err != nil {
+				t.Errorf("keygen failed: %v", err)
+				return
```
RJPercival: Use `t.Fatalf()` rather than `t.Errorf(); return`

core/crypto/vrf/p256/p256_test.go
```diff
@@ -66,6 +74,65 @@ func TestH2(t *testing.T) {
...
+				Params: &keyspb.Specification_EcdsaParams{
+					EcdsaParams: &keyspb.Specification_ECDSA{
+						Curve: keyspb.Specification_ECDSA_P256,
+					},
+				},
+			},
+			keygen: func(ctx context.Context, spec *keyspb.Specification) (proto.Message, error) {
+				return der.NewProtoFromSpec(spec)
+			},
+		},
```
RJPercival: Test case for not an ECDSA key?

core/crypto/vrf/p256/p256.go
gdbelvin: done

core/crypto/vrf/p256/p256_test.go
gdbelvin: done

core/crypto/vrf/p256/p256_test.go
gdbelvin: Added

gdbelvin:
Made the requested changes. PTAL

PR #849, opened 2017-10-25
Feature: Admin Storage

gdbelvin:
This PR creates a table for storing configuration information:
- DomainID - a string designating the domain. eg. "google.com", "phonenumber", 
- MapID
- LogID
- VRF key

It also supports soft deleting domains which will be garbage collected at a later time.


impl/sql/adminstorage/admin_storage_test.go
```diff
@@ -0,0 +1,199 @@
...
+				VRF:     &keyspb.PublicKey{Der: []byte("pubkeybytes")},
+				VRFPriv: &keyspb.PrivateKey{Der: []byte("privkeybytes")},
+			},
+			setDelete:   true,
+			isDeleted:   false,
+			readDeleted: false,
+			wantReadErr: false,
+		},
+	} {
+		t.Run(tc.desc, func(t *testing.T) {
```
cesarghali: Any specific reason for using Run as opposed to running the test cases sequentially?

gdbelvin: `t.Run` is a nice way of notating each subtest with the description string without prepending each error message with the description.

cesarghali: That's pretty cool, I didn't know about that :)

PR #852, opened 2017-10-30
 Put commitments in MapLeaf.ExtraData

gdbelvin:
The previous PR put EntryUpdates in `mutationstorage`.
This commit, reads EntryUpdates from mutation storage and saves
the commitment values in MapLeaf.ExtraData.

PR #853, opened 2017-10-30
Save EntryUpdates in mutationstorage

gdbelvin:
This PR updates `mutationstorage` to hold `EntryUpdates` which holds `Commitment`s along with `Entry`s. 

In PR #852 we'd like to save commitments in `MapLeaf.ExtraData`. This removes the need to have a separate table of commitments, and it also removes the need to synchronise the garbage collection that occurs when the Trillian Map is trimmed. 

Towards that vision, the sequencer will be filling `MapLeaf.ExtraData`, and thus it needs a place to read that info from.  Since the sequencer reads from the mutations table, `mustationstorage` is a natural place to put this information. 

#534


gdbelvin:
Closed by #852

PR #856, opened 2017-11-09
remove `either`

SCKelemen:
No other option is provided, other than 404. Therefore, `either` doesn't make any sense.


SCKelemen:
I signed it!

^this should probably be `I've signed it!`

PR #861, opened 2017-11-22
Update Docker and Kubernetes scripts 

gdbelvin:
This PR updates all the necessary components to get the following commands working again:
- `docker-compose up` 
- `kubectl apply -f deploy/kubernetes/.`

The PR contains the following changes:
- Build and push images with `docker-compose` 
   - No more complicated setup scripts
- Generate kubernetes configs with `kompose convert` 
   - No more complex, hand-crafted files. 
- Initialize `default` domain with an init container. 
   - No more exposed ports from Trillian Logs and Maps
   - No more complex setup scripts and json parsing
- One line server setup command, updated README.


deploy/kubernetes/log-server-service.yaml
```diff
@@ -0,0 +1,22 @@
...
+  annotations:
+    kompose.cmd: kompose convert -v --file ../../docker-compose.yml
+    kompose.version: 1.4.0 (HEAD)
+  creationTimestamp: null
+  labels:
+    io.kompose.service: log-server
+  name: log-server
+spec:
+  ports:
+  - name: "8090"
```
phad: looks odd - why name the port with a stringified number?  can't you give it a more meaningful label?

RJPercival: This file has been generated from docker-compose.yml, which doesn't allow ports to be given names unfortunately.

RJPercival: This file has been generated from docker-compose.yml, which doesn't allow ports to be given names unfortunately.

README.md
```diff
@@ -81,25 +81,26 @@ development.
...
-```
-
-3. Run Key Transparency
-- `docker-compose up -d`
+2. Watch it Run
 - `docker-compose logs --tail=0 --follow`
-- [https://localhost:8080/v1/users/foo@bar.com?app_id=app1](https://localhost:8080/v1/users/foo@bar.com?app_id=app1)
-- [https://localhost:8080/v1/domain/info](https://localhost:8080/v1/domain/info)
+- [https://localhost:8080/v1/domains/default/users/foo@bar.com?app_id=app1](https://localhost:8080/v1/domains/default/users/foo@bar.com?app_id=app1)
+- [https://localhost:8080/v1/domains/default/info](https://localhost:8080/v1/domains/default/info)
```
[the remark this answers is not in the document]
gdbelvin: Good idea.

gdbelvin:
Updated. PTAL.

docker-compose.yml
```diff
@@ -2,9 +2,9 @@ version: "3"
 services:
   prometheus:
     depends_on:
-      - kt-server
-      - kt-signer
-    image: us.gcr.io/key-transparency/prometheus
+      - server
+      - sequencer
+    image: us.gcr.io/key-transparency/prometheus:latest
```
RJPercival: Isn't the "latest" tag used by default?

docker-compose.yml
```diff
@@ -67,44 +67,46 @@ services:
       - --batch_size=50
       - --force_master=true
       - --alsologtostderr
+    ports:
+      - "8091"
```
RJPercival: No longer need port 8092?

docker-compose.yml
gdbelvin: It is, but there's a strange issue..
Kubernetes will only pull the latest image if `latest` is specified but I had difficulty getting Kubernetes to pull a fresh image without the `:latest` tag.

docker-compose.yml
gdbelvin: All the Trillian processes expose 8090 and 8091.  To avoid port clashing I was mapping this to 8092 so an external initialization script could create a Trillian Log. Now that I'm initializing things with a docker init container, these ports no longer need to be exposed. 

In this configuration Kubernets will make the port available internally to the cluster, and docker-compose will pick a random host port to map 8091 to.

docker-compose.yml
RJPercival: Ah that could be because you aren't specifying an ImagePullPolicy (https://kubernetes.io/docs/concepts/configuration/overview/#container-images).

PR #862, opened 2017-11-22
Fix/misc

gdbelvin:
This PR contains a variety of bug fixes for issues encountered during deployment to GCE:

- SQL that is compatible with MySQL.  
  - Use INTEGER which is supported by both SQLite and MySQL. 
  - BOOLEAN, TRUE, and FALse are not supported by SQLite.
- Catch `sql.ErrNoRows` and return `codes.NotFound`
- Skip tests that cannot be run when Trillian is using an SQLite backend.
- Add URL parameter `GET /v1/domains` for `ListDomains` as a convenience / debugging endpoint.
- Create logs via a connection to the log-server and maps via a connection to the map-server.
  - Previously all `CreateTree` operations happened through a connection to the map-server.
- Periodically refresh the list of domains the sequencer operates on. 
- Add the `der/proto` underscore import to decode VRF keys.


cmd/keytransparency-sequencer/main.go
```diff
@@ -41,8 +42,9 @@ var (
 	serverDBPath = flag.String("db", "db", "Database connection string")
 
 	// Info to connect to the trillian map and log.
-	mapURL = flag.String("map-url", "", "URL of Trilian Map Server")
-	logURL = flag.String("log-url", "", "URL of Trillian Log Server for Signed Map Heads")
+	mapURL  = flag.String("map-url", "", "URL of Trilian Map Server")
```
phad: typo: trilian

cmd/keytransparency-sequencer/main.go
```diff
@@ -70,7 +72,8 @@ func main() {
 	}
 	tlog := trillian.NewTrillianLogClient(lconn)
 	tmap := trillian.NewTrillianMapClient(mconn)
-	tadmin := trillian.NewTrillianAdminClient(mconn)
+	logadmin := trillian.NewTrillianAdminClient(lconn)
```
phad: not camelCase then? (same below)

cmd/keytransparency-sequencer/server.go
```diff
@@ -18,24 +18,26 @@ import (
...
 	"github.com/golang/glog"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/reflection"
 
-	"github.com/google/keytransparency/cmd/serverutil"
-	ktpb "github.com/google/keytransparency/core/proto/keytransparency_v1_grpc"
+	gpb "github.com/google/keytransparency/core/proto/keytransparency_v1_grpc"
+	_ "github.com/google/trillian/crypto/keys/der/proto"
```
phad: I'm not sure if it's a Go convention but I think anony imports like this are sometimes separated off into their own block

cmd/keytransparency-sequencer/server.go
```diff
@@ -18,24 +18,26 @@ import (
...
 	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
 )
 
 var (
-	addr     = flag.String("metrics-addr", ":8081", "The ip:port to publish metrics on")
+	addr     = flag.String("addr", ":8080", "The ip:port to serve on")
```
phad: not sure why the default number changed?

cmd/keytransparency-sequencer/server.go
```diff
@@ -18,24 +18,26 @@ import (
...
+
 	"github.com/golang/glog"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/reflection"
 
-	"github.com/google/keytransparency/cmd/serverutil"
-	ktpb "github.com/google/keytransparency/core/proto/keytransparency_v1_grpc"
+	gpb "github.com/google/keytransparency/core/proto/keytransparency_v1_grpc"
```
phad: was serverutil completely unused?

core/adminserver/admin_server.go
```diff
@@ -82,17 +82,19 @@ var (
 )
 
 type server struct {
-	storage adminstorage.Storage
-	client  tpb.TrillianAdminClient
-	keygen  keys.ProtoGenerator
+	storage   adminstorage.Storage
+	logClient tpb.TrillianAdminClient
```
phad: yeah you've camelCased these, I think you should on the earlier ones too.

core/adminserver/admin_server.go
```diff
@@ -82,17 +82,19 @@ var (
...
 
 // New returns a KeyTransparencyAdminService implementation.
-func New(storage adminstorage.Storage, client tpb.TrillianAdminClient, keygen keys.ProtoGenerator) gpb.KeyTransparencyAdminServiceServer {
+func New(storage adminstorage.Storage, logClient, mapClient tpb.TrillianAdminClient, keygen keys.ProtoGenerator) gpb.KeyTransparencyAdminServiceServer {
 	return &server{
-		storage: storage,
-		client:  client,
-		keygen:  keygen,
+		storage:   storage,
+		logClient: logClient,
```
phad: I kinda think logAdmin and mapAdmin would be better names here, as 'logClient'/'mapClient' make me think of the primary log and map APIs and not their related admin ones.

cmd/keytransparency-server/main.go
```diff
@@ -40,6 +40,7 @@ import (
 
 	gpb "github.com/google/keytransparency/core/proto/keytransparency_v1_grpc"
 	gauth "github.com/google/keytransparency/impl/google/authentication"
+	_ "github.com/google/trillian/crypto/keys/der/proto"
```
phad: anony ordering again (unless goimports disagrees, in which case let it do it's thing)

cmd/keytransparency-sequencer/server.go
phad: However don't bother fighting goimports which has the definitive say on these things.

core/adminserver/admin_server_test.go
```diff
@@ -37,6 +38,10 @@ func vrfKeyGen(ctx context.Context, spec *keyspb.Specification) (proto.Message,
 }
 
 func TestCreateRead(t *testing.T) {
+	// We can only run the integration tests if there is a MySQL instance available.
+	if provider := testdb.Default(); !provider.IsMySQL() {
+		t.Skipf("Skipping map integration test, SQL driver is %v", provider.Driver)
```
phad: I'm beginning to think this idiom is dangerous.  It's too easy to forget to set the relevant environment variable then be persuaded all is OK when the test silently passes.
What about t.Errorf() instead?

core/keyserver/keyserver.go
```diff
@@ -358,7 +360,10 @@ func (s *Server) GetDomainInfo(ctx context.Context, in *tpb.GetDomainInfoRequest
 		return nil, grpc.Errorf(codes.InvalidArgument, "Please specify a domain_id")
 	}
 	domain, err := s.admin.Read(ctx, in.DomainId, false)
-	if err != nil {
+	if err == sql.ErrNoRows {
+		glog.Errorf("adminstorage.Read(%v): %v", in.DomainId, err)
+		return nil, status.Errorf(codes.NotFound, "Domain %v not found", in.DomainId)
+	} else if err != nil {
 		glog.Errorf("adminstorage.Read(%v): %v", in.DomainId, err)
 		return nil, grpc.Errorf(codes.Internal, "Cannot fetch domain info")
```
phad: don't you want to say anything about in.DomainId in this error message?

core/sequencer/sequencer.go
```diff
@@ -122,15 +122,23 @@ func (s *Sequencer) Initialize(ctx context.Context, logID, mapID int64) error {
...
-	for _, d := range domains {
-		go s.StartSigning(ctx, d.LogID, d.MapID, d.MinInterval, d.MaxInterval)
+func (s *Sequencer) StartSequencingAll(ctx context.Context, refresh time.Duration) error {
+	started := make(map[string]bool)
+	ticker := time.NewTicker(refresh)
+
+	for range ticker.C {
+		domains, err := s.admin.List(ctx, false)
+		if err != nil {
+			return fmt.Errorf("admin.List(): %v", err)
```
phad: stop the ticker?

core/sequencer/sequencer.go
```diff
@@ -122,15 +122,23 @@ func (s *Sequencer) Initialize(ctx context.Context, logID, mapID int64) error {
...
+		}
+		for _, d := range domains {
+			if _, ok := started[d.Domain]; !ok {
```
phad: this can be simpler (and is less fragile) due to the nil value for bool being 'false':

if !started[d.Domain] {
}

see https://play.golang.org/p/WtCZc7JnGg

impl/sql/adminstorage/admin_storage.go
```diff
@@ -38,7 +38,7 @@ CREATE TABLE IF NOT EXISTS Domains(
   VRFPrivateKey         MEDIUMBLOB NOT NULL,
   MinInterval           BIGINT NOT NULL,
   MaxInterval           BIGINT NOT NULL,
-  Deleted               BOOLEAN,
+  Deleted               INTEGER,
```
phad: what motivates this?  has Deleted become a multi-state enum rather than a boolean?

phad: Oops just read the PR desc, I see it's a SQLite compatibility thing.  Ignore.

cmd/keytransparency-sequencer/main.go
gdbelvin: done

cmd/keytransparency-sequencer/main.go
gdbelvin: done

cmd/keytransparency-sequencer/server.go
gdbelvin: Just moved it up a few lines to keep the following import pattern:
- System
- Key Transparency
- Third Party
- Renamed and underscore imports

cmd/keytransparency-sequencer/server.go
gdbelvin: There's no fighting with goimports here, which keeps imports in their own blocks.
In this change I'm trying to stick with `pb` and `gpb` for proto imports, which is a convention.

cmd/keytransparency-sequencer/server.go
gdbelvin: Previously, metrics was an http service that just served metrics. 
Now, it's an https service that is a proper gRPC and HTTPS / JSON server.   The number change is for consistency with the other Key Transparency binaries.

core/adminserver/admin_server.go
gdbelvin: Yep. done.

core/adminserver/admin_server.go
gdbelvin: Good idea. Done.

core/adminserver/admin_server_test.go
gdbelvin: I agree with you. @codingllama what do you think?   If we make it error, we should probably make it error everywhere.

core/keyserver/keyserver.go
gdbelvin: done

core/sequencer/sequencer.go
gdbelvin: Awesome 👍

core/sequencer/sequencer.go
gdbelvin: Yep.

impl/sql/adminstorage/admin_storage.go
gdbelvin: 💯

core/sequencer/sequencer.go
```diff
@@ -132,13 +132,14 @@ func (s *Sequencer) StartSequencingAll(ctx context.Context, refresh time.Duratio
...
 		for _, d := range domains {
-			if _, ok := started[d.Domain]; !ok {
+			if !started[d.Domain] {
 				glog.Infof("StartSigning domain: %v", d.Domain)
 				started[d.Domain] = true
 				go s.StartSigning(ctx, d.LogID, d.MapID, d.MinInterval, d.MaxInterval)
 			}
 		}
 	}
+	ticker.Stop()
```
phad: It won't be stopped in the early-out return "admin.List(): ... " above.
How about this:

defer func() {
  ticker.Stop()
}()

immediately after making the new ticker?

gdbelvin: good point

PR #864, opened 2017-11-24
AccountManagerService API 

gdbelvin:
This PR creates the `AccountManagerService` definition. 
This supports delegated account management by a trusted third party. 
All actions on accounts are logged in the those accounts using distinguished keys and committed to in the master append-only log. 

This PR also experiments with bringing the proto directory futher inline with the [style guide](https://cloud.google.com/apis/design/directory_structure) 
- New directory structure 
  - `core/api`
      - `v1/ketransparency_proto`
      - `subapi/`
          - `v1/subapi_proto`
- Generated code will have a `_proto` ending in the package name
- Service definitions and message definitions will live side by side


core/proto/keytransparency_v1_grpc/keytransparency_v1_admin_grpc.proto
```diff
@@ -87,3 +73,32 @@ service KeyTransparencyAdminService {
...
+// The set of account recovery keys may rotate over time with new keys being
+// added every X months, X * N months before they become the currently active signing
+// key, and another Y months before they are deactivated from the list of valid
+// verifying keys. In order to maintain the ability of this service to
+// administer accounts, they must refresh their set of authorized public keys,
+// ideally every X months.
+service AccountRecoveryService {
+
+  // GetAuthorizedKeys returns the public keys associated with this service, and the currently active key.
+  rpc GetAutorizedKeys (keytransparency.v1.proto.GetAuthorizedKeysRequest) returns (keytransparency.v1.proto.GetAuthorizedKeysResponse) {}
```
codingllama: I'm unsure of how much adherence you want to https://cloud.google.com/apis/design/, but here it goes:

At a first glance, usually Get methods work on an entity, so the expected return of `GetAuthorizedKeys` would be an `AuthorizedKey` entity, instead of `GetAuthorizedKeysResponse`.

There's an underlying question of which entities you have and how you model them. Some alternatives would be to implement this as a view on GetAccount (which returns the keys as part of the account, provided the correct input) or have AuthorizedKeys as their own collection, possibly modelled as /accounts/$account_id/authorizedkeys/$key_id/. It's hard to say just within the context of this review, so you may want to have a broader API discussion first, outside of Github.

core/proto/keytransparency_v1_grpc/keytransparency_v1_admin_grpc.proto
```diff
@@ -87,3 +73,32 @@ service KeyTransparencyAdminService {
...
+
+  // CreateAccount creates a new account and initializes it with data and authorized_keys.
+  // If the account already exists, this operation will fail.
+  rpc CreateAccount(keytransparency.v1.proto.CreateAccountRequest) returns (keytransparency.v1.proto.CreateAccountResponse) {}
```
codingllama: Ditto for `CreateAccount`, which usually returns just `Account`.

Note that the returned Account doesn't have to be a "complete" object, you can omit information that's sensitive, for example.

core/proto/keytransparency_v1_grpc/keytransparency_v1_admin_grpc.proto
```diff
@@ -87,3 +73,32 @@ service KeyTransparencyAdminService {
...
+  // SetAccountData sets the public key for an account. 
+  rpc SetAccountData(keytransparency.v1.proto.SetAccountDataRequest) returns (keytransparency.v1.proto.SetAccountDataResponse) {}
```
codingllama: Maybe call it SetAccountPublicKey, if it's a specialised method just for that.  An alternative would be just a standard, masked UpdateAccount method.  Depends on the intent.

core/proto/keytransparency_v1_proto/keytransparency_v1_admin_proto.proto
```diff
@@ -20,30 +20,74 @@ package keytransparency.v1.proto;
...
-  string app_id = 2;
+// GetAuthorizedKeysRequest is empty. 
+// TODO(gbelvin): Should we shard these keys by domain_id and app_id? 
+message GetAuthorizedKeysRequest{}
+
+// GetAuthorizedKeysResponse contains the set of public keys that correspond 
+// to the private keys the AccountRecoveryService has access to.
+message GetAuthorizedKeysResponse{
+  // AccountRecoveryKeys is a map from pub_key_id to the public key itself.
+  map<string, keyspb.PublicKey> account_recovery_keys = 1;
```
codingllama: I can't see some of this information just from the Account entity (pub_key_id, active key). It should probably be available there too. An option would be an entity that represents pub_key_id + keyspb.PublicKey, so you'd have a `repeated` field both here and at Account.

core/proto/keytransparency_v1_proto/keytransparency_v1_admin_proto.proto
```diff
@@ -20,30 +20,74 @@ package keytransparency.v1.proto;
...
-  // errors is a map from user_ids to errors, if there was an error for that user.
-  map<string, string> errors = 1;
+// CreateAccountResponse represents success / failure. 
+message CreateAccountResponse{}
+
+// SetAccountDataRequest sets the data field for the account.
+// The account must have the service's current signing key in its list of authorized_keys in order to succeed.
+message SetAccountDataRequest {
+  // account contains data which will be applied to the account.
+  Account account = 1;
```
codingllama: This is looking a lot like an Update, in which case you could simply have one. If you do, I strongly advise having an update_mask. See https://cloud.google.com/apis/design/standard_methods#update.

core/proto/keytransparency_v1_proto/keytransparency_v1_admin_proto.proto
```diff
@@ -20,30 +20,74 @@ package keytransparency.v1.proto;
...
+}
+
+// SetAccountDataResponse indicates the sucess or failure of a SetAccountDataRequest.
+message SetAccountDataResponse {}
```
codingllama: Usually what goes in comes out. In this case, possibly an Account instead of the response message.

core/proto/keytransparency_v1_proto/keytransparency_v1_admin_proto.proto
```diff
@@ -20,30 +20,74 @@ package keytransparency.v1.proto;
...
+  repeated Account accounts = 1;
+  // add_account_recovery_keys specifies whether to add this service's account_recovery_keys to the set of authorized_keys.
+  // This must be set to true if any further operations from this API are meant to succeed. 
+  // If set to false, there must be at least one key in authorized_keys.
+  bool add_account_recovery_keys = 7;
 }
 
+message BatchCreateAccountResponse {
+  // errors is a list of errors in the same order as requests.
+  repeated google.rpc.Status errors = 1;
```
codingllama: Ditto for what goes in comes out. If the create input is `repeated Account`, I would expect that to be reflected in the output.

Since you also want the status code, and parallel lists are usually to be avoided, an out here is to have a `google.rpc.Status create_status` field in Account, which is never persisted and only ever set in BatchCreate's response.

core/proto/keytransparency_v1_grpc/keytransparency_v1_admin_grpc.proto
```diff
@@ -87,3 +74,69 @@ service KeyTransparencyAdminService {
...
+    option (google.api.http) = {
+      patch: "/v1/recovery/keysets/{keyset_id}/domains/{account.domain_id}/apps/{account.app_id}/users/{account.user_id}"
+      body: "account"
+    };
+  }
+
+  // BatchCreateAccount creates multiple accounts at once.
+  rpc BatchCreateAccount(keytransparency.v1.proto.BatchCreateAccountRequest) returns (keytransparency.v1.proto.BatchCreateAccountResponse) {
+    option (google.api.http) = {
+      post: "/v1/recovery/keysets/{keyset_id}:batch"
```
gdbelvin: Should domain_id be part of the url? How should that map onto repeated accounts?

core/proto/accountmanager/v1/accountmanager.proto
```diff
@@ -0,0 +1,156 @@
+// Copyright 2017 Google Inc. All Rights Reserved.
```
cesarghali: Can we follow the same naming scheme in `core/proto`, i.e., no `v1` directory?

If we want to introduce a `v1` directory, we can do that in a separate PR. We can have `core/proto/v1` and all protos goes there.

core/proto/accountmanager/v1/accountmanager.proto
```diff
@@ -0,0 +1,156 @@
...
+// - This service should rotate its signing keys periodically.
+//   New signing keys should be published far enough in advance that they can be added
+//   by everyone before they are used to avoid service disruption. 
+// 
+// Resources:
+// - domains/apps/keyset
+//   Each combination of domain_id and app_id has it's own keyset.
+// - domains/apps/users
+//   Users can be created and updated through this endpoint.
+service AccounManagerService {
```
cesarghali: Just to confirm, will this play well when importing?

core/proto/accountmanager/v1/accountmanager.proto
```diff
@@ -0,0 +1,156 @@
...
+}
+
+// BatchCreateUserRequest creates multiple accounts all at once.
+message BatchCreateUserRequest {
+   // domain_id identifies the domain.
+   string domain_id = 1;
+   // app_id identifies the application.
+   string app_id = 2;
+  // accounts is the set of accounts to create.
+  repeated keytransparency.v1.proto.User accounts = 3;
```
cesarghali: Should we use `repeated UpdateUserRequest` instead?

core/proto/accountmanager/v1/accountmanager.proto
```diff
@@ -0,0 +1,156 @@
...
+  // add_signing_keys specifies whether to add this service's signing_keys to the set of authorized_keys.
+  // This must be set to true if any further operations from this API are meant to succeed. 
+  // If set to false, there must be at least one key in authorized_keys.
+  bool add_signing_keys = 4;
+}
+
+// BatchCreateUserRequest creates multiple accounts at once.
```
cesarghali: Please fix the comment to reflect the object name.

core/proto/accountmanager/v1/accountmanager.proto
```diff
@@ -0,0 +1,156 @@
...
+} 
+
+// BatchUpdateUserRequest creates multiple accounts all at once.
+message BatchUpdateUserRequest {
+   // domain_id identifies the domain.
+   string domain_id = 1;
+   // app_id identifies the application.
+   string app_id = 2;
+  // accounts is the set of accounts to create.
+  repeated keytransparency.v1.proto.User accounts = 3;
```
cesarghali: Same here.

core/proto/accountmanager/v1/accountmanager.proto
```diff
@@ -0,0 +1,156 @@
...
+  // update_mask specifies which fields of account to update.
+  // For example: "data" or "authorized_keys"
+  google.protobuf.FieldMask update_mask = 4;
+}
+
+// BatchUpdateUserRequest creates multiple accounts at once.
```
cesarghali: Please fix the comment.

core/proto/keytransparency_v1_proto/keytransparency_v1_proto.proto
```diff
@@ -27,11 +27,29 @@ import "crypto/keyspb/keyspb.proto";
...
+  // app_id specifies that application.
+  string app_id = 2;
+  // user_id specifies the user.
+  string user_id = 3;
+  // public_key_data is the public key material for this account.
+  bytes public_key_data = 4;
+  // authorized_keys is the set of keys allowed to sign updates for this entry.
+  repeated keyspb.PublicKey authorized_keys = 5;
+  // status is set when account is part of a batch operation.
+  google.rpc.Status status = 6;
```
cesarghali: Just out of curiosity, how is this status gonna be used? Can't we use the returned error to report the status?

core/proto/accountmanager/v1/accountmanager.proto
gdbelvin: I am doing what I can to follow the directory structure outlined in the [style guide](https://cloud.google.com/apis/design/directory_structure). We previously didn't think we could follow the style guide in the open source, but I think I've been able to solve that in this PR.  If this pattern is successful, I will move the other protos to follow suit.

core/proto/accountmanager/v1/accountmanager.proto
gdbelvin: I don't think so.  `UpdateUserRequest` is an API level message. 
In terms of the API working on entities, it makes sense to send a list of the things you want to create in the batch create request, not a list of individual requests.

core/proto/accountmanager/v1/accountmanager.proto
gdbelvin: done

core/proto/accountmanager/v1/accountmanager.proto
gdbelvin: done

core/proto/accountmanager/v1/accountmanager.proto
gdbelvin: In what way are you thinking it could go wrong?  
It's certainly my intention for it to work well when importing :-)

core/proto/keytransparency_v1_proto/keytransparency_v1_proto.proto
gdbelvin: Typically, when you create or modify an object, you return the same object as a result.  This works great when creating or modifying a single `User`, but gets a little muddy when we're talking about a batch of users. How should we associate the particular status with the particular user?  Here we've just put it inside the User object. It will only be filled in the response to a batch message.

gdbelvin:
Updated PTAL

core/proto/accountmanager/v1/accountmanager.proto
cesarghali: SG.

core/proto/keytransparency_v1_proto/keytransparency_v1_proto.proto
cesarghali: SG, thanks for the explanation!

core/managementserver/server.go
```diff
@@ -0,0 +1,59 @@
...
+
+// New creates a new managementserver
+func New() *Server {
+	return &Server{}
+}
+
+// GetKeySet returns a list of public keys (a keyset) that corresponds to the signing keys
+// this service has for a given domain and app.
+func (s *Server) GetKeySet(context.Context, *pb.GetKeySetRequest) (*keypb.KeySet, error) {
+	panic("unimplemented")
```
codingllama: Do `return status.Errorf(codes.Unimplemented, "unimplemented")` instead?

https://godoc.org/google.golang.org/grpc/status#Errorf
https://godoc.org/google.golang.org/grpc/codes#Code

core/api/usermanager/v1/usermanager_proto/usermanager.proto
```diff
@@ -0,0 +1,156 @@
...
+   string app_id = 2;
+  // users is the set of users to create.
+  repeated keytransparency.v1.proto.User users = 3;
+  // add_signing_keys specifies whether to add this service's signing_keys to the set of authorized_keys.
+  // This must be set to true if any further operations from this API are meant to succeed.
+  // If set to false, there must be at least one key in authorized_keys.
+  bool add_signing_keys = 4;
+}
+
+// BatchCreateUserRequest creates multiple users at once.
```
codingllama: nit: BatchCreateUserResponse, not request. Possibly needs to be rephrased too.

core/api/usermanager/v1/usermanager_proto/usermanager.proto
```diff
@@ -0,0 +1,156 @@
...
+message BatchCreateUserResponse {
+  // users returns the list of created users.
+  repeated keytransparency.v1.proto.User users = 2;
+}
+
+// BatchUpdateUserRequest creates multiple users all at once.
```
codingllama: nit: s/creates/updates/

core/api/usermanager/v1/usermanager_proto/usermanager.proto
```diff
@@ -0,0 +1,156 @@
...
+message BatchUpdateUserRequest {
+   // domain_id identifies the domain.
+   string domain_id = 1;
+   // app_id identifies the application.
+   string app_id = 2;
+  // users is the set of users to create.
+  repeated keytransparency.v1.proto.User users = 3;
+  // update_mask specifies which fields of user to update.
+  // For example: "data" or "authorized_keys"
+  google.protobuf.FieldMask update_mask = 4;
```
codingllama: I don't recall seeing a BatchUpdate before, so there isn't much I can draw from, but since all updates have to use the same mask, do you think this method will end up being useful enough to be here? Are you foreseeing some sort of "batch backfill" that touches the same fields throughout large numbers of users?

Another question would be: does the BatchUpdate require a List method in order to be useful, or would the client be aware of the user space without it?

core/api/usermanager/v1/usermanager_proto/usermanager.proto
```diff
@@ -0,0 +1,156 @@
...
+}
+
+
+// The UserManagerService holds a set of signing keys which can be used to
+// sign user updates.
+// - This service must be registered with the server in order to submit user
+// create or update requests.
+// - The update requests this service generates will only be valid for users
+// that have added this service's signing keys to their list of authorized_keys.
+// - This service should carefully controll and log access to it's APIs.
```
codingllama: nit: typo on "controll"

core/api/usermanager/v1/usermanager_proto/usermanager.proto
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  // UpdateUserData sets the public key for an user.
+  rpc UpdateUser(UpdateUserRequest) returns (keytransparency.v1.proto.User) {
+    option (google.api.http) = {
+      patch: "/usermanager/v1/domains/{user.domain_id}/apps/{user.app_id}/users/{user.user_id}"
+      body: "user"
+    };
+  }
+
```
codingllama: Do you need to include Get / ListUsers here, or are those covered by another API? If the latter, it may be worth mentioned that somewhere in the docs (ie, "to read users use $otherAPI.$someMethod" or something like it.)

core/managementserver/server.go
```diff
@@ -0,0 +1,59 @@
...
+	pb "github.com/google/keytransparency/core/api/usermanager/v1/usermanager_proto"
+	keypb "github.com/google/keytransparency/core/proto/keymaster_proto"
+	ktpb "github.com/google/keytransparency/core/proto/keytransparency_v1_proto"
+)
+
+// Server implements pb.UserManagerServiceServer
+type Server struct{}
+
+// New creates a new managementserver
+func New() *Server {
```
codingllama: Do you need the indirection here, or could users just do `s := &Server{}` ?

core/api/usermanager/v1/usermanager_proto/usermanager.proto
```diff
@@ -0,0 +1,156 @@
...
+//   Each combination of domain_id and app_id has it's own keyset.
+// - domains/apps/users
+//   Users can be created and updated through this endpoint.
+service UserManagerService {
+
+  // GetKeySet returns a list of public keys (a keyset) that corresponds to the signing keys
+  // this service has for a given domain and app.
+  rpc GetKeySet (GetKeySetRequest) returns (keymaster.proto.KeySet) {
+    option (google.api.http) = {
+      get: "/usermanager/v1/domains/{domain_id}/apps/{app_id}/keyset"
```
codingllama: Since you have an implied 1:1 relationship between KeySets and Apps would it make sense to return them as part of the App entity? That would mean you have a "GetApp" method instead of "GetKeySet".

"No" is perfectly fine answer here, specially if there are motives for the design that I might be missing.

core/api/usermanager/v1/usermanager_proto/usermanager.proto
```diff
@@ -0,0 +1,156 @@
...
+// create or update requests.
+// - The update requests this service generates will only be valid for users
+// that have added this service's signing keys to their list of authorized_keys.
+// - This service should carefully controll and log access to it's APIs.
+// - This service should rotate its signing keys periodically.
+//   New signing keys should be published far enough in advance that they can be added
+//   by everyone before they are used to avoid service disruption.
+//
+// Resources:
+// - domains/apps/keyset
```
codingllama: Is it worth mentioning how Domains, Apps and KeySets are created? (IIRC, you have a pre-defined number of them in config files, right?)

core/api/usermanager/v1/usermanager_proto/usermanager.proto
gdbelvin: done

core/api/usermanager/v1/usermanager_proto/usermanager.proto
gdbelvin: done

core/api/usermanager/v1/usermanager_proto/usermanager.proto
gdbelvin: I foresee wanting to update all the authorized_keys in a consistent manner such as
- Remove Key X
- Add Key Y

But that doesn't work very well with the update_mask on a list does it?

Your list question is very astute.  The Key Transparency API does not have a list option on purpose in order to not help spammers.  This API, however, could keep track of the IDs that it created on it's own.

core/api/usermanager/v1/usermanager_proto/usermanager.proto
gdbelvin: thanks

core/api/usermanager/v1/usermanager_proto/usermanager.proto
gdbelvin: I'll add a line about it.  Thanks for thinking of it.

core/api/usermanager/v1/usermanager_proto/usermanager.proto
gdbelvin: Interesting idea.  
If the App type defined in usermanager is allowed to be different than the main API, then yes, we could do that.  It is possible, however, that we may want to make the keyset end point more broadly available than the other info we might want to put in GetApp

core/managementserver/server.go
gdbelvin: This function will be expanding to take arguments in the future :-)
I quite like calling `New` to create objects. Is that not canonical?

core/managementserver/server.go
gdbelvin: good idea.

core/managementserver/server.go
codingllama: A New() func is OK, but since Server is exported, a trivial New might get you worse readability than a Server literal (because of field labels vs method params).

This is just my 2c though.

PR #870, opened 2017-11-29
Mutation improvements

gdbelvin:
In the continued quest for easier account account logic...

This PR gives `entry.Mutation` the following new responsibilities:
- Fully setting up the `UpdateRequest` proto so callers don't have to manually fix it up.
- Verifying whether an update was applied correctly - `Check()`

As a result, the client can now operate on units of `entry.Mutation` rather than raw request protos. 
This allows us to fix an outstanding TODO that will resolve two racing clients gracefully by recreating requests from scratch on each `Retry` rather than blindly resubmitting the same request. 

Other nits:
- All functions that accept `userID` and `appID` should do so in `domainID`, `appID`, `userID` order.


core/mutator/entry/mutation.go
```diff
@@ -148,3 +153,19 @@ func (m *Mutation) sign(signers []signatures.Signer) (*pb.Entry, error) {
 	m.entry.Signatures = sigs
 	return m.entry, nil
 }
+
+// Check verifies that an update was successfully applied.
+// Returns nil if newLeaf is equal to the entry in this mutation.
+func (m *Mutation) Check(newLeaf []byte) error {
+	// XXX Mutations are no longer stable serialized byte slices, so we need to use
```
phad: XXX - is this a todo?

core/mutator/entry/mutation.go
```diff
@@ -148,3 +153,19 @@ func (m *Mutation) sign(signers []signatures.Signer) (*pb.Entry, error) {
...
+	// an equality operation on the proto itself.
+	leafValue, err := FromLeafValue(newLeaf)
+	if err != nil {
+		return fmt.Errorf("failed to decode current entry: %v", err)
+	}
+
+	if got, want := leafValue, m.entry; !proto.Equal(got, want) {
+		return errors.New("newLeaf does not match expected value")
```
phad: I've seen advice that there's no special reason to use errors.New just because you have no varargs items for a fmt.Errorf() call.  But I think you have pointed me to https://en.wikipedia.org/wiki/Uncontrolled_format_string on this topic, am I recalling correctly?

core/mutator/entry/mutation.go
```diff
@@ -148,3 +153,19 @@ func (m *Mutation) sign(signers []signatures.Signer) (*pb.Entry, error) {
 	m.entry.Signatures = sigs
 	return m.entry, nil
 }
+
+// Check verifies that an update was successfully applied.
+// Returns nil if newLeaf is equal to the entry in this mutation.
```
phad: It seems to me that the function is answering a yes/no question, but that errors can occur.
With this func signature the caller can't distinguish between 'no' and 'oops'.

A (bool, error) return would make it unambiguous but is ugly.

Better would be an error constant that the caller can check for.
You could just move the errors.New("message") to an constant in this package and return that.

Does Check need to be exported? I see it being used in this same file.  If so then the error constant should be too.

core/mutator/entry/mutation.go
```diff
@@ -59,9 +58,13 @@ func NewMutation(index []byte, domainID, appID, userID string) *Mutation {
 	}
 }
 
+// CopyPrevious indicates whether to overwrite all values in the current entry with
+// the previous entry's values.
+type CopyPrevious bool
```
phad: I don't think there's a readability gain from having a typedef'd bool.

In SetPrevious itself, you write 'if copyPrevious { ... }'
You'd write that whether or not you define the type here.

If you want the extra type safety / documentation then making this more of an enum so that values are named as well would be more helpful.

```
type CopyPrevious int
const (
  copyPrevious = iota
  dontCopyPrevious
)
```

maybe?

core/mutator/entry/mutation.go
```diff
@@ -156,16 +167,13 @@ func (m *Mutation) sign(signers []signatures.Signer) (*pb.Entry, error) {
 
 // Check verifies that an update was successfully applied.
 // Returns nil if newLeaf is equal to the entry in this mutation.
-func (m *Mutation) Check(newLeaf []byte) error {
+func (m *Mutation) Check(newLeaf []byte) (bool, error) {
```
phad: oops, I didn't see you've already thought about this since the previous commit.
Do consider the error constant alternative though.

core/client/grpcc/grpc_client.go
```diff
@@ -230,9 +228,9 @@ func (c *Client) ListHistory(ctx context.Context, userID, appID string, start, e
 
 // Update creates an UpdateEntryRequest for a user, attempt to submit it multiple
 // times depending on RetryCount.
-func (c *Client) Update(ctx context.Context, userID, appID string, profileData []byte,
+func (c *Client) Update(ctx context.Context, appID, userID string, profileData []byte,
```
phad: hope you caught all the call-sites of this change!

core/client/grpcc/grpc_client.go
```diff
@@ -244,58 +242,54 @@ func (c *Client) Update(ctx context.Context, userID, appID string, profileData [
...
-		err = c.Retry(ctx, req, opts...)
+		err = c.Retry(ctx, m, signers, opts...)
 	}
-	return req, err
+	return m, err
 }
 
-// Retry will take a pre-fabricated request and send it again.
-func (c *Client) Retry(ctx context.Context, req *pb.UpdateEntryRequest, opts ...grpc.CallOption) error {
+// Retry will take a mutation and send it again.
```
phad: nit: "Retry takes a mutation and..."

core/client/kt/requests.go
```diff
@@ -18,51 +18,45 @@ package kt
...
+	return index[:], nil
+}
 
-	oldLeaf := getResp.GetLeafProof().GetLeaf().GetLeafValue()
-	mutation := entry.NewMutation(index[:], domainID, appID, userID)
-	if err := mutation.SetPrevious(oldLeaf); err != nil {
-		return nil, fmt.Errorf("Error unmarshaling Entry from leaf proof: %v", err)
+// NewMutation creates a Mutation given the userID, desired state, and previous entry.
+func (v *Verifier) NewMutation(
+	domainID, appID, userID string,
```
phad: Having no stronger typing and a list of 3 different fooIDs makes me a bit nervous...  it'd be easy to transpose args by mistake in a call site and get some badly broken behaviour as a result...

integration/client_test.go
```diff
@@ -100,38 +100,41 @@ func TestEmptyGetAndUpdate(t *testing.T) {
...
 		authorizedKeys []*keyspb.PublicKey
 	}{
-		{false, false, context.Background(), "noalice", signers1, authorizedKeys1},                // Empty
-		{false, true, GetNewOutgoingContextWithFakeAuth("bob"), "bob", signers1, authorizedKeys1}, // Insert
-		{false, false, context.Background(), "nocarol", signers1, authorizedKeys1},                // Empty
-		{true, false, context.Background(), "bob", signers1, authorizedKeys1},                     // Not Empty
-		{true, true, GetNewOutgoingContextWithFakeAuth("bob"), "bob", signers1, authorizedKeys1},  // Update
-		{true, true, GetNewOutgoingContextWithFakeAuth("bob"), "bob", signers2, authorizedKeys2},  // Update, changing keys
-		{true, true, GetNewOutgoingContextWithFakeAuth("bob"), "bob", signers3, authorizedKeys3},  // Update, using new keys
+		{"Empty", false, false, context.Background(), "noalice", signers1, authorizedKeys1},
```
phad: This is pushing the limit of what's readable and maintainable in a test table full of one-liners (and no labels!)

core/client/grpcc/grpc_client.go
gdbelvin: I think I did... the unit tests helped me track down places where I had missed it.

core/mutator/entry/mutation.go
gdbelvin: Should I use a plain `bool` then?

core/client/kt/requests.go
gdbelvin: Yeah... Consistency between APIs should help, but I'm not sure what else to do.

core/mutator/entry/mutation.go
gdbelvin: `(bool, error)` is what I ended up doing.   
I thought about defining a public error, but then considered that it would force calling sites into a just as awkward error parsing. 

```
if err := m.Check(); err == ErrNotEqual {
} else if err != nil {
return err
}
```

core/mutator/entry/mutation.go
gdbelvin: It's more of a warning and a TODO that's related to our discussion of ObjectHash. 
Storing objects is really nice because I get to avoid lots of `UnMarshals`, but then I need to do object comparison in a reliable way...

core/mutator/entry/mutation.go
gdbelvin: That's right.  A secure way of using `fmt.Errorf` in this case would look like:
```go
 fmt.Errorf("%v", "A string with no variables")
```

integration/client_test.go
gdbelvin: Added labels :-)

gdbelvin:
Updated. PTAL

core/mutator/entry/mutation.go
phad: OK, I don't mind much either way, good you considered both options.

core/client/kt/requests.go
phad: Well, the strong typing option is available:

type DomainID string
type AppID string
type UserID string

Do you think that would cause a lot of churn and annoyance?

core/client/grpcc/grpc_client.go
gdbelvin: This comment slipped by.  Fixed now.

core/client/kt/requests.go
gdbelvin: I kind of think it would cause more churn and type casting than really needed. 
If a call site gets it wrong all subsequent crypto proofs will fail, so there's an easy way to tell for sure if you got it right.

PR #871, opened 2017-12-05
bump grpc gateway to v1.3

gdbelvin:
This is in preparation for using the `go_package` annotation in proto files which currently causes the grpc-gateway proto generator to crash. Note that this update might need to be coordinated with trillian.


impl/google/authentication/google_test.go
```diff
@@ -51,6 +51,8 @@ func TestGoogleValidateCreds(t *testing.T) {
 		t.Fatalf("Failed to create GoogleAuth: %v", err)
 	}
 
+	mux := runtime.NewServeMux()
```
phad: Why does this commit belong in this PR?  It looks unrelated (but I'm not familiar with runtime.AnnotateContext)

gdbelvin: This commit is in this PR because the function signature for `runtime.AnnotateContext` changed between grpc-gateway v1.2 and v1.3.

phad: OK - I didn't look too closely. :-)

PR #874, opened 2017-12-07
Switch from GetLeaves to GetLeavesByRevision.

phad:
Do this where a specific revision is required.
When retrieving the latest revision, GetLeaves is still used.
In a later trillian API change the special '-1' revision will
no longer be used.

### DIFF
--- a/core/keyserver/keyserver.go
+++ b/core/keyserver/keyserver.go
@@ -116,9 +116,10 @@ func (s *Server) getEntry(ctx context.Context, domainID, userID, appID string, f
 		glog.Errorf("tlog.GetLatestSignedLogRoot(%v): %v", domain.LogID, err)
 		return nil, grpc.Errorf(codes.Internal, "Cannot fetch SignedLogRoot")
 	}
-	// Use the log as the athoritative source of the latest revision.
+	// Use the log as the authoritative source of the latest revision.
 	if revision < 0 {
 		// The maximum index in the log is one minus the number of items in the log.
+		// TODO(phad): could GetTreeSize() return 0?
 		revision = logRoot.GetSignedLogRoot().GetTreeSize() - 1
 	}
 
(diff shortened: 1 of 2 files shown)


core/keyserver/keyserver.go
```diff
@@ -116,9 +116,10 @@ func (s *Server) getEntry(ctx context.Context, domainID, userID, appID string, f
 		glog.Errorf("tlog.GetLatestSignedLogRoot(%v): %v", domain.LogID, err)
 		return nil, grpc.Errorf(codes.Internal, "Cannot fetch SignedLogRoot")
 	}
-	// Use the log as the athoritative source of the latest revision.
+	// Use the log as the authoritative source of the latest revision.
 	if revision < 0 {
 		// The maximum index in the log is one minus the number of items in the log.
+		// TODO(phad): could GetTreeSize() return 0?
```
gdbelvin: It _shouldn't_ because we start the log with at least one item in it.

core/keyserver/keyserver.go
```diff
@@ -116,9 +116,10 @@ func (s *Server) getEntry(ctx context.Context, domainID, userID, appID string, f
 		glog.Errorf("tlog.GetLatestSignedLogRoot(%v): %v", domain.LogID, err)
 		return nil, grpc.Errorf(codes.Internal, "Cannot fetch SignedLogRoot")
 	}
-	// Use the log as the athoritative source of the latest revision.
+	// Use the log as the authoritative source of the latest revision.
```
gdbelvin: Please add a TODO to add a similar GetEntryByRevision as well.  You can assign it to me.

core/keyserver/keyserver.go
phad: OK, thanks.  If this did occur the math on revision will cause GetLeavesByRevision to fail with an error.

PR #875, opened 2017-12-08
Fix core/sequencer/sequencer.go; remove obsolete 'Revision'.

phad:


gdbelvin:
Please run gofmt, and I think it will pass the presubmit checks.

PR #876, opened 2017-12-08
Use specific interface 

gdbelvin:
We were using a more general interface (grpc.Client) to the Key Transparency client than needed.


phad:
Travis says:
```
integration/testutil.go:172:21: cannot use cc (type *grpc.ClientConn) as type keytransparency_v1_grpc.KeyTransparencyServiceClient in argument to grpcc.New:
	*grpc.ClientConn does not implement keytransparency_v1_grpc.KeyTransparencyServiceClient (missing GetDomainInfo method)
```

Looks like integration/testutil.go needs an update too.

PR #877, opened 2017-12-08
Move User proto type to type directory

gdbelvin:
This conforms with the API style guide

This commit also scopes keytransparency with the google.keytransparency prefix, also per the style guide.


core/api/usermanager/v1/usermanager_proto/usermanager.proto
```diff
@@ -16,11 +16,11 @@ syntax = "proto3";
 
 option go_package = "github.com/google/keytransparency/core/api/usermanager/v1/usermanager_proto";
 
-package keytransparency.usermanager.v1;
+package google.keytransparency.usermanager.v1;
 
 import "google/api/annotations.proto";
 import "google/protobuf/field_mask.proto";
-import "proto/keytransparency_v1_proto/keytransparency_v1_proto.proto";
+import "api/type/type_proto/type.proto";
```
phad: Wrong import ordering.  goimports should fix that?

core/api/type/type_proto/type.proto
```diff
@@ -0,0 +1,39 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+option go_package = "github.com/google/keytransparency/core/api/type/type_proto";
```
phad: I wonder if 'types' would be a more pragmatic choice than 'type' given that 'type' is a Go reserved word.
I know the import will be type_proto but it still might be better to avoid potential confusion?

core/api/usermanager/v1/usermanager_proto/usermanager.proto
```diff
@@ -36,7 +36,7 @@ message GetKeySetRequest {
 // It is the responsibility of authorized callers to verify that domain_id/app_id/user_id is correct.
 message CreateUserRequest {
   // user is the user to create.
-  keytransparency.v1.proto.User user = 2;
+  type.User user = 2;
```
phad: Then this would be `types.User user = 2;` which I think reads better in any case. wdyt?

core/api/type/type_proto/type.proto
gdbelvin: I think this would produce a slightly odd URL for proto type definitions:
 `keytransparency/types/typename`

core/api/usermanager/v1/usermanager_proto/usermanager.proto
gdbelvin: What is the correct ordering here? 
goimports doesn't work on proto files ... :-)

core/api/usermanager/v1/usermanager_proto/usermanager.proto
gdbelvin: see above

core/api/usermanager/v1/usermanager_proto/usermanager.proto
phad: If only!  There's not enough Go.

import "api/type/..."
import "google/..."

core/api/type/type_proto/type.proto
phad: OK.

core/api/usermanager/v1/usermanager_proto/usermanager.proto
phad: OK

PR #880, opened 2017-12-11
Move the monitor api to core/api/monitor

gdbelvin:
Merge monitor_v1_proto with monitor_v1_grpc and move it to the api directory. 

Merge #879 first.

PR #881, opened 2017-12-12
Decouple api/monitor and api/v1

gdbelvin:
Merge #880 first

This PR returns extra information inside of rpc.Status, rather than creating an explicit dependency on the main keytransparency types, somewhat simplifying the relationships between the APIs.


core/api/monitor/v1/monitor_proto/monitor.proto
```diff
@@ -58,12 +58,7 @@ message GetMonitoringResponse {
 
   // errors contains a list of error string representing the verification checks
```
phad: nit: "of error strings"

core/monitor/verify.go
```diff
@@ -55,6 +57,15 @@ var (
 	ErrNotMatchingMapRoot = errors.New("recreated root does not match")
 )
 
+func appendErr(errList []error, s *status.Status, err error) []error {
```
phad: I was idly wondering whether:
 * `errList` might be a type:   `type errList []error`
 * `append` might be a method for an `errList` receiver:  func (el *errList) append(e error) {}`
So the calls in the code further down would be:
```
errs := &errList{}
errs.append(s, err)
...
```
As in this playground: https://play.golang.org/p/Lus2oqao8X

But actuallly I think it's uglier.  So nvm :)

core/monitorserver/server.go
```diff
@@ -91,11 +92,12 @@ func (s *Server) getResponseByRevision(epoch int64) (*pb.GetMonitoringResponse,
 	}
 
 	for _, err := range r.Errors {
-		resp.Errors = append(resp.Errors, err.Error())
-	}
-	if len(r.Errors) > 0 {
-		// data to replay the verification steps:
-		resp.ErrorData = r.Response
+		s, ok := status.FromError(err)
```
phad: This feels like it can be restructured to remove some of the indentation.  How about:
```
if s, ok := status.FromError(err); ok {
  resp.Errors = append(resp.Errors, s.Proto())
  continue
}
resp.Errors = append(resp.Errors, status.New(codes.Unknown, err.Error()).Proto())
```

core/api/monitor/v1/monitor_proto/monitor.proto
gdbelvin: thanks

core/monitorserver/server.go
gdbelvin: I like how it shortens it.  Question though..   typically we want to make the indented path the error path, but this makes it the success path.  Is that worth worrying over?

core/monitor/verify.go
gdbelvin: I actually like it.

core/monitor/verify.go
```diff
@@ -153,31 +149,28 @@ func (m *Monitor) verifyMutations(muts []*pb.MutationProof, oldRoot, expectedNew
...
+			errs.append(status.Newf(codes.DataLoss, "failed to serialize: %v", err).WithDetails(newValue))
 		}
 
 		// BUG(gdbelvin): Proto serializations are not idempotent.
 		// - Upgrade the hasher to use ObjectHash.
 		// - Use deep compare between the tree and the computed value.
 		newLeafHash, err := m.mapHasher.HashLeaf(mapID, index, newLeaf)
 		if err != nil {
-			errList = append(errList, err)
+			errs = append(errs, err)
```
phad: Maybe define in addition a single-arg `func (el errList) append(err)` so this case looks more like the ones above?

core/monitorserver/server.go
```diff
@@ -86,19 +87,19 @@ func (s *Server) getResponseByRevision(epoch int64) (*pb.State, error) {
 		return nil, grpc.Errorf(codes.NotFound, "Could not find monitoring response for epoch %d", epoch)
 	}
 
-	resp := &pb.State{
-		Smr:                r.Smr,
-		SeenTimestampNanos: r.Seen.UnixNano(),
-	}
-
+	// Convert errors into rpc.Status
+	errs := make([]*statuspb.Status, 0, len(r.Errors))
```
phad: (optional, since this is a different package, and exporting `errList` makes you extra work)
This block which maps `[]error -> []*statuspb.Status` would be nice as a func on `errList`

core/monitorserver/server.go
phad: Hmm, that's a good point.  Possibly not worth the change in approach just to save a few lines, since as you point out most Go readers expect the indented bit to be early-out for error handling.

phad:
```
s, ok := status.FromError(err)
if !ok {
  resp.Errors = append(resp.Errors, status.New(codes.Unknown, err.Error()).Proto())
  continue
}
resp.Errors = append(resp.Errors, s.Proto())
```

This reverses it.

PR #882, opened 2017-12-12
Split GetMutations API into GetEpoch and ListMutations 

gdbelvin:
Merge #887 and #888 First 

This PR splits the `GetMutations` endpoint into `GetEpoch` and `ListMutations`. 

The previous design combined epoch objects, and their various proofs, with mutations and their proofs, making it difficult to determine sane pagination points, and reusable code blocks. 

This PR separates the two entities into separate API end points, per API style. 
- `GetEpoch` can now be cleanly iterated over
   - It also has a small streaming RPC for faster notification of new epochs.
   - It could also make a natural gossip reply endpoint in the future.
- `ListMutations` can now be cleanly iterated over without duplicating epoch information
   - It can now handle very large diffs between epochs.  
   - And it gets it's own streaming RPC too. 

As a result of these changes, the monitoring code processing diffs between epochs can be broken up and simplified into:
1. Waiting for new epochs to arrive
2. Grouping new epochs into pairs
3. Fetching mutations
4. Verifying the difference between each epoch pair.


core/api/monitor/v1/monitor_proto/monitor.proto
```diff
@@ -0,0 +1,102 @@
...
+import "api/v1/keytransparency_proto/keytransparency.proto";
+import "trillian.proto";
+
+// GetMonitoringRequest contains the input parameters of the GetMonitoring APIs.
+message GetMonitoringRequest {
+  // kt_url is the URL of the keytransparency server for which the monitoring
+  // result will be returned.
+  string kt_url = 2;
+
+  // domain_id identifies the merkel tree being monitored.
```
phad: typo: Merkle

core/api/monitor/v1/monitor_proto/monitor.proto
```diff
@@ -0,0 +1,102 @@
...
+// validated signed map roots.
+//
+// - Signed Map Roots can be collected using the GetSignedMapRoot APIs.
+// - Monitor resources are named:
+//   - /monitor/v1/{kt-url}/domains/{domain_id}/revisions/{epoch}
+//   - /monitor/v1/{kt-url}/domains/{domain_id}/revision:latest
+//
+service MonitorService {
+  // GetSignedMapRoot returns the latest valid signed map root the monitor
+  // observed. Additionally, the response contains additional data necessary to
```
phad: Additionally ... additional
Maybe 'Additionally, the response contains extra data necessary...'

core/api/monitor/v1/monitor_proto/monitor.proto
```diff
@@ -0,0 +1,102 @@
...
+  //
+  // Returns the signed map root for the latest epoch the monitor observed. If
+  // the monitor could not reconstruct the map root given the set of mutations
+  // from the previous to the current epoch it won't sign the map root and
+  // additional data will be provided to reproduce the failure.
+  rpc GetSignedMapRoot(GetMonitoringRequest) returns (GetMonitoringResponse) {
+    option (google.api.http) = { get: "/monitor/v1/{kt_url}/domains/{domain_id}/revisions:latest" };
+  }
+
+  // GetSignedMapRootByRevision works similar to GetSignedMapRoot but returns
```
phad: s/similar/similarly/

core/api/v1/keytransparency_proto/admin.proto
```diff
@@ -113,15 +103,15 @@ service KeyTransparencyAdminService {
   }
 
   // GetDomain returns the confiuration information for a given domain.
-  rpc GetDomain(GetDomainRequest) returns (GetDomainResponse) {
+  rpc GetDomain(GetDomainRequest) returns (Domain) {
```
phad: Replacing GetDomainResponse{Domain} with Domain is fine if you're feeling confident that we'll never want to associate any further data with the Domain in the Get response.

core/api/v1/keytransparency_proto/keytransparency.proto
```diff
@@ -198,25 +194,20 @@ message UpdateEntryResponse {
   GetEntryResponse proof = 1;
 }
 
-// GetMutationsRequest contains the input parameters of the GetMutation APIs.
-message GetMutationsRequest {
+// GetEpochRequest identifies a particular epoch.
```
phad: Nit on the wording:  the parameters within GetEpochRequest are to identify the particular epoch, but the GetEpochRequest itself is to retrieve the mutations and related proofs

Could you say:  'GetEpochRequest identifies a particular epoch when retrieving the epoch's mutations and related proofs'?

core/api/v1/keytransparency_proto/keytransparency.proto
```diff
@@ -236,57 +227,58 @@ message GetMutationsResponse {
...
+//   Each Epoch contains the root of the sparse merkle tree and the changes that 
+//   occured that epoch and the previous.  
+// - Each Domain has a collection of applications named `/v1/domains/*/apps/`.
+//   Each Application has distinct users and permissions, but are indistinguishable 
+//   from eachother just by looking at the tree.
+// - Each Application has a collection of Users named `/v1/domains/*/apps/*/users`.
+//   Each User contains public key material, permissions for who is allowed to change 
+//   that user, and signatures indicating who made the last change.  
+// - Each User also has a collection of historical values for user named   
+//   `/v1/domains/*/apps/*/users/*/history`.
```
phad: The resource model description is great.  It's inherently hierarchical, so the comments could reflect that:

// The API has a collection of domains:
// `/v1/domains/`
//      * each domain has a single sparse Merkle tree, append-only log,
//        and other public key material ...
//        Each domain has a collection of snapshots, called Epochs
// `/v1/domains/*/epochs/`
//      * each epoch contains the root of the sparse Merkle tree...
//         etc.
// `/v1/domains/*/apps/`
//         etc.
// `/v1/domains/*/apps/*/users`
//         etc.
// `/v1/domains/*/apps/*/users/*/history`
//         etc.

core/api/v1/keytransparency_proto/keytransparency.proto
```diff
@@ -236,57 +227,58 @@ message GetMutationsResponse {
...
+  // GetDomain returns the information needed to verify the specified domain.
+  rpc GetDomain(GetDomainRequest) returns (Domain) {
+    option (google.api.http) = { get: "/v1/domains/{domain_id}" };
+  }
+
+  // GetEpoch returns a list of mutations between an epoch and its predecessor.
+  //
+  // Returns a list of mutations and their inclusion proofs along with the epoch
+  // signed map root.
+  rpc GetEpoch(GetEpochRequest) returns (Epoch){
```
phad: missing space at end: `(Epoch){`  -> `(Epoch) {`

core/api/v1/keytransparency_proto/admin.proto
phad: Oh I just saw #885 which explains why this is the way it now is.

core/api/monitor/v1/monitor_proto/monitor.proto
gdbelvin: thanks

core/api/monitor/v1/monitor_proto/monitor.proto
gdbelvin: done

core/api/v1/keytransparency_proto/admin.proto
gdbelvin: yep

core/api/v1/keytransparency_proto/keytransparency.proto
gdbelvin: This is now `GetState`

core/api/v1/keytransparency_proto/keytransparency.proto
gdbelvin: Thanks for this suggestion.

core/api/v1/keytransparency_proto/keytransparency.proto
gdbelvin: Thanks

phad:
Is this PR ready for me to look at yet?

gdbelvin:
Not yet.  I'll ping you when Travis is green.

core/api/v1/keytransparency_proto/keytransparency.proto
```diff
@@ -238,6 +238,19 @@ message Epoch {
 //  * Each Epoch contains the root of the sparse merkle tree and the changes that
 //    occured that epoch and the previous.
 //
+// Each Domain has a collection of applications:
+// `/v1/domains/*/apps/`
+//  * Each Application has distinct users and permissions, but are indistinguishable
+//    from eachother just by looking at the tree.
```
phad: nit: 'each other'

core/api/v1/keytransparency_proto/keytransparency.proto
```diff
@@ -217,7 +217,26 @@ message Epoch {
   repeated bytes log_consistency = 4;
   // log_inclusion proves that smr is part of log_root at index=srm.MapRevision.
   repeated bytes log_inclusion = 5;
-  // mutation contains mutation information.
+}
+
+// ListMutationsRequest contains the input parameters of the GetMutation APIs.
+message ListMutationsRequest {
+  // domain_id is the domain for which epochs are being requested.
```
phad: epochs?  not mutations?

core/api/v1/keytransparency_proto/keytransparency.proto
```diff
@@ -217,7 +217,26 @@ message Epoch {
...
+  string domain_id = 5;
+  // epoch specifies the epoch number in which mutations will be returned.
+  int64 epoch = 1;
+  // page_token defines the starting point for pagination. An empty
+  // value means start from the beginning. A non-empty value requests the next
```
phad: if all there is to this is empty/non-empty then it could be bool!

The non-empty value must be coming from somewhere, and must encoding the particular page start point, I'd imagine?

core/api/v1/keytransparency_proto/keytransparency.proto
```diff
@@ -273,6 +298,16 @@ service KeyTransparencyService {
     option (google.api.http) = { get: "/v1/domains/{domain_id}/epochs:stream" };
   }
 
+  // ListMutations returns a list of mutations in a specific epoch.
+  rpc ListMutations(ListMutationsRequest) returns (ListMutationsResponse) {
+    option (google.api.http) = { get: "/v1/domains/{domain_id}/epochs/{epoch}/mutations" };
+  }
+
+  // ListMutationsStream is a streaming list of mutatinos in a specific epoch.
```
phad: sp. 'mutatinos'  -  this is some newly discovered elementary particle right?

core/keyserver/epochs.go
```diff
@@ -35,33 +30,19 @@ import (
...
-		factory:   factory,
-	}
-}
+var (
+	// Each page contains pageSize profiles. Each profile contains multiple
+	// keys. Assuming 2 keys per profile (each of size 2048-bit), a page of
+	// size 16 will contain about 8KB of data.
+	defaultPageSize = int32(16)
+	// Maximum allowed requested page size to prevent DOS.
+	maxPageSize = int32(16)
```
phad: is it odd to have default == max ?

core/keyserver/epochs.go
```diff
@@ -82,18 +63,49 @@ func (s *Server) GetMutations(ctx context.Context, in *pb.GetMutationsRequest) (
 		return nil, status.Error(codes.Internal, "Get signed map root failed")
 	}
 
-	// Get highest and lowest sequence number.
-	meta, err := internal.MetadataFromMapRoot(resp.GetMapRoot())
+	// MapRevisions start at 0. Log leaf indexes starts at 0.
```
phad: nit: indices

core/keyserver/epochs.go
```diff
@@ -82,18 +63,49 @@ func (s *Server) GetMutations(ctx context.Context, in *pb.GetMutationsRequest) (
...
+		LogRoot:        logRoot.GetSignedLogRoot(),
+		LogConsistency: logConsistency.GetProof().GetHashes(),
+		LogInclusion:   logInclusion.GetProof().GetHashes(),
+	}, nil
+}
 
-	highestSeq := uint64(meta.GetHighestFullyCompletedSeq())
-	lowestSeq, err := s.lowestSequenceNumber(ctx, domain.MapID, in.PageToken, in.Epoch-1)
+// GetEpochStream is a streaming API similar to GetMutations.
+func (s *Server) GetEpochStream(in *pb.GetEpochRequest, stream pb.KeyTransparencyService_GetEpochStreamServer) error {
```
phad: you can replace the func args with _ since they're not in use yet.

core/keyserver/epochs.go
```diff
@@ -248,11 +272,7 @@ func (s *Server) inclusionProofs(ctx context.Context, domainID string, indexes [
...
 }
 
-//
-// Streaming RPCs
-//
-
-// GetMutationsStream is a streaming API similar to GetMutations.
-func (s *Server) GetMutationsStream(in *pb.GetMutationsRequest, stream pb.MutationService_GetMutationsStreamServer) error {
-	return grpc.Errorf(codes.Unimplemented, "GetMutationsStream is unimplemented")
+// ListMutationsStream is a streaming list of mutatinos in a specific epoch.
```
phad: same tyop here.

core/keyserver/epochs.go
```diff
@@ -117,47 +129,87 @@ func (s *Server) GetMutations(ctx context.Context, in *pb.GetMutationsRequest) (
...
-		NextPageToken:  nextPageToken,
+	return &pb.ListMutationsResponse{
+		Mutations:     mutations,
+		NextPageToken: nextPageToken,
 	}, nil
 }
 
-func (s *Server) logProofs(ctx context.Context, logID, firstTreeSize int64, epoch int64) (*tpb.GetLatestSignedLogRootResponse, *tpb.GetConsistencyProofResponse, *tpb.GetInclusionProofResponse, error) {
+// lowestSequenceNumber picks a lower bound on sequence number.
+// It returns the max between token and the high water mark of the previous epoch.
```
phad: are these three funcs pure moves and no edits along the way?

core/keyserver/epochs.go
```diff
@@ -117,47 +129,87 @@ func (s *Server) GetMutations(ctx context.Context, in *pb.GetMutationsRequest) (
...
+		return 0, err
+	}
+	tokenSeq, err := s.parseToken(token)
+	if err != nil {
+		return 0, err
+	}
+	return max(lastSeq, tokenSeq), nil
+}
+
+// getHighestFullyCompeltedSeq fetches the map root at epoch and returns
```
phad: typo: Compelted

core/api/v1/keytransparency_proto/keytransparency.proto
gdbelvin: The idea is that `page_token` should be opaque to the client. 
Fetching the next page involves passing the `next_page_token` from the previous response. 
To start from zero, don't send `page_token`.

core/keyserver/epochs_test.go
```diff
@@ -173,15 +178,21 @@ func TestLowestSequenceNumber(t *testing.T) {
 		{"4", 0, 4, true},
 		{"4", 1, 4, true},
 		{"some_token", 0, 0, false},
-		{"", 1, 6, true},
+		{"", 2, 6, true},
```
phad: it's not totally obvious from github's diffs shown here why 1 became 2.
I can see that lowestSequenceNumber has changed a fair bit but without detailed poking about I'm not sure what's prompted 1 -> 2

core/api/v1/keytransparency_proto/keytransparency.proto
gdbelvin: Thanks.  I missed this one in my API cleanup.

core/api/v1/keytransparency_proto/keytransparency.proto
gdbelvin: Lol.  The elementary particles discovered when typing too late at night :-)

core/api/v1/keytransparency_proto/keytransparency.proto
gdbelvin: Done

core/api/v1/keytransparency_proto/keytransparency.proto
gdbelvin: Updated the documentation. To fetch the next page, the client supplies the`next_page_token` value from the previous response.

core/api/v1/keytransparency_proto/keytransparency.proto
gdbelvin: Fixed.

core/keyserver/epochs.go

core/keyserver/epochs.go
gdbelvin: Good idea.

core/keyserver/epochs.go
gdbelvin: Fixed

core/keyserver/epochs.go
gdbelvin: Fixed. Replaced with 2048

core/keyserver/epochs.go

core/keyserver/epochs.go
gdbelvin: Done

core/keyserver/epochs.go
gdbelvin: Done

core/keyserver/epochs.go
gdbelvin: ok

core/keyserver/epochs.go
gdbelvin: Done

core/keyserver/epochs_test.go
gdbelvin: I'm not sure what the previous code was doing, but 2 seems like the correct number for the following reasons: 
- The test setup prepares two epochs:
   - Epoch 1 with a highest sequenced value of 6
   - Epoch 2 with a highest sequenced value of 10
- When asking for the mutations associated with epoch 2, and no `page_token`, we should start with 6 and go to 10.

core/client/mutationclient/client.go
```diff
@@ -37,74 +38,79 @@ const pageSize = 16
...
-				response <- monitorResp
-				epoch++
+// StreamEpochs repeatedly fetches epochs and sends them to out until GetEpoch
+// returns an error other than NotFound or until ctx.Done is closed.  When
+// GetEpoch returns NotFound, it waits one pollPeriod before trying again.
+func (c *Client) StreamEpochs(ctx context.Context, domainID string, startEpoch int64, out chan<- *pb.Epoch) error {
+	wait := time.NewTicker(c.pollPeriod).C
+	for i := startEpoch; ; {
+		// time out if we exceed the poll period:
+		cctx, cancel := context.WithTimeout(ctx, c.pollPeriod)
```
phad: What do you think about scoping cctx within a local func?
```
epoch, err := func() {
  ctx, cancel := context.WithTimeout(ctx, c.pollPeriod)
  defer cancel()
  return c.client.GetEpoch(ctx, ... )
}()
```
So the derived context doesn't live longer than necessary and cancel is done with defer which keeps it close to where it is obtained?

core/client/mutationclient/client.go
```diff
@@ -37,74 +38,79 @@ const pageSize = 16
...
-		req := &pb.GetMutationsRequest{
-			DomainId: domainID,
-			PageSize: pageSize,
-		}
-		resp, err := c.client.GetMutations(ctx, req, opts...)
+// EpochMutations fetches all the mutations in an epoch
+func (c *Client) EpochMutations(ctx context.Context, epoch *pb.Epoch) ([]*pb.MutationProof, error) {
+	mutations := []*pb.MutationProof{}
+	token := ""
+	for more := true; more; more = token != "" {
```
phad: This is a bit of an eyeful.  How about:
```
for {
  resp, err := ...
  if err != nil { ... }
  mutations = ...;
  token = resp.GetNextPageToken()
  if token == "" {
    break
  }
}
```

core/monitor/monitor.go
```diff
@@ -92,49 +93,104 @@ func New(mclient pb.MutationServiceClient,
...
 }
 
-// ProcessLoop continuously fetches mutations and processes them.
-func (m *Monitor) ProcessLoop(domainID string, period time.Duration) {
-	mutCli := mutationclient.New(m.mClient, period)
-	responses, errs := mutCli.StartPolling(domainID, 1)
-	for {
+// EpochPair is two adjacent epochs.
+type EpochPair struct {
+	EpochA, EpochB *pb.Epoch
```
phad: Could simply `A` and `B` suffice, since the type is `EpochPair` and the fields are `*pb.Epoch` ?

core/monitor/monitor.go
```diff
@@ -92,49 +93,104 @@ func New(mclient pb.MutationServiceClient,
...
+}
+
+// EpochPairs consumes epochs (0, 1, 2) and produces pairs (0,1), (1,2).
+func EpochPairs(ctx context.Context, epochs <-chan *pb.Epoch, pairs chan<- EpochPair) error {
```
phad: Neat.

core/monitor/monitor.go
```diff
@@ -92,49 +93,104 @@ func New(mclient pb.MutationServiceClient,
...
-		smr, err = m.signMapRoot(resp)
+// ProcessLoop continuously fetches mutations and processes them.
+func (m *Monitor) ProcessLoop(ctx context.Context, domainID string, period time.Duration) error {
+	mutCli := mutationclient.New(m.mClient, period)
+	cctx, cancel := context.WithCancel(ctx)
+	errc := make(chan error)
+	epochs := make(chan *pb.Epoch)
+	pairs := make(chan EpochPair)
+
+	go func(ctx context.Context, domainID string, epochs chan<- *pb.Epoch) {
```
phad: I think it ought to be fine not to pass `domainID` and `epochs` in as arguments, they should be captured from the outer scope?

core/monitor/monitor.go
```diff
@@ -92,49 +93,104 @@ func New(mclient pb.MutationServiceClient,
...
+		errc <- mutCli.StreamEpochs(ctx, domainID, 0, epochs)
+	}(cctx, domainID, epochs)
+	go func(ctx context.Context, epochs <-chan *pb.Epoch, pairs chan<- EpochPair) {
```
phad: The same should be the case here with `epochs` and `pairs`.  I think.

core/monitor/monitor.go
```diff
@@ -92,49 +93,104 @@ func New(mclient pb.MutationServiceClient,
...
-		Smr:      smr,
-		Seen:     time.Now(),
-		Errors:   errs,
-		Response: resp,
-	}); err != nil {
-		return fmt.Errorf("monitorstorage.Set(%v, _): %v", resp.Epoch, err)
+	errA := <-errc
+	errB := <-errc
+	if errA != nil || errB != nil {
+		return fmt.Errorf("failed fetching epochs: (fetch: %v, pair: %v)", errA, errB)
```
phad: Is it racy to assume that `errA` and `errB` will be in this particular order?
Could the pairing go-routine emit an error to `errc` terminate before the fetching go-routine does?

core/monitor/monitor.go
```diff
@@ -92,49 +93,104 @@ func New(mclient pb.MutationServiceClient,
...
+// VerifyEpochMutations validates that epochA + mutations = epochB.
+func (m *Monitor) VerifyEpochMutations(epochA, epochB *pb.Epoch, mutations []*pb.MutationProof) []error {
+	revision := epochB.GetSmr().GetMapRevision()
+	if errs := m.VerifyEpoch(epochB); len(errs) > 0 {
+		glog.Errorf("Invalid Epoch %v: %v", revision, errs)
+		return errs
+	}
+
+	// Fetch Previous root.
+	SMRA := epochA.GetSmr()
```
phad: I'd find smrA and smrB more readable somehow.  But we are supposed to capitalize acronyms.  SMRa / SMRb ?

core/monitor/verify.go
```diff
@@ -90,64 +90,45 @@ func (e *ErrList) Proto() []*statuspb.Status {
 	return errs
 }
 
-// VerifyMutationsResponse verifies a response received by the GetMutations API.
+// VerifyEpoch verifies a response received by the GetMutations API.
 // Additionally to the response it takes a complete list of mutations. The list
```
phad: nit: "In addition to ..."

core/monitorstorage/storage.go
```diff
@@ -52,6 +52,7 @@ type Result struct {
 }
 
 // Interface is the interface that stores and retrieves monitoring results.
+// TODO(gbelvin): make multi-tennent.
```
phad: tenant.  Tennent is a type of high alcohol lager favoured in Scotland :)

integration/monitor_test.go
```diff
@@ -107,20 +104,17 @@ func TestMonitor(t *testing.T) {
...
 
-		mutResp, err := mutCli.PollMutations(ctx, env.Domain.DomainId, tc.queryEpoch)
-		if err != nil {
-			t.Fatalf("Could not query mutations: %v", err)
-		}
-
-		if err := mon.Process(mutResp); err != nil {
-			t.Fatalf("Monitor could not process mutations: %v", err)
+		domainID := env.Domain.DomainId
+		cctx, cancel := context.WithTimeout(ctx, 30*time.Millisecond)
```
phad: You could do the scoped local func / defer thing here too if you want.
It's all very small, so maybe not worthwhile at present, but could be if someone added code after `ProcessLoop` without thinking about the `cancel`.

core/monitor/monitor.go
gdbelvin: Done.

core/monitor/monitor.go
gdbelvin: Done.

core/monitor/monitor.go
gdbelvin: :-)

core/monitor/monitor.go
gdbelvin: Sounds good.

core/client/mutationclient/client.go
gdbelvin: Good idea.

core/monitor/monitor.go
gdbelvin: Technically yes, but I don't know how that plays out when we're naming variables that should be private.  It's odd to have a private variable that begins with a capital no? `smrA`.

core/monitor/verify.go
gdbelvin: Updated this old comment.

core/monitorstorage/storage.go
gdbelvin: Thanks.

integration/monitor_test.go
gdbelvin: The reason I don't do it here is because I want to _block_, let `ProcessLoop` do it's thing,  timeout, then check the results. If I did it in parallel, I'd have a race condition.

gdbelvin:
Ready for another look.  
Sorry about all the force push's.  This PR went through _many_ rebase operations to gently pull out separate units of work and to keep it inline with HEAD.

core/keyserver/validate.go
```diff
@@ -48,6 +48,8 @@ var (
 	// ErrInvalidStart occurs when the start epoch of ListEntryHistoryRequest
 	// is not valid (not in [1, currentEpoch]).
 	ErrInvalidStart = errors.New("invalid start epoch")
+	// ErrInvalidPageSize occurs when the page size is > 0.
```
phad: "page size is < 0"

core/monitor/monitor_test.go
```diff
@@ -0,0 +1,57 @@
...
+	ctx := context.Background()
+	for _, tc := range []struct {
+		in  []int64
+		out []struct {
+			a, b int64
+		}
+	}{
+		{in: []int64{0, 1, 2}, out: []struct{ a, b int64 }{{0, 1}, {1, 2}}},
+	} {
+		t.Log("begin")
```
phad: remove the t.Log() / t.Logf() calls if you're happy everything works.

integration/monitor_test.go
phad: OK.  I can't see the original code with comments in context so we'll leave this.

core/keyserver/validate.go
gdbelvin: Doh!

core/monitor/monitor_test.go
gdbelvin: Done

PR #885, opened 2017-12-12
Cleanup the admin api

gdbelvin:
Per the API style guide, entity endpoints should return entities.
Let's do that.


phad:
LGTM post-merge.

PR #887, opened 2017-12-15
Remove unused Sequencer API

gdbelvin:
The sequencer API was created for a unwritten feature that streamed new epochs from the sequencer through the front ends, back out to monitors. 

However
- the next PR contains `GetEpoch` and `StreamEpochs` as a front end API feature.  We no longer need a separate API definition. 
- Since the Trillian Log gates the creation of new epochs anyway, we no longer need to stream them from the sequencer back to the front ends.  Front ends can poll the log themselves for new tree roots, reducing much complexity. 

Removing this API
- Reduces inter-proto dependencies
- Removes an "internal only" API definition in the `/core/proto` package.
- Removes an unnecessary proto package and the maintenance costs with it.

As a result, as of this PR, all the protos now exist in `core/api` and we can move the `gen.go` file to that directory. :-)


core/api/gen.go
```diff
@@ -12,16 +12,15 @@
...
 
-//go:generate protoc -I=. -I=$GOPATH/src/github.com/google/trillian/ -I=$GOPATH/src/github.com/googleapis/googleapis/ --go_out=,plugins=grpc:$GOPATH/src api/monitor/v1/monitor_proto/monitor.proto
-//go:generate protoc -I=. -I=$GOPATH/src/github.com/google/trillian/ -I=$GOPATH/src/github.com/googleapis/googleapis/ --grpc-gateway_out=logtostderr=true:. api/monitor/v1/monitor_proto/monitor.proto
+//go:generate protoc -I=. -I=$GOPATH/src/github.com/google/trillian/ -I=$GOPATH/src/github.com/googleapis/googleapis/ --go_out=,plugins=grpc:$GOPATH/src monitor/v1/monitor_proto/monitor.proto
+//go:generate protoc -I=. -I=$GOPATH/src/github.com/google/trillian/ -I=$GOPATH/src/github.com/googleapis/googleapis/ --grpc-gateway_out=logtostderr=true:. monitor/v1/monitor_proto/monitor.proto
 
-//go:generate protoc -I=. -I=$GOPATH/src/github.com/google/trillian/ -I=$GOPATH/src/github.com/googleapis/googleapis --go_out=:$GOPATH/src api/usermanager/v1/usermanager_proto/usermanager.proto
```
phad: what happened to this line?

core/api/v1/keytransparency_proto/keytransparency.proto
```diff
@@ -28,7 +28,7 @@ import "crypto/keyspb/keyspb.proto";
 import "crypto/sigpb/sigpb.proto";
 import "trillian.proto";
 import "trillian_map_api.proto";
-import "api/v1/keytransparency_proto/admin.proto";
+import "v1/keytransparency_proto/admin.proto";
```
phad: Is the removal of 'api/' here a consequence of moving the gen.go file into the api/ dir?

core/api/gen.go
gdbelvin: I did some experiments and discovered that the following two lines are redundant, since the grpc invocation will also compile the proto message objects. 
- `--go_out=:$GOPATH/src`
- `--go_out=,plugins=grpc:$GOPATH/src`

core/api/v1/keytransparency_proto/keytransparency.proto
gdbelvin: That's right. Due to the `-I=.` line in the protoc invocation, the root search path for imports will be `core/api`.

core/api/gen.go
phad: Cool, thanks for explaining.

PR #890, opened 2017-12-15
Add default UKNOWN values to enums

gdbelvin:


core/api/type/type_proto/keymaster.proto
```diff
@@ -48,8 +49,9 @@ message SigningKey {
 message VerifyingKey {
   // KeyStatus defines a key status.
   enum KeyStatus {
-    ACTIVE = 0;      // indicates a key that is currently being used.
-    DEPRECATED = 1;  // indicates a key that is not in use.
+    UKNOWN = 0;
+    ACTIVE = 2;      // indicates a key that is currently being used.
```
phad: Should it be ACTIVE = 1, DEPRECATED = 3?
Or are you trying to keep numbers in sync with those in KeyStatus?
(If so, I wouldn't try)

gdbelvin: I'm not explicitly trying to keep them in sync, but it's not a bad idea..

core/api/type/type_proto/keymaster.proto
```diff
@@ -32,9 +32,10 @@ message Metadata {
 message SigningKey {
   // KeyStatus defines a key status.
   enum KeyStatus {
-    ACTIVE = 0;      // indicates a key that is currently being used.
-    INACTIVE = 1;    // indicates a key status that is not active.
-    DEPRECATED = 2;  // indicates a key that is not in use.
+    UKNOWN = 0;
```
phad: typo: UNKNOWN

core/api/type/type_proto/keymaster.proto
```diff
@@ -48,8 +49,9 @@ message SigningKey {
 message VerifyingKey {
   // KeyStatus defines a key status.
   enum KeyStatus {
-    ACTIVE = 0;      // indicates a key that is currently being used.
-    DEPRECATED = 1;  // indicates a key that is not in use.
+    UKNOWN = 0;
```
phad: typo: UNKNOWN

core/api/type/type_proto/keymaster.proto
```diff
@@ -48,8 +49,9 @@ message SigningKey {
...
+    ACTIVE = 1;      // indicates a key that is currently being used.
+    DEPRECATED = 3;  // indicates a key that is not in use.
```
phad: I'm not sure of the benefit of aligning the enum values like this.  I'd just use 2 here.  There should not be any casting between the two KeyStatus types in any case!

core/api/type/type_proto/keymaster.proto
gdbelvin: Woops.  Thanks.

core/api/type/type_proto/keymaster.proto
gdbelvin: Yeah, no casting indeed.

PR #897, opened 2018-01-03
 Set a maximum batch size for sequencing

gdbelvin:
Prevent a crazy number of mutations from going into a single epoch
and overloading the trillian map.

DiffBase #896

### DIFF
--- a/core/fake/mutation_storage.go
+++ b/core/fake/mutation_storage.go
@@ -32,28 +32,28 @@ func NewMutationStorage() *MutationStorage {
 	}
 }
 
-// ReadRange returns the list of mutations
-func (m *MutationStorage) ReadRange(ctx context.Context, mapID int64, startSequence uint64, endSequence uint64, count int32) (uint64, []*pb.EntryUpdate, error) {
-	if startSequence > uint64(len(m.mtns[mapID])) {
-		panic("startSequence > len(m.mtns[mapID])")
+// ReadPage paginates through the list of mutations
+func (m *MutationStorage) ReadPage(_ context.Context, mapID, start, end int64, pageSize int32) (int64, []*pb.EntryUpdate, error) {
+	if start > int64(len(m.mtns[mapID])) {
+		panic("start > len(m.mtns[mapID])")
 	}
-	// Adjust endSequence.
-	if endSequence-startSequence > uint64(count) {
-		endSequence = startSequence + uint64(count)
+	// Adjust end.
+	if end-start > int64(pageSize) {
+		end = start + int64(pageSize)
 	}
-	if endSequence > uint64(len(m.mtns[mapID])) {
-		endSequence = uint64(len(m.mtns[mapID]))
+	if end > int64(len(m.mtns[mapID])) {
+		end = int64(len(m.mtns[mapID]))
 	}
-	return endSequence, m.mtns[mapID][startSequence:endSequence], nil
+	return end, m.mtns[mapID][start:end], nil
 }
 
-// ReadAll is unimplemented
-func (m *MutationStorage) ReadAll(ctx context.Context, mapID int64, start uint64, count int) (uint64, []*pb.EntryUpdate, error) {
+// ReadBatch is unimplemented
+func (m *MutationStorage) ReadBatch(context.Context, int64, int64, int32) (int64, []*pb.EntryUpdate, error) {
 	return 0, nil, nil
 }
 
 // Write stores a mutation
-func (m *MutationStorage) Write(ctx context.Context, mapID int64, mutation *pb.EntryUpdate) (uint64, error) {
+func (m *MutationStorage) Write(_ context.Context, mapID int64, mutation *pb.EntryUpdate) (int64, error) {
 	m.mtns[mapID] = append(m.mtns[mapID], mutation)
-	return uint64(len(m.mtns[mapID])), nil
+	return int64(len(m.mtns[mapID])), nil
 }
--- a/core/fake/mutation_storage.go
+++ b/core/fake/mutation_storage.go
@@ -48,7 +48,7 @@ func (m *MutationStorage) ReadRange(ctx context.Context, mapID int64, startSeque
 }
 
 // ReadAll is unimplemented
-func (m *MutationStorage) ReadAll(ctx context.Context, mapID int64, startSequence uint64) (uint64, []*pb.EntryUpdate, error) {
+func (m *MutationStorage) ReadAll(ctx context.Context, mapID int64, start uint64, count int) (uint64, []*pb.EntryUpdate, error) {
 	return 0, nil, nil
 }
 
--- a/core/mutator/mutator.go
+++ b/core/mutator/mutator.go
@@ -67,7 +67,8 @@ type MutationStorage interface {
 	// ReadAll reads all mutations starting from the given sequence number.
 	// Note that startSequence is not included in the result. ReadAll also
 	// returns the maximum sequence number read.
-	ReadAll(ctx context.Context, mapID int64, startSequence uint64) (uint64, []*pb.EntryUpdate, error)
+	// ReadAll will not return more than count entries.
+	ReadAll(ctx context.Context, mapID int64, startSequence uint64, count int) (uint64, []*pb.EntryUpdate, error)
 	// Write saves the mutation in the database. Write returns the sequence
 	// number that is written.
 	Write(ctx context.Context, mapID int64, mutation *pb.EntryUpdate) (uint64, error)
--- a/core/mutator/mutator.go
+++ b/core/mutator/mutator.go
@@ -58,18 +58,15 @@ type Mutator interface {
 
 // MutationStorage reads and writes mutations to the database.
 type MutationStorage interface {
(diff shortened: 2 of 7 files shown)


core/fake/mutation_storage.go
```diff
@@ -33,7 +33,7 @@ func NewMutationStorage() *MutationStorage {
 }
 
 // ReadRange returns the list of mutations
-func (m *MutationStorage) ReadRange(txn transaction.Txn, mapID int64, startSequence uint64, endSequence uint64, count int32) (uint64, []*pb.EntryUpdate, error) {
+func (m *MutationStorage) ReadRange(ctx context.Context, mapID int64, startSequence uint64, endSequence uint64, count int32) (uint64, []*pb.EntryUpdate, error) {
```
phad: Consider using _ for ctx here, as it is unused. Same in the two funcs below where args are not in use.

core/mutator/mutator.go
```diff
@@ -67,7 +67,8 @@ type MutationStorage interface {
 	// ReadAll reads all mutations starting from the given sequence number.
 	// Note that startSequence is not included in the result. ReadAll also
 	// returns the maximum sequence number read.
-	ReadAll(ctx context.Context, mapID int64, startSequence uint64) (uint64, []*pb.EntryUpdate, error)
+	// ReadAll will not return more than count entries.
```
phad: It seems like it might need a new name given that for suitable `count` values it doesn't read all mutations any more.
Maybe adding `count` is a hint that we now have a paginated read API?  So it could be called `ReadPage`?

phad: Or `ReadBatch` since you talk about batch sizes in the sequencer.

core/sequencer/sequencer.go
```diff
@@ -58,6 +58,9 @@ var (
 	})
 )
 
+// MaxBatchSize limits the number of mutations that will be processed per epoch.
+var MaxBatchSize = 1000
```
phad: Batch?  Page?  Would be good to have consistent terminology with the `MutationStorage.Read` API itself.

core/mutator/mutator.go
```diff
@@ -67,7 +67,8 @@ type MutationStorage interface {
...
+	ReadAll(ctx context.Context, mapID int64, startSequence uint64, count int) (uint64, []*pb.EntryUpdate, error)
```
phad: Here `count` is `int` but in `ReadRange` it is `int32`.

impl/sql/mutationstorage/mutations_test.go
```diff
@@ -246,7 +230,8 @@ func TestReadAll(t *testing.T) {
 			},
 		},
 	} {
-		maxSequence, results, err := readAll(ctx, m, tc.startSequence)
+		maxcount := 1000
```
phad: Nit: `maxCount`. Do you want this to be `const`?

Perhaps it ought to be a field of the test table so you can test with boundary and invalid values?

core/fake/mutation_storage.go
gdbelvin: Done

core/mutator/mutator.go
gdbelvin: I'll make them both int32

core/sequencer/sequencer.go
gdbelvin: This isn't perfect, but I'm trying on `ReadPage` and `ReadBatch` for size. 
`ReadPage` lists mutations by page between a start and end sequence number.
`ReadBatch` lists mutations by batch between a start sequence number to the current sequence number.

core/mutator/mutator.go
gdbelvin: `ReadRage` is becoming `ReadPage`, and `ReadAll` is becoming `ReadBatch`.  Is that too confusing?

impl/sql/mutationstorage/mutations_test.go
gdbelvin: added

gdbelvin:
Updated. PTAL

core/mutator/mutator.go
phad: It's a bit confusing.  It feels like they're both doing the same thing really, and the main difference is whether it's the client doing a bit of math on start/end/count or the API implementation.  Does the client know currentSequence?

Another possibility might be for `count` not to be supplied through the API at all, but for the implementation to make it's own decision about whether to truncate output or not, and to indicate with a flag if truncation has occurred.  wdyt?

core/mutator/mutator.go
phad: ok

core/fake/mutation_storage.go
```diff
@@ -32,28 +32,28 @@ func NewMutationStorage() *MutationStorage {
 	}
 }
 
-// ReadRange returns the list of mutations
-func (m *MutationStorage) ReadRange(ctx context.Context, mapID int64, startSequence uint64, endSequence uint64, count int32) (uint64, []*pb.EntryUpdate, error) {
-	if startSequence > uint64(len(m.mtns[mapID])) {
+// ReadPage returns the list of mutations
```
phad: Maybe say 'ReadPage paginates through the list of mutations'

core/fake/mutation_storage.go
```diff
@@ -32,28 +32,28 @@ func NewMutationStorage() *MutationStorage {
...
+func (m *MutationStorage) ReadPage(_ context.Context, mapID, start, end int64, pageSize int32) (int64, []*pb.EntryUpdate, error) {
+	if start > int64(len(m.mtns[mapID])) {
 		panic("startSequence > len(m.mtns[mapID])")
```
phad: update panic text

core/fake/mutation_storage.go
```diff
@@ -32,28 +32,28 @@ func NewMutationStorage() *MutationStorage {
...
 	}
 	// Adjust endSequence.
```
phad: update comment

core/mutator/mutator.go
```diff
@@ -58,18 +58,14 @@ type Mutator interface {
...
-	// first. ReadRange also returns the maximum sequence number read.
-	ReadRange(ctx context.Context, mapID int64, startSequence, endSequence uint64, count int32) (uint64, []*pb.EntryUpdate, error)
-	// ReadAll reads all mutations starting from the given sequence number.
-	// Note that startSequence is not included in the result. ReadAll also
-	// returns the maximum sequence number read.
-	// ReadAll will not return more than count entries.
-	ReadAll(ctx context.Context, mapID int64, startSequence uint64, count int) (uint64, []*pb.EntryUpdate, error)
+	// ReadPage returns mutations in the interval (start, end] for mapID.
+	// pageSize specifies the maximum number of items to return.
+	// Returns the maximum sequence number returned.
```
phad: If pageSize were left to the implementation then the client would be able to determine if truncation had occured, right?  Since the returned maxSeqNum would be < end

core/mutator/mutator.go
```diff
@@ -58,18 +58,14 @@ type Mutator interface {
...
+	ReadPage(ctx context.Context, mapID, start, end int64, pageSize int32) (int64, []*pb.EntryUpdate, error)
+	// ReadBatch returns mutations in the interval (start, ∞] for mapID.
+	// ReadBatch will not return more than batchSize entries.
+	ReadBatch(ctx context.Context, mapID, start int64, batchSize int32) (int64, []*pb.EntryUpdate, error)
```
phad: Doc the returned int64 here - looks like it's the same as for ReadPage

core/sequencer/sequencer.go
```diff
@@ -59,7 +59,7 @@ var (
 )
 
 // MaxBatchSize limits the number of mutations that will be processed per epoch.
-var MaxBatchSize = 1000
+var MaxBatchSize = int32(1000)
```
phad: this is const, isn't it?

core/fake/mutation_storage.go
gdbelvin: good idea.

core/fake/mutation_storage.go
gdbelvin: done

core/fake/mutation_storage.go
gdbelvin: done

core/mutator/mutator.go
gdbelvin: This API is used to paginate through the mutations in an epoch. As a result, the client doesn't really know `start` or `end`, which makes setting a `pageSize` useful.

core/sequencer/sequencer.go
gdbelvin: The thought is that this should be a flag at some point, but we'll leave that for later.

gdbelvin:
I can see where you're coming from on `ReadPage` and `ReadBatch` being pretty similar. 
`ReadBatch` at the moment functions like the dequing part of our queue.  I will be changing this shortly to separate queuing from storing mutations and keeping the APIs separate will help.

core/mutator/mutator.go
phad: OK, fair enough.

PR #900, opened 2018-01-04
Index mutations by domainID

gdbelvin:
This PR switches the mutation storage table from using `mapID` as the primary mutation key and replaces it with `domainID`.  This achieves a slightly cleaner abstraction and brings internal and external APIs into alignment.


gdbelvin:
This PR has been rebased and is ready for a second (much cleaner) look

PR #903, opened 2018-01-08
Separate queue from mutation storage

### LINKED ISSUE #902: Separate Queue from Mutation Storage
During an engineering review it was pointed out that a performance bottleneck was introduced when the queue and mutation storage were combined.  The queue's ability to absorb traffic would be greatly increased if we removed the strict ordering requirement from it.

gdbelvin:
Prior to this PR, mutations were de facto sequenced by the storage layer (via AUTOINCREMENT). 
This artificial combining of the queue and mutation storage introduced a bottleneck and some dual-purpose interfaces. 

This PR cleanly separates the queue from mutation storage. As a result:
- We are no longer bound by an `AUTOINCREMENT` field, which removes the bottleneck. 
- We no longer need to save a high watermark in the `SignedMapHead`
- The queue no longer needs to keep track of where it started. By definition now, anything in the queue has not yet been delivered.  

closes #902


core/fake/mutation_storage.go
```diff
@@ -16,50 +16,50 @@ package fake
...
+	"fmt"
 
 	pb "github.com/google/keytransparency/core/api/v1/keytransparency_proto"
-	"github.com/google/keytransparency/core/mutator"
 )
 
 // MutationStorage implements mutator.Mutation
 type MutationStorage struct {
-	mtns map[string][]*pb.EntryUpdate
+	// mtns[DomainID][Epoch]
```
phad: remove commented-out code?

phad: Or are you hinting at the values to be used for the string and int64 keys?

core/fake/mutation_storage.go
```diff
@@ -16,50 +16,50 @@ package fake
...
-	// Adjust end.
-	if end-start > int64(pageSize) {
-		end = start + int64(pageSize)
+	mutationList, ok := domain[revision]
+	if !ok {
+		return 0, nil, fmt.Errorf("DomainID: %v, revision %v not found", domainID, revision)
 	}
-	if end > int64(len(m.mtns[domainID])) {
-		end = int64(len(m.mtns[domainID]))
+	if got, want := int(start), len(mutationList); got >= want {
```
phad: I think the 'got' / 'want' idiom obscures the intent of your code.
I wouldn't use it outside of test code personally.

core/fake/mutation_storage.go
```diff
@@ -16,50 +16,50 @@ package fake
...
-	mutations := make([]*pb.Entry, 0, len(entryUpdates))
-	for _, e := range entryUpdates {
-		mutations = append(mutations, e.Mutation)
+	end := int(start) + int(pageSize)
+	if end > len(mutationList) {
+		end = len(mutationList)
 	}
-	return end, mutations, nil
+	mutations := mutationList[int(start):end]
+	return start + int64(len(mutations)), mutations, nil
```
phad: How about returning: `int64(end), mutations, nil` ?
If start + pageSize > end, then you've calculated how to clip end already.

core/keyserver/epochs.go
```diff
@@ -126,8 +120,8 @@ func (s *Server) ListMutations(ctx context.Context, in *pb.ListMutationsRequest)
 	}
 
 	nextPageToken := ""
-	if len(mutations) == int(in.PageSize) && maxSequence != highestSeq {
-		nextPageToken = fmt.Sprintf("%d", maxSequence)
+	if len(mutations) == int(in.PageSize) {
+		nextPageToken = fmt.Sprintf("%d", max+1)
```
phad: Has the meaning of max returned from ReadPage changed?  if not why the +1 here?

core/keyserver/epochs_test.go
```diff
@@ -38,47 +36,32 @@ const (
 	domainID = "domain"
 )
 
-func updates(t *testing.T, start, end int) []*pb.EntryUpdate {
-	if start > end {
-		t.Fatalf("start=%v > end=%v", start, end)
-	}
-	kvs := make([]*pb.EntryUpdate, 0, end-start)
+func genMutations(start, end int) []*pb.Entry {
+	mutations := make([]*pb.Entry, 0, end-start)
```
phad: Happy to just let this crash if `start` > `end`?

core/keyserver/epochs_test.go
```diff
@@ -38,47 +36,32 @@ const (
...
+			return err
 		}
-	}
-	tmap.SetLeaves(ctx, &tpb.SetMapLeavesRequest{
-		Metadata: mustMetadataAsAny(t, &pb.MapperMetadata{
-			HighestFullyCompletedSeq: int64(end),
-		}),
-	})
-}
+		tmap.SetLeaves(ctx, &tpb.SetMapLeavesRequest{})
```
phad: Is this still needed?

core/keyserver/epochs_test.go
```diff
@@ -38,47 +36,32 @@ const (
...
-	for _, kv := range kvs {
-		if _, err := mutations.Write(ctx, domainID, kv); err != nil {
-			t.Fatalf("mutations.Write failed: %v", err)
+func prepare(ctx context.Context, domainID string, m mutator.MutationStorage, tmap *fake.MapServer) error {
+	for _, rev := range []struct {
+		epoch      int64
+		start, end int
+	}{
+		{epoch: 1, start: 1, end: 6},
+		{epoch: 2, start: 7, end: 10},
```
phad: This table of epoch/start/end values look like test data but aren't part of the tests that follow.
  Should the table be passed in?

core/fake/mutation_storage.go
gdbelvin: It was meant as a comment. I've converted it now to english text.

core/fake/mutation_storage.go
gdbelvin: I was originally using `got` and `want` inside the if block, but now that that's no longer the case, you're right, it obscures the code's meaning.

core/fake/mutation_storage.go
gdbelvin: Great idea.

core/keyserver/epochs_test.go
gdbelvin: The meaning has indeed changed.  
`ReadPage` used to return the interval `(start, end]`, which is a bit odd, but served our purposes when we were using a high water mark.  Now it returns `[start, start+pageSize]`. As a result, we need to add 1 to fetch the next page.  Is there a better way of doing this?

core/keyserver/epochs_test.go
gdbelvin: I've moved this to the test itself.

core/keyserver/epochs_test.go
gdbelvin: SetLeaves advances the revision number. I'll add a comment.

gdbelvin:
Ready for another look

core/fake/mutation_storage.go
```diff
@@ -44,15 +44,14 @@ func (m *MutationStorage) ReadPage(_ context.Context, domainID string, revision,
 	if !ok {
 		return 0, nil, fmt.Errorf("DomainID: %v, revision %v not found", domainID, revision)
 	}
-	if got, want := int(start), len(mutationList); got >= want {
+	if int(start) > len(mutationList) {
```
phad: This has changed from `>=` to `>`  -- which is correct?

gdbelvin: It turns out that they are almost equivalent. 
```
mutationList[len(mutationlist):] = [] ~= nil
```

I think returning `[]` makes a little more sense though, so I'll keep it as is.

PR #911, opened 2018-01-22
Setup integration tests for dependency injection.

gdbelvin:
This PR removes the `/integration` directory and creates 
- `core/integration` - integration test functions and `*Env`.
- `impl/integration` - integration test runner and concrete implementation of `*Env`

This allows other implementations to also run the integration tests with their own `*Env` implementations.


integration/env.go
```diff
@@ -125,24 +112,24 @@ func NewEnv(t *testing.T) *Env {
...
-		t.Fatalf("Failed to load vrf pubkey: %v", err)
+		return nil, fmt.Errorf("env: Failed to load vrf pubkey: %v", err)
 	}
 
 	// Common data structures.
-	mutations, err := mutationstorage.New(sqldb)
+	mutations, err := mutationstorage.New(db)
 	if err != nil {
-		log.Fatalf("Failed to create mutations object: %v", err)
+		log.Fatalf("env: Failed to create mutations object: %v", err)
```
phad: Could this one be Errorf() like the previous ones you changed?

integration/main_test.go
```diff
@@ -0,0 +1,43 @@
...
+	"testing"
+
+	"github.com/google/trillian/storage/testdb"
+)
+
+// TestIntegration runs all KeyTransparency integration tests.
+func TestIntegration(t *testing.T) {
+	// We can only run the integration tests if there is a MySQL instance available.
+	if provider := testdb.Default(); !provider.IsMySQL() {
+		t.Skipf("Skipping map integration test, SQL driver is %v", provider.Driver)
```
phad: s/map/KT/  ?

integration/env.go
gdbelvin: yep, missed that one.

integration/main_test.go
gdbelvin: yep

PR #914, opened 2018-01-26
Fix integration tests

gdbelvin:
This PR follows the API changes in google/trillian#951

### DIFF
--- a/core/fake/trillian_log_client.go
+++ b/core/fake/trillian_log_client.go
@@ -33,33 +33,33 @@ func NewTrillianLogClient() *LogServer {
 }
 
 // QueueLeaf increments the size of the tree.
-func (l *LogServer) QueueLeaf(ctx context.Context, in *tpb.QueueLeafRequest, opts ...grpc.CallOption) (*tpb.QueueLeafResponse, error) {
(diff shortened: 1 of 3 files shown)


core/fake/trillian_log_client.go
```diff
@@ -68,21 +68,26 @@ func (l *LogServer) GetLatestSignedLogRoot(ctx context.Context, in *tpb.GetLates
 }
 
 // GetSequencedLeafCount is not implemented.
-func (l *LogServer) GetSequencedLeafCount(ctx context.Context, in *tpb.GetSequencedLeafCountRequest, opts ...grpc.CallOption) (*tpb.GetSequencedLeafCountResponse, error) {
+func (*LogServer) GetSequencedLeafCount(context.Context, *tpb.GetSequencedLeafCountRequest, ...grpc.CallOption) (*tpb.GetSequencedLeafCountResponse, error) {
 	panic("not implemented")
```
Martin2112: Unrelated to PR but should these return a canonical Unimplemented error rather than panic?

gdbelvin: Hmm.  Good point.  
I'm about to delete this in the next PR though.

PR #915, opened 2018-01-26
Add GetLatestEpoch API

gdbelvin:
The latest epoch is of interest in two use cases:
  - A client that is waiting to see if it's changes have made it into the next revision. 
  - A gossiping peer that may not support gRPC streaming APIs. 

By offering this API, clients will be able to efficiently poll just the Log Root until it changes before requesting a full proof for an updated entry.  This is more much more efficient and would remove the need for a client to needlessly resubmit their `UpdateEntry` requests before a new epoch has been created.


core/keyserver/epochs.go
```diff
@@ -34,27 +36,55 @@ var (
...
 	// Get signed map root by revision.
 	resp, err := s.tmap.GetSignedMapRootByRevision(ctx, &tpb.GetSignedMapRootByRevisionRequest{
-		MapId:    domain.MapID,
-		Revision: in.Epoch,
+		MapId:    d.MapID,
+		Revision: revision,
 	})
 	if err != nil {
-		glog.Warningf("GetSignedMapRootByRevision(%v, %v): %v", domain.MapID, in.Epoch, err)
+		glog.Warningf("GetSignedMapRootByRevision(%v, %v): %v", d.MapID, revision, err)
```
Martin2112: This style of error message might make it hard to track down where they occurred. Might want to prefix the text with the rpc method name.

core/keyserver/epochs.go
```diff
@@ -134,48 +164,28 @@ func (*Server) ListMutationsStream(in *pb.ListMutationsRequest, stream pb.KeyTra
...
-	*tpb.GetConsistencyProofResponse,
-	*tpb.GetInclusionProofResponse,
-	error) {
-	// Lookup log and map info.
-	logRoot, err := s.tlog.GetLatestSignedLogRoot(ctx,
-		&tpb.GetLatestSignedLogRootRequest{
-			LogId: logID,
-		})
+// logProofs returns the proofs for a given epoch.
+func (s *Server) logProofs(ctx context.Context, d *domain.Domain, firstTreeSize int64, epoch int64) (*tpb.SignedLogRoot, *tpb.Proof, *tpb.Proof, error) {
```
Martin2112: Wonder if the proofs for an epoch could be wrapped into a single object for a cleaner signature.

core/keyserver/keyserver.go
```diff
@@ -77,55 +77,66 @@ func New(tlog tpb.TrillianLogClient,
...
+	proto.Merge(resp, entryProof)
+	return resp, nil
+}
+
+// getEntryByRevision returns an entry and its proofs.
+// getEntryByRevision does NOT populate the following fields:
+// - LogRoot
+// - LogConsistency
+func (s *Server) getEntryByRevision(ctx context.Context, sth *tpb.SignedLogRoot, d *domain.Domain, userID, appID string, revision int64) (*pb.GetEntryResponse, error) {
+	if revision < int64(0) {
```
Martin2112: Is that cast required?

core/keyserver/keyserver.go
```diff
@@ -147,37 +158,19 @@ func (s *Server) getEntry(ctx context.Context, domainID, userID, appID string, f
...
-
-	// Inclusion proof.
+	// SignedMapHead to SignedLogRoot inclusion proof.
+	secondTreeSize := sth.GetTreeSize()
 	logInclusion, err := s.tlog.GetInclusionProof(ctx,
 		&tpb.GetInclusionProofRequest{
-			LogId: domain.LogID,
+			LogId: d.LogID,
 			// SignedMapRoot must be placed in the log at MapRevision.
 			// MapRevisions start at 1. Log leaves start at 1.
```
Martin2112: Can you clarify meaning of "log leaves start at 1" here as leaf indices are 0 based.

core/keyserver/keyserver.go
```diff
@@ -363,3 +367,57 @@ func (s *Server) GetDomain(ctx context.Context, in *pb.GetDomainRequest) (*pb.Do
...
+			glog.Errorf("tlog.GetConsistency(%v, %v, %v): %v",
+				d.LogID, firstTreeSize, secondTreeSize, err)
+			return nil, nil, status.Errorf(codes.Internal, "Cannot fetch log consistency proof")
+		}
+	}
+	return sth, logConsistency.GetProof(), nil
+}
+
+// latestRevision returns the latest map revision, given the latest sth.
+// The log is the authoritative source of the latest revision.
```
Martin2112: Naming seems a bit odd given it takes a log STH and returns a map revision. mapRevisionFor() maybe?

core/keyserver/keyserver.go
```diff
@@ -190,29 +183,35 @@ func (s *Server) getEntry(ctx context.Context, domainID, userID, appID string, f
...
+		Smr:          getResp.GetMapRoot(),
+		LogInclusion: logInclusion.GetProof().GetHashes(),
 	}, nil
 }
 
 // ListEntryHistory returns a list of EntryProofs covering a period of time.
 func (s *Server) ListEntryHistory(ctx context.Context, in *pb.ListEntryHistoryRequest) (*pb.ListEntryHistoryResponse, error) {
 	// Lookup log and map info.
-	domain, err := s.domains.Read(ctx, in.DomainId, false)
+	domainID := in.GetDomainId()
```
Martin2112: GetDomainId() - should be GetDomainID().
Also looks like you often get the id then do domains.Read() on it, which might be a helper perhaps?

core/keyserver/epochs.go
gdbelvin: Good idea.

core/keyserver/keyserver.go
gdbelvin: nope. Removed.

core/keyserver/epochs.go
gdbelvin: That's a bit better.

core/keyserver/keyserver.go
gdbelvin: The comment is incorrect.  Updated:
`MapRevisions start at 0. Log leaves start at 0.`

core/keyserver/keyserver.go
gdbelvin: `GetDomainId` is a function generated from protobuf. The protobuf people decided to _not_ follow Go's standard for initialisms unfortunately: golang/protobuf#156.

The emerging pattern is that all functions supply `domain_id` and the first thing they do is `domains.Read`. Not sure what the best way to pull out a helper.  It feels more like an interceptor that puts something in `ctx`, but that feels a little heavy handed and makes the code path harder to follow.

core/keyserver/keyserver.go
gdbelvin: cool

core/keyserver/keyserver.go
Martin2112: Ah, well it was just a suggestion.

gdbelvin:
@Martin2112 thank you for the review! 
I've pulled out the `Catch log initialized errors` commit into a separate PR: #916 and added unit tests.  PTAL?  Beers on me.

gdbelvin:
Now that #916 is in, this PR is a whole lot simpler!
PTAL

core/keyserver/epochs.go
```diff
@@ -36,27 +36,55 @@ var (
 	maxPageSize = int32(2048) // 8MB
 )
 
-// GetEpoch returns a list of mutations paged by epoch number.
+// GetLatestEpoch returns the latest epoch. The current epoch tracks the SignedLogRoot.
+func (s *Server) GetLatestEpoch(ctx context.Context, in *pb.GetLatestEpochRequest) (*pb.Epoch, error) {
+	// Lookup log and map info.
+	d, err := s.domains.Read(ctx, in.DomainId, false)
+	if err != nil {
+		glog.Errorf("adminstorage.Read(%v): %v", in.DomainId, err)
```
Martin2112: Might want to prefix the errors with the method. Up to you.

core/keyserver/epochs.go
```diff
@@ -36,27 +36,55 @@ var (
...
+	currentEpoch, err := mapRevisionFor(sth)
+	if err != nil {
+		glog.Errorf("mapRevisionFor(log %v, sth%v): %v", d.LogID, sth, err)
+		return nil, err
+	}
+
+	return s.getEpochByRevision(ctx, d, in.GetFirstTreeSize(), currentEpoch)
+}
+
+// GetEpoch returns a the requested epoch.
```
Martin2112: typo

core/keyserver/epochs.go

core/keyserver/epochs.go

PR #916, opened 2018-01-26
Catch log initialized errors

gdbelvin:
One of the first things the KeyTransparency sequencer does when initializing a fresh domain (Trillian Log / Map pair), is to put the Map's initial 0'th SignedMapHead into the Log at index 0.  This establishes the 1:1 relationship between Map revisions and the Log indexes at which they must be stored.  If this, however, does not occur, it causes follow on errors that would be nice to catch in a single place. 

Secondly, because clients verify that `SignedMapHead`s are included in the Trillian Log, the key server must not serve any data from a Map revision (SMH) that the Log has not sequenced and signed yet. The `SignedLogRoot` (STH) is effectively the source of truth for what map revision to serve. This was not accurately implemented in `GetEpochHistory` which was using the latest SMH rather than STH as it's source of truth for the `currentEpoch`. 

This PR refactors a few functions in `keyserver.go` to 
- Clearly define the latest SignedLogRoot as the source of truth for the latest map revision. 
- Establish a single code path for both retrieving the `currentEpoch` and verifying that it meets our initialized expectations.

### DIFF
--- a/core/keyserver/epochs.go
+++ b/core/keyserver/epochs.go
@@ -65,16 +65,16 @@ func (s *Server) GetEpoch(ctx context.Context, in *pb.GetEpochRequest) (*pb.Epoc
 	// supposed to create at least one revision on startup.
 	respEpoch := resp.GetMapRoot().GetMapRevision()
 	// Fetch log proofs.
-	logRoot, logConsistency, logInclusion, err := s.logProofs(ctx, domain, in.GetFirstTreeSize(), respEpoch)
+	logProof, err := s.logProofs(ctx, domain, in.GetFirstTreeSize(), respEpoch)
 	if err != nil {
 		return nil, err
 	}
 	return &pb.Epoch{
 		DomainId:       domain.DomainID,
 		Smr:            resp.GetMapRoot(),
-		LogRoot:        logRoot,
-		LogConsistency: logConsistency.GetHashes(),
-		LogInclusion:   logInclusion.GetHashes(),
+		LogRoot:        logProof.LogRoot,
+		LogConsistency: logProof.LogConsistency.GetHashes(),
+		LogInclusion:   logProof.LogInclusion.GetHashes(),
 	}, nil
 }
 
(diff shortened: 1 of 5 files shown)


core/keyserver/epochs.go
```diff
@@ -54,7 +56,7 @@ func (s *Server) GetEpoch(ctx context.Context, in *pb.GetEpochRequest) (*pb.Epoc
 		Revision: in.Epoch,
 	})
 	if err != nil {
-		glog.Warningf("GetSignedMapRootByRevision(%v, %v): %v", domain.MapID, in.Epoch, err)
+		glog.Errorf("GetEpoch(); GetSignedMapRootByRevision(%v, %v): %v", domain.MapID, in.Epoch, err)
```
Martin2112: ';' probably a typo.

core/keyserver/epochs.go
```diff
@@ -134,48 +136,78 @@ func (*Server) ListMutationsStream(in *pb.ListMutationsRequest, stream pb.KeyTra
...
+			LogId: d.LogID,
+			// SignedMapRoot must be in the log at MapRevision.
+			LeafIndex: epoch,
+			TreeSize:  secondTreeSize,
 		})
 	if err != nil {
-		glog.Errorf("tlog.GetLatestSignedLogRoot(%v): %v", logID, err)
-		return nil, nil, nil, status.Error(codes.Internal, "Cannot fetch SignedLogRoot")
+		glog.Errorf("logProofs(): log.GetInclusionProof(%v, %v, %v): %v", d.LogID, epoch, secondTreeSize, err)
+		return nil, status.Error(codes.Internal, "Cannot fetch log inclusion proof")
```
Martin2112: Might want to include err in what's returned but not sure.

core/keyserver/epochs.go
```diff
@@ -134,48 +136,78 @@ func (*Server) ListMutationsStream(in *pb.ListMutationsRequest, stream pb.KeyTra
...
+// latestLogRootProof returns the lastest SignedLogRoot and it's consistency proof.
+func (s *Server) latestLogRootProof(ctx context.Context, d *domain.Domain, firstTreeSize int64) (*tpb.SignedLogRoot, *tpb.Proof, error) {
+
+	sth, err := s.latestLogRoot(ctx, d)
+	if err != nil {
+		return nil, nil, err
 	}
-	secondTreeSize := logRoot.GetSignedLogRoot().GetTreeSize()
 	// Consistency proof.
+	secondTreeSize := sth.GetTreeSize()
```
Martin2112: Might want to short circuit if the secondTreeSize == firstTreeSize. If this can happen it should probably be tested.

core/keyserver/epochs.go
gdbelvin: yep

core/keyserver/epochs.go

core/keyserver/epochs.go
gdbelvin: It looks like Trillian does the short circuiting logic.  Not sure if it makes sense to duplicate that logic here?  The verifying client code come from the Trillian repo, so it should be tested.

Martin2112: Well you know the proof will be empty and it saves an RPC round trip. It's up to you if you think that's worth it.

PR #926, opened 2018-02-12
Cleanup the client

gdbelvin:
This PR tries to simplify the slightly convoluted layout of the Key Transparency client. 

- Collapse `core/client/kt` and `core/client/grpcc` into a simpler `core/client` package.
- Refactor map verification into a new `MapVerifier` 
- Only update `client.trusted` from `GetAndVerifyEntry`


gdbelvin:
PTAL

core/client/client.go
```diff
@@ -12,32 +12,27 @@
...
 
-	"github.com/google/keytransparency/core/client/kt"
 	"github.com/google/keytransparency/core/crypto/signatures"
-	"github.com/google/keytransparency/core/crypto/vrf"
 	"github.com/google/keytransparency/core/crypto/vrf/p256"
 	"github.com/google/keytransparency/core/mutator"
 	"github.com/google/keytransparency/core/mutator/entry"
 
 	"github.com/google/trillian"
 	"github.com/google/trillian/client"
```
RJPercival: It might be clearer to use an alias for this import, since it has the same name as this package. The name "tclient" seems to be used elsewhere.

core/client/verify.go
```diff
@@ -12,59 +12,56 @@
...
 		logVerifier: logVerifier,
 	}
 }
 
+// Index computes the index from a VRF proof.
+func (v *Verifier) Index(vrfProof []byte, domainID, appID, userID string) ([]byte, error) {
+	uid := vrf.UniqueID(userID, appID)
+	index, err := v.vrf.ProofToHash(uid, vrfProof)
+	if err != nil {
+		return nil, fmt.Errorf("vrf.ProofToHash(%v, %v): %v", appID, userID, err)
```
RJPercival: This error message makes it look like appID and userID were passed to ProofToHash(), when that was only indirectly the case.

core/client/verify.go
```diff
@@ -73,7 +70,7 @@ func New(vrf vrf.PublicKey,
 //  - Verify consistency proof from log.Root().
 //  - Verify inclusion proof.
 func (v *Verifier) VerifyGetEntryResponse(ctx context.Context, domainID, appID, userID string,
-	trusted *trillian.SignedLogRoot, in *pb.GetEntryResponse) error {
+	trusted trillian.SignedLogRoot, in *pb.GetEntryResponse) error {
```
RJPercival: "trusted" sounds like a bool.

core/client/verify.go
```diff
@@ -101,46 +98,36 @@ func (v *Verifier) VerifyGetEntryResponse(ctx context.Context, domainID, appID,
...
 		return ErrNilProof
 	}
+	leafProof.Leaf.Index = index[:]
 
-	leaf := leafProof.GetLeaf().GetLeafValue()
-	proof := leafProof.GetInclusion()
-	expectedRoot := in.GetSmr().GetRootHash()
-	mapID := in.GetSmr().GetMapId()
-	if err := merkle.VerifyMapInclusionProof(mapID, index[:], leaf, expectedRoot, proof, v.hasher); err != nil {
+	if err := v.mapVerifier.VerifyMapLeafInclusion(in.GetSmr(), leafProof); err != nil {
```
RJPercival: Go is pretty opinionated about capitalisation of initialisms - `GetSmr` should be `GetSMR`.

core/client/verify.go
RJPercival: Why are you taking a copy of trillian.SignedLogRoot?

core/client/client.go
```diff
@@ -12,32 +12,27 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package grpcc is a client for communicating with the Key Server.  It wraps
-// the gRPC apis in a rpc system neutral interface and verifies all responses.
-package grpcc
+// Package client is a client for communicating with the Key Server.
+// It wraps the gRPC apis and verifies all responses.
```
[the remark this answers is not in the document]
gdbelvin: done

core/client/client.go
gdbelvin: done

core/client/verify.go
gdbelvin: done

core/client/verify.go
gdbelvin: Because I don't want this function to be able to modify the argument.

core/client/verify.go
gdbelvin: `GetSmr` is unfortunately generated directly by protoc from "smr" in the proto file.

core/client/verify.go
RJPercival: This function *could* modify parts of the argument even with this change, since `trusted.RootHash` and `trusted.Signature` are pointers.

core/client/verify.go
```diff
@@ -12,59 +12,55 @@
...
 	"fmt"
-	"io/ioutil"
-	"log"
 
 	"github.com/google/keytransparency/core/crypto/commitments"
 	"github.com/google/keytransparency/core/crypto/vrf"
 	"github.com/google/keytransparency/core/mutator/entry"
-
 	"github.com/google/trillian"
 	"github.com/google/trillian/client"
```
RJPercival: This import should also have been aliased as "tclient".

core/client/verify_test.go
```diff
@@ -23,6 +23,7 @@ import (
 	"github.com/google/keytransparency/core/fake"
 
 	"github.com/google/trillian"
+	"github.com/google/trillian/client"
```
RJPercival: This import should also have been aliased as "tclient".

PR #929, opened 2018-02-21
Use real Trillian verifier in integration tests

gdbelvin:
This change catches a bug in the client which was not using the correct root for inclusion proof verification.  

This PR 
 - Uses a full KT client configured directly from the domain proto.
 - Sets and checks unique values for the public keys in user accounts.

### DIFF
--- a/core/client/client.go
+++ b/core/client/client.go
@@ -124,6 +124,13 @@ func New(ktClient pb.KeyTransparencyClient,
 	}
 }
 
+func (c *Client) updateTrusted(newTrusted *trillian.SignedLogRoot) {
+	if newTrusted.TimestampNanos > c.trusted.TimestampNanos &&
+		newTrusted.TreeSize >= c.trusted.TreeSize {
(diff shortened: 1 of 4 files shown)


core/client/client.go
```diff
@@ -124,6 +124,13 @@ func New(ktClient pb.KeyTransparencyClient,
 	}
 }
 
+func (c *Client) updateTrusted(newTrusted *trillian.SignedLogRoot) {
+	if newTrusted.TimestampNanos > c.trusted.TimestampNanos &&
+		newTrusted.TreeSize >= c.trusted.TreeSize {
+		c.trusted = *newTrusted
```
Martin2112: If you get a newer root with a smaller tree size that's possibly a bad thing that should maybe be exposed.

gdbelvin: Hmm, that plays into the server skew question from yesterday. 
If we get an older root than the currently trusted root, the consistency proof will fail.

PR #930, opened 2018-02-23
 Don't fail silently when uninitialized 

gdbelvin:
Now that log and map initializtion are part of the admin server, this
check and quick-fix are no longer needed. If by this point the map
really is not setup, a serious error has occurred.

This PR also contains a few formatting cleanups.

### DIFF
--- a/impl/integration/env.go
+++ b/impl/integration/env.go
@@ -167,11 +167,13 @@ func NewEnv() (*Env, error) {
 	seq := sequencer.New(logEnv.Log, mapEnv.Map, entry.New(), domainStorage, mutations, queue)
 	// Only sequence when explicitly asked with receiver.Flush()
 	d := &domaindef.Domain{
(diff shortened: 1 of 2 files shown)


impl/integration/env.go
```diff
@@ -167,11 +167,16 @@ func NewEnv() (*Env, error) {
 	seq := sequencer.New(logEnv.Log, mapEnv.Map, entry.New(), domainStorage, mutations, queue)
 	// Only sequence when explicitly asked with receiver.Flush()
 	d := &domaindef.Domain{
-		DomainID: domainPB.DomainId,
-		LogID:    domainPB.Log.TreeId,
-		MapID:    domainPB.Map.TreeId,
+		DomainID:    domainPB.DomainId,
```
Martin2112: This code looks familiar.

gdbelvin: It does :-)  - will be removed in next PR.

PR #931, opened 2018-02-23
Fix/seq wait

gdbelvin:
Diffbase: #930 

The Trillian Log API contract does NOT guarantee that items will be sequenced in the same order they are submitted. 

Eventually we will usethe AddSequencedLeavesAPI, but until that is available, we can refuse to send any more leaves to the log until it as sequenced the one we just sent it.

### DIFF
--- a/core/sequencer/sequencer.go
+++ b/core/sequencer/sequencer.go
@@ -17,7 +17,6 @@ package sequencer
 
 import (
 	"context"
-	"crypto/sha256"
 	"encoding/json"
 	"fmt"
 	"math"
@@ -33,6 +32,7 @@ import (
 
 	pb "github.com/google/keytransparency/core/api/v1/keytransparency_proto"
 	tpb "github.com/google/trillian"
+	tclient "github.com/google/trillian/client"
 )
 
 var (
@@ -70,6 +70,7 @@ func init() {
 type Sequencer struct {
 	domains     domain.Storage
 	tmap        tpb.TrillianMapClient
+	logAdmin    tpb.TrillianAdminClient
 	tlog        tpb.TrillianLogClient
 	mutatorFunc mutator.Func
 	mutations   mutator.MutationStorage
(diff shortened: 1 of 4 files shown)


core/sequencer/sequencer.go
```diff
@@ -297,22 +310,11 @@ func (s *Sequencer) createEpoch(ctx context.Context, d *domain.Domain, msgs []*m
...
-			LeafValue:        smrJSON,
-			LeafIdentityHash: idHash[:],
-		},
-	}); err != nil {
-		return fmt.Errorf("trillianLog.QueueLeaf(logID: %v, leaf: %v): %v",
-			logID, smrJSON, err)
-	}
-	return nil
+
+	return logClient.AddLeaf(ctx, smrJSON)
```
AlCutter: I'd add a comment here to be explicit that this call blocks until the leaf is actually integrated into the tree.

(Separately, might there be a more descriptive name we could change the function to?)

impl/integration/env.go
```diff
@@ -130,14 +130,19 @@ func NewEnv() (*Env, error) {
...
+	seq := sequencer.New(logEnv.Log, logEnv.Admin, mapEnv.Map, entry.New(), domainStorage, mutations, queue)
 	// Only sequence when explicitly asked with receiver.Flush()
 	d := &domaindef.Domain{
-		DomainID: domainPB.DomainId,
-		LogID:    domainPB.Log.TreeId,
-		MapID:    domainPB.Map.TreeId,
+		DomainID:    domainPB.DomainId,
+		LogID:       domainPB.Log.TreeId,
+		MapID:       domainPB.Map.TreeId,
+		MinInterval: 60 * time.Hour,
```
Martin2112: 60 hours?

AlCutter:
There are really about 4 different PRs in here (it seems like the commits are mostly unrelated), it would be much better to have multiple small separate PRs if possible.

impl/integration/env.go
gdbelvin: This supports only sequencing when `receiver.Flush` is manually called.

gdbelvin:
Added comments as requested.  Please review #930 upon which this PR is based first.

core/sequencer/sequencer.go
gdbelvin: Added comment

gdbelvin:
Rebased and ready for a second look

PR #932, opened 2018-02-23
 Set private keys directly during CreateDomain

gdbelvin:
This allows the test environment to use the same public keys repeatedly,
allowing tests to generate the same test vectors repeatedly.


core/adminserver/admin_server.go
```diff
@@ -167,12 +148,53 @@ func (s *Server) GetDomain(ctx context.Context, in *pb.GetDomainRequest) (*pb.Do
...
+// otherwise, it generates a new key with keygen.
+func privKeyOrGen(ctx context.Context, privKey *any.Any, keygen keys.ProtoGenerator) (proto.Message, error) {
+	if privKey != nil {
+		var keyProto ptypes.DynamicAny
+		if err := ptypes.UnmarshalAny(privKey, &keyProto); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal privatekey: %v", err)
+		}
+		return keyProto.Message, nil
+	}
+	return keygen(ctx, &keyspb.Specification{
```
Martin2112: Maybe pull out the keyspec as it's used in several places.

core/adminserver/admin_server.go
```diff
@@ -167,12 +148,53 @@ func (s *Server) GetDomain(ctx context.Context, in *pb.GetDomainRequest) (*pb.Do
...
+				},
+			},
+		}
+	}
+
+	config.Tree.Description = fmt.Sprintf("KT domain %s", domainID)
+	return &config
+}
+
 // CreateDomain reachs out to Trillian to produce new trees.
```
Martin2112: Existing typo.

core/adminserver/admin_server.go
gdbelvin: Done

PR #935, opened 2018-02-23
Clarify root of trust in monitor

gdbelvin:
This PR slightly reworks the monitor to clarify what the currently trusted log root is. 

It also reworks the integration test to be easier to follow.

### DIFF
--- a/cmd/keytransparency-monitor/main.go
+++ b/cmd/keytransparency-monitor/main.go
@@ -86,7 +86,7 @@ func main() {
 	store := fake.NewMonitorStorage()
 
 	// Create monitoring background process.
-	mon, err := monitor.NewFromConfig(ktClient, config, signer, store)
+	mon, err := monitor.NewFromDomain(ktClient, config, signer, store)
 	if err != nil {
 		glog.Exitf("Failed to initialize monitor: %v", err)
 	}
--- a/cmd/keytransparency-monitor/main.go
+++ b/cmd/keytransparency-monitor/main.go
@@ -26,6 +26,7 @@ import (
 	"github.com/google/keytransparency/core/fake"
 	"github.com/google/keytransparency/core/monitor"
 	"github.com/google/keytransparency/core/monitorserver"
+	"github.com/google/trillian"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 
 	"github.com/google/trillian/crypto"
@@ -89,7 +90,10 @@ func main() {
 	if err != nil {
 		glog.Exitf("Failed to initialize monitor: %v", err)
 	}
-	go mon.ProcessLoop(ctx, *domainID, store.LatestEpoch(), *pollPeriod)
+
+	// TODO(gbelvin): persist trusted roots
+	trusted := trillian.SignedLogRoot{}
+	go mon.ProcessLoop(ctx, *domainID, trusted, *pollPeriod)
 
 	// Monitor Server.
 	srv := monitorserver.New(store)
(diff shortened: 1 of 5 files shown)


core/integration/monitor_tests.go
```diff
@@ -43,89 +42,91 @@ amFdON6OhjYnBmJWe4fVnbxny0PfpkvXtg==
...
-	if err != nil {
-		t.Fatalf("Couldn't retrieve domain info: %v", err)
-	}
-	signer, err := pem.UnmarshalPrivateKey(monitorPrivKey, "")
+	privKey, err := pem.UnmarshalPrivateKey(monitorPrivKey, "")
 	if err != nil {
 		t.Fatalf("Couldn't create signer: %v", err)
 	}
-	mapVerifier, err := tclient.NewMapVerifierFromTree(resp.GetMap())
+	signer := crypto.NewSHA256Signer(privKey)
```
phad: move this down to the line preceding that declaring `store`

cmd/keytransparency-monitor/main.go
```diff
@@ -85,11 +86,14 @@ func main() {
...
-	mon, err := monitor.NewFromConfig(ktClient, config, signer, store)
+	mon, err := monitor.NewFromDomain(ktClient, config, signer, store)
 	if err != nil {
 		glog.Exitf("Failed to initialize monitor: %v", err)
 	}
-	go mon.ProcessLoop(ctx, *domainID, store.LatestEpoch(), *pollPeriod)
+
+	// TODO(gbelvin): persist trusted roots
+	trusted := trillian.SignedLogRoot{}
+	go mon.ProcessLoop(ctx, *domainID, trusted, *pollPeriod)
```
phad: what happened to `store.LatestEpoch()`?

core/monitor/verify.go
```diff
@@ -91,19 +91,15 @@ func (e *ErrList) Proto() []*statuspb.Status {
 }
 
 // VerifyEpoch verifies that epoch is correctly signed and included in the append only log.
```
phad: You should probably explain what the new param `trusted` is for, and why it might differ from `epoch.GetLogRoot()`

core/monitor/verify.go
```diff
@@ -91,19 +91,15 @@ func (e *ErrList) Proto() []*statuspb.Status {
...
 
-	if m.trusted == nil {
-		m.trusted = epoch.GetLogRoot()
-	}
-
-	if err := m.logVerifier.VerifyRoot(m.trusted, epoch.GetLogRoot(), epoch.GetLogConsistency()); err != nil {
+	if err := m.logVerifier.VerifyRoot(trusted, epoch.GetLogRoot(), epoch.GetLogConsistency()); err != nil {
 		// this could be one of ErrInvalidLogSignature, ErrInvalidLogConsistencyProof
-		errs.AppendStatus(status.Newf(codes.DataLoss, "VerifyRoot: %v", err).WithDetails(m.trusted, epoch))
+		errs.AppendStatus(status.Newf(codes.DataLoss, "VerifyRoot: %v", err).WithDetails(epoch, trusted))
```
phad: Did you intend to transpose `m.trusted` and `epoch` here?

core/integration/monitor_tests.go
gdbelvin: done

cmd/keytransparency-monitor/main.go
gdbelvin: `store.LatestEpoch` was only a number.  For consistency proofs, we need to store the full STH.  I'll leave that for a future PR, however, as the monitor is not on the front burner right now.

core/monitor/verify.go
gdbelvin: done

core/monitor/verify.go
gdbelvin: I did. thanks for catching that.

PR #939, opened 2018-02-27
Make Timeout a test env variable.

gdbelvin:
Different test environments will run at different speeds.
Therefore it makes sense to make timeout a part of Env.

This PR also expands the use of timeouts to several more
functions which could timeout.


gdbelvin:
This PR raises a bit of a question about what the best place to implement timeouts since they can occur on virtually every RPC.

Issue #944, opened 2018-03-03
Fix: Find leaking transaction


gdbelvin:

gdbelvin:
Can't reproduce

PR #946, opened 2018-03-03
Add loadtest hammer operation to the client

gdbelvin:
Load test the open source continuous integration server. 
Each operation takes ~2s to complete end to end.
QPS peaked at ~80 and revealed a leaking transaction in the Trillian Map: #944 

screen shot 2018-03-03 at 13 26 27

### DIFF
--- a/cmd/keytransparency-client/cmd/hammer.go
+++ b/cmd/keytransparency-client/cmd/hammer.go
@@ -43,9 +43,11 @@ var (
 )
 
 func init() {
-	log.SetOutput(memLog)
-
+	// Silence "logging before flag.Parse"
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
+	flag.CommandLine.Parse([]string{})
+
+	log.SetOutput(memLog)
 
 	RootCmd.AddCommand(hammerCmd)
 
@@ -168,18 +170,19 @@ func draw(latencies []float64, data *tm.DataTable, qps int64) {
 	chart := tm.NewLineChart(100, 20)
 	tm.Println(chart.Draw(data))
 
-	// Latency Hist
-	tm.Printf("Latency Histogram:\n")
-	width := 50
-	height := 10
-
-	hist := histogram.Hist(height, latencies)
-	histogram.Fprint(tm.Output, hist, histogram.Linear(width))
-
 	// Console output
+	tm.Printf("Debug Output:\n")
 	box := tm.NewBox(180, 10, 0)
 	box.Write(memLog.Bytes())
 	tm.Println(box)
+
+	// Latency Hist
+	width := 50
+	height := 10
+
+	tm.Printf("Latency Histogram:\n")
+	hist := histogram.Hist(height, latencies)
+	histogram.Fprint(tm.Output, hist, histogram.Linear(width))
 
 	tm.Flush()
 }
--- a/cmd/keytransparency-client/cmd/hammer.go
+++ b/cmd/keytransparency-client/cmd/hammer.go
@@ -23,11 +23,17 @@ import (
 	"sync"
 	"time"
 
+	"github.com/google/keytransparency/core/authentication"
+
 	"github.com/aybabtme/uniplot/histogram"
+	"github.com/golang/glog"
 	"github.com/paulbellamy/ratecounter"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
 
 	tm "github.com/buger/goterm"
+	tpb "github.com/google/keytransparency/core/api/type/type_proto"
 )
 
 var (
@@ -45,6 +51,7 @@ func init() {
 	RootCmd.AddCommand(hammerCmd)
 
 	hammerCmd.Flags().IntVar(&maxWorkers, "workers", 1, "Number of parallel workers")
+	hammerCmd.Flags().DurationVar(&ramp, "ramp", 1*time.Second, "Time to spend ramping up")
 }
 
 // hammerCmd represents the post command
@@ -64,10 +71,6 @@ var hammerCmd = &cobra.Command{
 		return nil
 	},
 }
-
-var (
-	qps = make([][2]float64, 0, 1000)
-)
 
 func parallel(ctx context.Context, maxWorkers int) {
 	times := make(chan time.Duration)
@@ -90,12 +93,22 @@ func parallel(ctx context.Context, maxWorkers int) {
 type job func() error
 
 func generateJobs(ctx context.Context, jobs chan<- job) {
+	i := 0
 	for {
 		select {
 		case <-ctx.Done():
 			return
 		case jobs <- bindJob(ctx, i):
 			i++
+		}
+	}
+}
+
+func bindJob(ctx context.Context, i int) func() {
+	userID := fmt.Sprintf("user_%v", i)
+	return func() {
+		if err := writeOp(ctx, "app1", userID); err != nil {
+			glog.Errorf("write(%v): %v", userID, err)
 		}
 	}
 }
@@ -160,7 +173,7 @@ func draw(latencies []float64, data *tm.DataTable, qps int64) {
 	tm.Printf("Total Requests: %v\n", len(latencies))
 	tm.Printf("Recent QPS:    %v\n", qps)
 
-	// QPS Chart
+	// Threads per QPS graph
 	tm.Printf("QPS Chart:\n")
 	chart := tm.NewLineChart(100, 20)
 	tm.Println(chart.Draw(data))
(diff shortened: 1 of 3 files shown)


cmd/keytransparency-client/cmd/hammer.go
```diff
@@ -0,0 +1,219 @@
...
+	"sync"
+	"time"
+
+	"github.com/google/keytransparency/core/authentication"
+
+	"github.com/aybabtme/uniplot/histogram"
+	"github.com/paulbellamy/ratecounter"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
```
Martin2112: Are these new dependencies or is KT already using them?

cmd/keytransparency-client/cmd/hammer.go
```diff
@@ -0,0 +1,219 @@
...
+	RootCmd.AddCommand(hammerCmd)
+
+	hammerCmd.Flags().IntVar(&maxWorkers, "workers", 1, "Number of parallel workers")
+	hammerCmd.Flags().DurationVar(&ramp, "ramp", 1*time.Second, "Time to spend ramping up")
+}
+
+// hammerCmd represents the post command
+var hammerCmd = &cobra.Command{
+	Use:   "hammer",
+	Short: "Loadtest the server",
```
Martin2112: What server?

cmd/keytransparency-client/cmd/hammer.go
```diff
@@ -0,0 +1,219 @@
...
+		}
+	},
+	RunE: func(cmd *cobra.Command, args []string) error {
+		ctx := context.Background()
+		parallel(ctx, maxWorkers)
+		return nil
+	},
+}
+
+func parallel(ctx context.Context, maxWorkers int) {
```
Martin2112: Very generic name.

cmd/keytransparency-client/cmd/hammer.go
```diff
@@ -0,0 +1,219 @@
...
+	times := make(chan time.Duration)
+	jobs := make(chan job)
+	var wg sync.WaitGroup
+	go recordLatencies(ctx, times)
+	go generateJobs(ctx, jobs)
+
+	// Slowly add workers up to maxWorkers
+	ramp := time.NewTicker(time.Duration(ramp.Nanoseconds() / int64(maxWorkers)))
+	for ; workers < maxWorkers && ctx.Err() == nil; <-ramp.C {
```
Martin2112: I think this might break if someone sets maxWorkers flag negative.

cmd/keytransparency-client/cmd/hammer.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+func bindJob(ctx context.Context, i int) func() error {
+	userID := fmt.Sprintf("user_%v", i)
+	return func() error {
+		return writeOp(ctx, "app1", userID)
+	}
+}
+
+func runJobsInThread(ctx context.Context, id int, jobs <-chan job, times chan<- time.Duration, wg *sync.WaitGroup) {
```
Martin2112: I don't think thread is a Go term. Maybe just runJobs or runJobsParallel.

cmd/keytransparency-client/cmd/hammer.go
```diff
@@ -0,0 +1,219 @@
...
+				log.Printf("f(): %v", err)
+				continue
+			}
+			times <- time.Since(start)
+		}
+	}
+}
+
+func recordLatencies(ctx context.Context, times <-chan time.Duration) {
+	//start := time.Now()
```
Martin2112: Delete commented out code.

cmd/keytransparency-client/cmd/hammer.go
```diff
@@ -0,0 +1,219 @@
...
+	latencies := make([]float64, 0, 1000)
+	refresh := time.NewTicker(250 * time.Millisecond)
+	newworker := time.NewTicker(time.Duration(ramp.Nanoseconds() / int64(maxWorkers)))
```
Martin2112: Just worker? If not then newWorker.

cmd/keytransparency-client/cmd/hammer.go
```diff
@@ -0,0 +1,219 @@
...
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
+	userCreds := authentication.GetFakeCredential(userID)
+	c, err := GetClient(ctx, userCreds)
+	if err != nil {
+		return fmt.Errorf("error connecting: %v", err)
+	}
+
+	// Update.
```
Martin2112: Comment is not adding much value. Suggest either expanding or removing it.

cmd/keytransparency-client/cmd/root.go
```diff
@@ -76,11 +76,11 @@ func init() {
 	cobra.OnInitialize(initConfig)
 	RootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.keytransparency.yaml)")
 
-	RootCmd.PersistentFlags().String("domain", "google.com", "Domain within the KT server")
-	RootCmd.PersistentFlags().String("kt-url", "35.184.134.53:8080", "URL of Key Transparency server")
+	RootCmd.PersistentFlags().String("domain", "default", "Domain within the KT server")
+	RootCmd.PersistentFlags().String("kt-url", "35.225.194.168:8080", "URL of Key Transparency server")
```
Martin2112: Does it make sense to hardcode an IP address?

cmd/keytransparency-client/cmd/hammer.go
gdbelvin: Key Transparency is already using the `spf13` dependencies.  The `ratecounter` and `histogram` dependencies are new. I should also point out that we are not importing the `cmd` directory - only the `core` directory.

cmd/keytransparency-client/cmd/hammer.go
gdbelvin: The server target and other authentication variables are defined in `root.go`

cmd/keytransparency-client/cmd/hammer.go
gdbelvin: renamed to `runHammer`

cmd/keytransparency-client/cmd/hammer.go
gdbelvin: I'll make it uint

cmd/keytransparency-client/cmd/hammer.go
gdbelvin: how's `worker`?

cmd/keytransparency-client/cmd/hammer.go
gdbelvin: done

cmd/keytransparency-client/cmd/hammer.go
Martin2112: Ack.

cmd/keytransparency-client/cmd/hammer.go
Martin2112: It looked like this short message could be shown to the user and I was just questioning whether it was useful.

cmd/keytransparency-client/cmd/hammer.go
Martin2112: SGTM.

cmd/keytransparency-client/cmd/hammer.go
gdbelvin: `newWorker` because it ticks at about the same time a new worker is created.

cmd/keytransparency-client/cmd/hammer.go
gdbelvin: done

cmd/keytransparency-client/cmd/root.go
gdbelvin: It helps everything "just" work, for now.

cmd/keytransparency-client/cmd/hammer.go
gdbelvin: It's not the most useful.  I'll add the following longer description: 
```
Sends update requests for user_1 through user_n using a select number of workers in parallel
```

gdbelvin:
Ready for a second look

cmd/keytransparency-client/cmd/root.go
Martin2112: Default addresses for load tests can be risky. I've seen outages caused by this, though on GCE it's probably not as dangerous. e.g. if you'd put some internal network address and someone downloads it who happens to have some server running on that address.

PR #948, opened 2018-03-14
Use types.LogRootV1

gdbelvin:
Following google/trillian#1048, use types.LogRootV1 as the local
reference of what the most up-to-date log root is.

### DIFF
--- a/core/adminserver/admin_server.go
+++ b/core/adminserver/admin_server.go
@@ -277,7 +277,7 @@ func (s *Server) initialize(ctx context.Context, logTree, mapTree *tpb.Tree) err
 
 	// If the tree is empty and the map is empty,
 	// add the empty map root to the log.
-	if logRoot.GetTreeSize() != 0 ||
+	if logRoot.TreeSize != 0 ||
 		mapRoot.GetMapRoot().GetMapRevision() != 0 {
 		return nil // Init not needed.
 	}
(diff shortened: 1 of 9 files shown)


core/adminserver/admin_server.go
```diff
@@ -277,7 +277,7 @@ func (s *Server) initialize(ctx context.Context, logTree, mapTree *tpb.Tree) err
 
 	// If the tree is empty and the map is empty,
 	// add the empty map root to the log.
-	if logRoot.GetTreeSize() != 0 ||
+	if logRoot.TreeSize != 0 ||
```
cesarghali: Doesn't using getters protect from a nil object?

core/client/client.go
```diff
@@ -235,7 +247,7 @@ func (c *Client) Update(ctx context.Context, u *tpb.User, signers []signatures.S
 
 // QueueMutation signs an entry.Mutation and sends it to the server.
 func (c *Client) QueueMutation(ctx context.Context, m *entry.Mutation, signers []signatures.Signer) error {
-	req, err := m.SerializeAndSign(signers, c.trusted.GetTreeSize())
+	req, err := m.SerializeAndSign(signers, int64(c.trusted.TreeSize))
```
cesarghali: Same comment about getters here.

core/adminserver/admin_server.go
gdbelvin: logRoot used to be a proto. Now it is a go struct that is verified.

core/client/client.go
gdbelvin: see above.

PR #957, opened 2018-04-13
Refactor hammer as a library

gdbelvin:
This PR allows hammer operations to be used in different contexts.
It also slightly simplifies the channel logic that was used in the hammer.

### DIFF
--- a/core/client/hammer/hammer.go
+++ b/core/client/hammer/hammer.go
@@ -18,10 +18,10 @@ package hammer
 import (
 	"context"
 	"fmt"
+	"log"
 	"sync"
 	"time"
 
-	"github.com/golang/glog"
 	"github.com/google/tink/go/tink"
 	"google.golang.org/grpc"
 
@@ -60,6 +60,8 @@ func New(ctx context.Context, ktAddr, domainID string, timeout time.Duration, ke
 	if err != nil {
 		return nil, err
 	}
+
+	log.Printf("config: %#v", config)
 
 	authorizedKeys, err := keyset.GetPublicKeysetHandle()
 	if err != nil {
@@ -137,7 +139,7 @@ func (h *Hammer) printStats(period time.Duration, outputs <-chan error) {
 			select {
 			case <-refresh.C:
 				rate := totalRequests / time.Since(startTime).Seconds()
-				glog.Infof("QPS: %v, \tRequests: %v, \tWorkers: %v", rate, totalRequests, h.workers)
+				log.Printf("QPS: %v, \tRequests: %v, \tWorkers: %v", rate, totalRequests, h.workers)
 			case <-stop:
 				refresh.Stop()
 				return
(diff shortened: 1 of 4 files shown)


core/client/hammer/hammer.go
```diff
@@ -0,0 +1,190 @@
...
+		appID:   fmt.Sprintf("hammer_%v", time.Now().Format("2006-01-02/15:04:05")),
+		config:  config,
+
+		signers:        []*tink.KeysetHandle{keyset},
+		authorizedKeys: authorizedKeys.Keyset(),
+	}, nil
+}
+
+// Run runs operationCount with up to maxWorkers over ramp duration.
+func (h *Hammer) Run(ctx context.Context, operationCount, maxWorkers int, ramp time.Duration) {
```
pavelkalinnikov: Maybe make this whole thing (down to `worker`) `ctx`-cancellable?

core/client/hammer/hammer.go
```diff
@@ -0,0 +1,190 @@
...
+	go func() {
+		for i := 0; i < count; i++ {
+			select {
+			case <-ctx.Done():
+				close(jobs)
+				return
+			case jobs <- i:
+			}
+		}
+		close(jobs)
```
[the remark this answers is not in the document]
gdbelvin: Good idea

core/client/hammer/hammer.go
```diff
@@ -0,0 +1,190 @@
...
+			rate := totalRequests / time.Since(startTime).Seconds()
+			glog.Infof("QPS: %v, \tRequests: %v, \tWorkers: %v", rate, totalRequests, h.workers)
+		}
+	}()
+
+	// Collect outputs until output closes.
+	for range outputs {
+		totalRequests++
+	}
+	refresh.Stop()
```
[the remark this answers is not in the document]
gdbelvin: Thanks for catching that

core/client/hammer/hammer.go
```diff
@@ -0,0 +1,196 @@
...
+	// Print stats asyncronously until refresh closes.
+	refresh := time.NewTicker(period)
+	stop := make(chan struct{})
+	go func() {
+		for {
+			select {
+			case <-refresh.C:
+				rate := totalRequests / time.Since(startTime).Seconds()
+				glog.Infof("QPS: %v, \tRequests: %v, \tWorkers: %v", rate, totalRequests, h.workers)
+			case <-stop:
```
pavelkalinnikov: Shouldn't you write to `stop` channel to make this `case` happen?

core/client/hammer/hammer.go
pavelkalinnikov: You use `case <-ctx.Done():` clause below in `gen` function to make it early cancellable, right? I was suggesting to maybe do the same in `launchWorkers` and `printStats`. But after second thought I see that those will actually terminate after `gen` gets cancelled, as they both depend on it through channels.
So you can ignore this comment.

core/client/hammer/hammer.go
gdbelvin: `close(stop)` causes the channel return immediately

Issue #912, opened 2018-01-25
Replace ObjectHash


gdbelvin:
- Define a canonical encoding of SignedMapHead using TLS encoding
   - This matches CT's STH in style. 
   - Removes the need to reimplement ObjectHash in JavaScript and other languages.

PR #967, opened 2018-04-27
Use RFC6962 log hashing strategy for log of map roots

gdbelvin:
Now that MapRoots are signed `byte[]` objects, we have a proper
bytes string to place in the Trillian Log where previously we only
had an abstract object that needed objecthash.

As a result, we can use a simpler SHA256 based hash strategy defined in RFC6962 rather
than object hash for placing things in the log.

This PR also registers the hashers in the client verifier where they are used in `NewVerifierFromDomain`.

#912


core/testdata/domain.json
```diff
@@ -1,10 +1,10 @@
 {
 	"domain_id": "domain_5577006791947779410",
 	"log": {
-		"tree_id": 4013249521620300357,
+		"tree_id": 2029501983648168255,
 		"tree_state": 1,
 		"tree_type": 1,
-		"hash_strategy": 3,
+		"hash_strategy": 1,
```
RJPercival: Are these JSON files marshaled protobufs? If so, using the enum names instead of their values would be nicer.

core/adminserver/admin_server.go
```diff
@@ -45,7 +44,7 @@ var (
 			DisplayName:        "KT SMH Log",
 			TreeState:          tpb.TreeState_ACTIVE,
 			TreeType:           tpb.TreeType_LOG,
-			HashStrategy:       tpb.HashStrategy_OBJECT_RFC6962_SHA256,
+			HashStrategy:       tpb.HashStrategy_RFC6962_SHA256,
```
RJPercival: Might be helpful to document that the verifier needs to import whichever hash algorithms you choose to use here.

core/testdata/domain.json
gdbelvin: Good idea. Done.

Note that this requires using `jsonpb` to do the encoding rather than `encoding/json`.

core/adminserver/admin_server.go
gdbelvin: Done

RJPercival:
Is CI broken? I've seen a couple of builds on this PR timeout.

gdbelvin:
The reason I can't use `jsonpb` for both is because the second data structure is a slice of Go Objects: `[]testdata.GetEntryResponseVector`, rather than a protobuf object.

The build timeouts are odd...

gdbelvin:
I diagnosed the test timeouts as some of the hashers not being registered when the tests were spinning up Trillian servers in-process.

PR #968, opened 2018-05-01
Add UnitTest for PaginateHistory

gdbelvin:
This PR adds a long overdue unit test and unittesting framework for client operations. 

Included:
* A mock KeyTransparency server
* An interface for mocking out the verifier
* Removed a global variable `pageSize`. It's now a client variable.


core/client/client.go
```diff
@@ -74,6 +69,19 @@ var (
 	Vlog = log.New(ioutil.Discard, "", 0)
 )
 
+// Verifier is used to verify specific outputs from Key Transparency.
+type Verifier interface {
+	// Index computes the index from a VRF proof.
```
pavelkalinnikov: What is VRF?

gdbelvin: VRF means [Verifiable Random Function](https://en.wikipedia.org/wiki/Verifiable_random_function) 

KT has a VRF implementation in `core/crypto/vrf/`

core/testutil/mockserver.go
```diff
@@ -0,0 +1,67 @@
...
+
+// MockKT runs a mock KeyTransparency server.
+type MockKT struct {
+	Server *ktmock.MockKeyTransparencyServer
+	Client pb.KeyTransparencyClient
+	Addr   string
+}
+
+// NewMockKT returns a new mock Key Transparency server listening on a random port .
+// Returns the started server, the listener it's using for connection and a
```
pavelkalinnikov: The pure `Listener` object is not there, only the server wrapper.
`Returns the started server, a client connected to it, and a close ...`

core/client/verify.go
```diff
@@ -38,26 +38,26 @@ var (
 	ErrNilProof = errors.New("nil proof")
 )
 
-// Verifier is a client helper library for verifying request and responses.
-type Verifier struct {
-	vrf         vrf.PublicKey
-	mapVerifier *tclient.MapVerifier
-	logVerifier *tclient.LogVerifier
+// Verify is a client helper library for verifying request and responses.
+type Verify struct {
```
pavelkalinnikov: Is there a way to name this with a noun? Some ideas:
- Call this `Verifier` and the interface - `IVerifier`. Not sure whether Go naming conventions allows this.
- `Verification(Client)`, `VerifierImpl`, `VerifiedClient` ...

core/client/client_test.go
```diff
@@ -104,4 +112,137 @@ func TestCompressHistory(t *testing.T) {
...
+						Values:    r.items,
+					}, nil)
+			}
+
+			if _, _, err = c.PaginateHistory(ctx, appID, userID, tc.start, tc.end); err != tc.wantErr {
+				t.Errorf("PaginateHistory(): %v, want %v", err, tc.wantErr)
+			}
+		})
+	}
+
```
pavelkalinnikov: nit: drop empty line?

core/client/client_test.go
```diff
@@ -104,4 +112,137 @@ func TestCompressHistory(t *testing.T) {
...
+			}
+
+			for _, r := range tc.reqs {
+				s.Server.EXPECT().ListEntryHistory(gomock.Any(), matchers.ProtoEqual(&pb.ListEntryHistoryRequest{
+					AppId:    appID,
+					UserId:   userID,
+					Start:    r.wantStart,
+					PageSize: r.wantSize,
+				})).
+					Return(&pb.ListEntryHistoryResponse{
```
pavelkalinnikov: nit: maybe merge this line with the previous one?

core/client/client.go
```diff
@@ -74,6 +69,19 @@ var (
...
+	Index(vrfProof []byte, domainID, appID, userID string) ([]byte, error)
+	// VerifyGetEntryResponse verifies everything about a GetEntryResponse:
```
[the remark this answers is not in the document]
gdbelvin: done

core/client/client.go
```diff
@@ -151,17 +164,18 @@ func min(x, y int32) int32 {
...
 	if start < 0 {
 		return nil, nil, fmt.Errorf("start=%v, want >= 0", start)
 	}
 	allRoots := make(map[uint64]*types.MapRootV1)
 	allProfiles := make(map[uint64][]byte)
-	epochsReceived := int64(0)
-	epochsWant := end - start + 1
-	for epochsReceived < epochsWant {
-		count := min(int32((end-start)+1), pageSize)
+	epochsWant := int(end - start + 1)
```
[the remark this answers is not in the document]
gdbelvin: Good idea. It produces a slightly nasty set of casts below, but it's probably safer.

core/testutil/mockserver.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+// MockKT runs a mock KeyTransparency server.
+type MockKT struct {
+	Server *ktmock.MockKeyTransparencyServer
+	Client pb.KeyTransparencyClient
+	Addr   string
+}
+
+// NewMockKT returns a new mock Key Transparency server listening on a random port .
```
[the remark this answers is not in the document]
gdbelvin: done

core/testutil/mockserver.go
```diff
@@ -0,0 +1,67 @@
...
+// Returns the started server, the listener it's using for connection and a
+// close function that must be defer-called on the scope the server is meant to
```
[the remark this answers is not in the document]
gdbelvin: yep

core/client/verify.go
gdbelvin: I struggled with the naming conventions too. 
is `RealVerifier` appropriate?  The interface exists so we can have a `FakeVerifier`

core/client/client_test.go
gdbelvin: That does look better.

core/client/client.go
```diff
@@ -151,17 +164,18 @@ func min(x, y int32) int32 {
...
 	}
 	allRoots := make(map[uint64]*types.MapRootV1)
 	allProfiles := make(map[uint64][]byte)
-	epochsReceived := int64(0)
 	epochsWant := end - start + 1
-	for epochsReceived < epochsWant {
-		count := min(int32((end-start)+1), pageSize)
+	for int64(len(allProfiles)) < epochsWant {
+		log.Printf("pageSize: %v", c.pageSize)
+		count := min(int32(epochsWant-int64(len(allProfiles))), c.pageSize)
```
pavelkalinnikov: Hm, this bit is just as problematic as the previous version due to shrinking cast. Probably making a `min` function for `int64` is too much hassle. Instead, how about rewinding this back, and simply checking `end-start+1` for an overflow before casting it down to `int32` in line 175?

gdbelvin: Making `min` accept `int64` is actually the better solution. It defers the downcast till the end when we expect the value to be small.

pavelkalinnikov: Oh, I agree. Wasn't aware that the `min` function was used only here.

pavelkalinnikov:
This now LGTM.
@taknira do you want to do a final proofread?

One general question. I noticed you use both `Vlog` and `glog.V` for verbose logging. I wonder why you don't use just `glog`.

core/client/verify.go
```diff
@@ -38,26 +38,26 @@ var (
 	ErrNilProof = errors.New("nil proof")
 )
 
-// Verifier is a client helper library for verifying request and responses.
-type Verifier struct {
-	vrf         vrf.PublicKey
-	mapVerifier *tclient.MapVerifier
-	logVerifier *tclient.LogVerifier
+// RealVerifier is a client helper library for verifying request and responses.
```
pavelkalinnikov: Maybe you could make a note that this implements `Verifier`? That would explain the seemingly unclear `Real` bit.

gdbelvin: Done

gdbelvin:
`Vlog` provides visual logging on the command line during interactive operations. 
It is separate from `glog` which provides debugging and general info.

Kat, would you mind taking a look?

core/client/client.go
taknira: Index computes the index of what?  Add to the comment to make it really clear :)

core/client/client.go
```diff
@@ -74,6 +69,19 @@ var (
...
 )
 
+// Verifier is used to verify specific outputs from Key Transparency.
+type Verifier interface {
+	// Index computes the index from a VRF proof.
+	Index(vrfProof []byte, domainID, appID, userID string) ([]byte, error)
+	// VerifyGetEntryResponse verifies everything about a GetEntryResponse.
+	VerifyGetEntryResponse(ctx context.Context, domainID, appID, userID string, trusted types.LogRootV1, in *pb.GetEntryResponse) (*types.MapRootV1, *types.LogRootV1, error)
+	// VerifyEpoch verifies that epoch is correctly signed and included in the append only log.
+	// VerifyEpoch also verifies that epoch.LogRoot is consistent with the last trusted SignedLogRoot.
```
taknira: Did the *pb.Epoch parameter used to be called 'epoch' or something?  Maybe capitalize in these comments to make it clear you're talking about the type rather than the variable name - 'VerifyEpoch verifies that the Epoch is correctly...', 'VerifyEpoch also verifies that Epoch.LogRoot is ....'

core/client/client.go
gdbelvin: Index computes the index of an appID, userID pair from a VRF proof, obtained from the server.

core/client/client.go
gdbelvin: I'll just rename `in` to `epoch` :-)

core/client/client_test.go
```diff
@@ -104,4 +112,135 @@ func TestCompressHistory(t *testing.T) {
...
+					UserId:   userID,
+					Start:    r.wantStart,
+					PageSize: r.wantSize,
+				})).Return(&pb.ListEntryHistoryResponse{
+					NextStart: r.next,
+					Values:    r.items,
+				}, nil)
+			}
+
+			if _, _, err = c.PaginateHistory(ctx, appID, userID, tc.start, tc.end); err != tc.wantErr {
```
taknira: This method of testing PaginateHistory seems odd, given you could test based on state - i.e. just test PaginateHistory based on its inputs and outputs, rather than checking that it's calling the correct server endpoints with the right arguments....  In fact, by doing the way you currently are explicitly *doesn't* test some of the logic within PaginateHistory outside of what it sends/receives to/from the server.

In addition to this, (and as discussed separately from this PR,) use of gomock (and mocks in general) tends to be discouraged from a readability standpoint, preferring real objects or fakes.

If you were to test PaginateHistory based on inputs and outputs (and compare them to expected outputs) you could just fake the one rpc it calls (ListEntryHistory is it?), which would remove the need for mocking, making the test code much simpler.  (Although incidentally it looks like the PaginateHistory code doesn't call that directly, but instead calls client.VerifiedListHistory, which then calls ListEntryHistory - I couldn't find a test for VerifiedListHistory, so using this method of testing could also be appropriate for that function too.)

This method of testing would be specifically testing the PaginateHistory code, and assuming the underlying functions that it calls work as expected.

gdbelvin:
Yay! the mocks have been removed.  Thanks @taknira for helping keep up the code excellence. 
PTAL.

core/client/client_test.go
```diff
@@ -118,113 +119,155 @@ func TestPaginateHistory(t *testing.T) {
...
+			6:  {Smr: &trillian.SignedMapRoot{MapRoot: []byte{6}}},
+			7:  {Smr: &trillian.SignedMapRoot{MapRoot: []byte{7}}},
+			8:  {Smr: &trillian.SignedMapRoot{MapRoot: []byte{8}}},
+			9:  {Smr: &trillian.SignedMapRoot{MapRoot: []byte{9}}},
+			10: {Smr: &trillian.SignedMapRoot{MapRoot: []byte{10}}},
+		},
+	}
+	s, stop, err := testutil.NewFakeKT(srv)
+	if err != nil {
+		t.Fatalf("NewMockKT(): %v", err)
```
taknira: It's not a mock anymore :P

core/client/client_test.go
```diff
@@ -118,113 +119,155 @@ func TestPaginateHistory(t *testing.T) {
...
-			}
-
-			if _, _, err = c.PaginateHistory(ctx, appID, userID, tc.start, tc.end); err != tc.wantErr {
+			_, values, err := c.PaginateHistory(ctx, appID, userID, tc.start, tc.end)
+			if err != tc.wantErr {
 				t.Errorf("PaginateHistory(): %v, want %v", err, tc.wantErr)
 			}
+			if got, want := values, tc.wantValues; !reflect.DeepEqual(got, want) {
+				t.Errorf("PaginateHistory().values: \n%#v, want \n%#v, diff: \n%v",
+					got, want, pretty.Compare(got, want))
```
taknira: (Optional)  Is there *any* chance at all that pretty.Compare and reflect.DeepEqual would return different results?  Might be worth using pretty.Compare for the initial comparison too, and checking the diff is empty of something?  Completely up to you though, as the two funcs will likely be in agreement the majority of the time.  Just a musing :)

PR #971, opened 2018-05-06
Missing masterPassword input in post.go

weikengchen:
Just start to try Key Transparency (just work-in-progress? I can't wait to use it)

I find that when I want to post a public key, there will be a notice "please enter a master key", but post.go does not accept the password `-p` as input.


weikengchen:
A side question: is Key Transparency ready to an extent that a class project can launch a server and create a local key Merkle tree?

gdbelvin:
Hi @weikengchen, thanks for spotting and fixing this! 

Key Transparency is ready for prototyping and classroom use.  
Key Transparency is in the middle of polishing and tuning so it can be used in production.

PR #973, opened 2018-05-10
Move authentication to grpc interceptor

gdbelvin:
This PR supports greater flexibility for deployment specific authentication schemes by moving authentication from the `core/` code base to an `impl/` specific grpc interceptor which can easily be swapped out with more appropriate authentication logic in specific deployments. 

Changes:
- `core/authentication` moved to `impl/authentication`
-  `ValidateCreds` moved to `AuthFunc` to match the `grpc_auth.AuthFunc` interface. 
    -  `SecurityContext` moved to `ValidatedSecurity` and is now stored in `context.Context`
- `impl/google/authentication` merged with `impl/authentication` to use the new `ValidatedSecurity`
-  Removed `auth` from `core/keyserver` :smile:
-  Wrote a *simple* auth interceptor to support per-method authentication configurations.


impl/authentication/google.go
```diff
@@ -58,13 +57,17 @@ func NewGoogleAuth() (*GAuth, error) {
...
 	}
-	return &GAuth{googleService}, nil
+	return &GAuth{service: googleService}, nil
 }
 
-// ValidateCreds authenticate the information present in ctx.
-func (a *GAuth) ValidateCreds(ctx context.Context) (*authentication.SecurityContext, error) {
-	// Get Tokeninfo from credentials.
-	tokenInfo, err := a.validateToken(ctx)
+// AuthFunc authenticate the information present in ctx.
```
AlCutter: authenticates

impl/authentication/interceptor.go
```diff
@@ -0,0 +1,63 @@
...
+	"context"
+
+	"google.golang.org/grpc"
+
+	"github.com/golang/glog"
+	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
+	grpc_auth "github.com/grpc-ecosystem/go-grpc-middleware/auth"
+)
+
+// UnaryServerInterceptor returns a new unary server interceptors that performs per-request auth.
```
AlCutter: "interceptor that"

impl/authentication/interceptor.go
```diff
@@ -0,0 +1,63 @@
...
+func UnaryServerInterceptor(authFuncs map[string]grpc_auth.AuthFunc) grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		glog.Infof("In authentication interceptor")
```
AlCutter: Should this be a `V(...).Infof`?

impl/authentication/interceptor.go
```diff
@@ -0,0 +1,63 @@
...
+		authFunc, ok := authFuncs[info.FullMethod]
+		if !ok {
+			glog.Infof("auth interceptor: no hander for %v", info.FullMethod)
```
AlCutter: Warning / V?

impl/authentication/interceptor.go
```diff
@@ -0,0 +1,63 @@
...
+		}
+		newCtx, err := authFunc(ctx)
+		if err != nil {
+			return nil, err
+		}
+		return handler(newCtx, req)
+	}
+}
+
+// StreamServerInterceptor returns a new unary server interceptors that performs per-request auth.
```
AlCutter: "interceptor that"

impl/authorization/authorization.go
```diff
@@ -38,30 +42,37 @@ func New() authorization.Authorization {
 // authorized to carry the given permission. A call is authorized if:
 //  1. userID matches the identity in sctx,
```
AlCutter: Does `sctx` need to change now that `SecurityContext` is not used?

impl/authorization/authorization.go
```diff
@@ -38,30 +42,37 @@ func New() authorization.Authorization {
...
 		}
 	}
-	return fmt.Errorf("%v is not authorized to perform %v on resource defined by <domainID=%v, appID=%v>", sctx.Identity(), permission, domainID, appID)
+	return status.Errorf(codes.PermissionDenied, "%v is not authorized to perform %v on %v", sctx.Email, permission, rLabel)
 }
 
 func resourceLabel(domainID, appID string) string {
-	return fmt.Sprintf("%v|%v", domainID, appID)
+	return fmt.Sprintf("domains/%v/apps/%v",
+		strings.Replace(domainID, "/", "_", -1),
```
AlCutter: this implies that two resources with differing domains (and/or apps) "blah/blah" and "blah_blah" are treated identically, is that ok?

cmd/keytransparency-server/main.go
```diff
@@ -121,11 +123,17 @@ func main() {
 	// Create gRPC server.
 	queue := mutator.MutationQueue(mutations)
 	ksvr := keyserver.New(tlog, tmap, logAdmin, mapAdmin,
-		entry.New(), auth, authz, domains, queue, mutations)
+		entry.New(), authz, domains, queue, mutations)
 	grpcServer := grpc.NewServer(
 		grpc.Creds(creds),
+		// All streaming methods are unauthenticated for now.
```
Martin2112: That looks like a TODO to me.

cmd/keytransparency-server/main.go
```diff
@@ -121,11 +123,17 @@ func main() {
...
 		grpc.StreamInterceptor(grpc_prometheus.StreamServerInterceptor),
-		grpc.UnaryInterceptor(grpc_prometheus.UnaryServerInterceptor),
+		grpc.UnaryInterceptor(grpc_middleware.ChainUnaryServer(
+			grpc_prometheus.UnaryServerInterceptor,
+			authentication.UnaryServerInterceptor(map[string]grpc_auth.AuthFunc{
+				"/google.keytransparency.v1.KeyTransparency/UpdateEntry": authFunc,
```
Martin2112: Can we avoid hardcoding this? Presumably there will be a bunch of methods under the same prefix x/y in future.

impl/authentication/context.go
```diff
@@ -14,24 +14,25 @@
...
-	"errors"
-)
+import "context"
 
-var (
-	// ErrMissingAuth occurs when authentication information is missing.
-	ErrMissingAuth = errors.New("auth: missing authentication header")
-)
+// ValidatedSecurity is the auth value stored in the Contexts.
+type ValidatedSecurity struct {
```
Martin2112: Maybe this should be called SecurityContext, or the key renamed so they match.

impl/authentication/fake_test.go
```diff
@@ -0,0 +1,56 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
```
Martin2112: Time passes ... Thorin sings about gold.

impl/authentication/fake_test.go
```diff
@@ -0,0 +1,56 @@
...
+func TestBasicValidateCreds(t *testing.T) {
+	ctx := context.Background()
+	for _, tc := range []struct {
+		desc      string
+		ctx       context.Context
+		wantEmail string
+		wantCode  codes.Code
+	}{
+		{desc: "missing authentication", ctx: ctx, wantCode: codes.Unauthenticated},
+		{desc: "working case", ctx: WithOutgoingFakeAuth(ctx, "foo"), wantEmail: "foo"},
```
Martin2112: Aren't there more than two cases e.g. corrupt token? There are multiple errors defined below, which makes me think so.

impl/authentication/google.go
```diff
@@ -99,13 +104,16 @@ func setDifference(a, b []string) []string {
 	return diff
 }
 
+func parseToken(accessToken string) *oauth2.Token {
```
Martin2112: Does this add much value?

impl/authentication/interceptor.go
Martin2112: +1

impl/authorization/authorization.go
```diff
@@ -38,30 +42,37 @@ func New() authorization.Authorization {
 // authorized to carry the given permission. A call is authorized if:
 //  1. userID matches the identity in sctx,
 //  2. or, sctx's identity is authorized to do the action in domainID and appID.
-func (a *authz) IsAuthorized(sctx *authentication.SecurityContext,
+func (a *authz) Authorize(ctx context.Context,
 	domainID, appID, userID string, permission authzpb.Permission) error {
+	sctx, ok := authentication.FromContext(ctx)
+	if !ok {
+		return status.Errorf(codes.Unauthenticated, "Request does not contain a ValidatedSecurity object")
```
Martin2112: The other code returns fixed errors but this doesn't. Should this do the same?

impl/authorization/authorization_test.go
```diff
@@ -85,101 +89,105 @@ func setup() *authz {
...
 	} {
-		err := a.IsAuthorized(tc.sctx, tc.domainID, tc.appID, tc.userID, tc.permission)
-		if got, want := err == nil, tc.success; got != want {
-			t.Errorf("%v: IsAuthorized err == nil: %v, want %v", tc.description, got, want)
+		// Convert outgoing context to incoming context.
+		inCtx := metautils.ExtractOutgoing(tc.ctx).ToIncoming(ctx)
+		sctx, err := authentication.FakeAuthFunc(inCtx)
+		if err != nil {
+			t.Errorf("FakeAuthFunc(): %v", err)
+			continue
```
Martin2112: maybe t.FailNow() ?

impl/integration/env.go
```diff
@@ -159,13 +160,19 @@ func NewEnv() (*Env, error) {
...
 
 	queue := mutator.MutationQueue(mutations)
 	server := keyserver.New(logEnv.Log, mapEnv.Map, logEnv.Admin, mapEnv.Admin,
-		entry.New(), auth, authz, domainStorage, queue, mutations)
-	gsvr := grpc.NewServer()
+		entry.New(), authz, domainStorage, queue, mutations)
+	gsvr := grpc.NewServer(
+		grpc.UnaryInterceptor(
+			authentication.UnaryServerInterceptor(map[string]grpc_auth.AuthFunc{
+				"/google.keytransparency.v1.KeyTransparency/UpdateEntry": authFunc,
```
Martin2112: Deja vu.

cmd/keytransparency-server/main.go
gdbelvin: I don't anticipate adding authenticated streaming methods right now. 
This comment is a reminder to add the interceptor if we do add an authenticated streaming rpc.

cmd/keytransparency-server/main.go
gdbelvin: We could replace this with a regex in the future if we need to, but simpler seems easier to test and maintain for the moment.

impl/authentication/context.go
gdbelvin: Renamed.

gdbelvin: Renamed

impl/authentication/fake_test.go
gdbelvin: :-)

impl/authentication/fake_test.go
gdbelvin: `FakeAuthFunc` is fairly simple.  I'm getting 100% code coverage.

impl/authentication/google.go
gdbelvin: done

impl/authentication/google.go
gdbelvin: Not really, except that it nicely separates the ideas of parsing and authenticating.

impl/authentication/interceptor.go
gdbelvin: Yep

impl/authentication/interceptor.go
gdbelvin: yes

impl/authentication/interceptor.go
gdbelvin: done

impl/authorization/authorization.go
gdbelvin: The other fixed errors weren't really needed. I've removed them now.

impl/authorization/authorization.go
gdbelvin: The `/` character is not allowed in domainIDs or appIDs.  If they were allowed, a specially crafted domainID or appID could cross boundaries. eg `domainID = "blah/apps/apptakeover`

impl/authorization/authorization_test.go
gdbelvin: We want to run the rest of the tests :-)
I'll put this in a `t.Run()`

impl/integration/env.go
gdbelvin: The integration tests act like a sort of "main", wiring together libraries much like main does. 
As a result, this intentionally lives in `impl/` and allows other environments to setup their own custom `envs`

impl/authorization/authorization.go
AlCutter: Sure, but my point is that there are now multiple "preimages" which result in identical resource labels - this function doesn't enforce the rule that "/"s aren't allowed, it just quietly papers over that situation resulting in the multiple "preimage" thing.
Maybe that's OK, I don't know :)

Is the rule about the contents of domainID and appID enforced elsewhere? 
if so this function possibly doesn't need to `Sprintf`, and if it isn't enforced elsewhere then maybe this function should return an error in the case where a passed-in ID contains the character?

gdbelvin:
PTAL

cmd/keytransparency-server/main.go
Martin2112: OK, as you're the one who'll have to write the postmortem when that doesn't happen.

impl/authentication/fake_test.go
Martin2112: OK we shouldn't really be testing fakes though. Are there tests for the multiple possible errors defined or is that elsewhere?

impl/authorization/authorization.go
Martin2112: OK, this possibly addresses my point above.

impl/authorization/authorization_test.go
Martin2112: r.Run() is better.

impl/integration/env.go
Martin2112: OK if it doesn't rely on those strings matching (my earlier point) then it might not be a problem.

Martin2112:
I think the potential only blocker is the point Al made about the preimages.

impl/authorization/authorization.go
gdbelvin: Excellent point. I'll make this return an error.

cmd/keytransparency-server/main.go
gdbelvin: I'll add a reference to the streaming interceptor with an empty map to make things clearer.

impl/authentication/fake_test.go
gdbelvin: `TestGoogleAuthn` tests those code paths.

impl/integration/env.go
gdbelvin: I don't see a place in the generated proto files where these strings are publicly exported.  Not sure what the best solution is.

gdbelvin:
Fixed the resource label pre-image issue.

gdbelvin:
PTAL

PR #974, opened 2018-05-11
Also move authorization to the grpc interceptor

gdbelvin:
This PR completes the move from server based authentication and authorization to grpc interceptor based authorization and authentication. 

This should make it much easier for specific deployments to swap in their own appropriate authentication and authorization functions. 

Diffbase #973


impl/authentication/context.go
```diff
@@ -14,24 +14,25 @@
...
+}
+
+// securityContextKey identifies ValidatedSecurity within context.Context.
+var securityContextKey struct{}
 
-// Authenticator provides services to authenticate users.
-type Authenticator interface {
-	// ValidateCreds authenticate the information present in ctx.
-	ValidateCreds(ctx context.Context) (*SecurityContext, error)
+// FromContext returns a ValidatedSecurity from the current context.
```
Martin2112: This doc wasn't updated for the rename (x2 mentions)

gdbelvin: Thanks.

PR #977, opened 2018-05-16
Add directory structure layout to README

### LINKED ISSUE #540: Explain repo layout
Add a bit of documentation explaining which bits of the repo are designed to be common, which interfaces need to be implemented differently for other production environments, etc.

gdbelvin:
Fixes #540


README.md
```diff
@@ -113,8 +113,27 @@ Key Transparency and its [Trillian](https://github.com/google/trillian) backend
...
+    * [**sequencer**](core/sequencer): sequencer implementation
+    * [**adminserver**](core/adminserver): admin server implementation
+* [**impl**](impl): environment specific modules:
+    * [**authentication**](impl/authentication): grpc authentication interceptor
+    * [**sql**](impl/sql): mysql implementations of storage modules
+    * [**integration**](impl/integration): environment specific integration tests
+* [**cmd**](cmd): binaries
+    * [**keytransparency-client**](cmd/keytransparency-client): Key Transparency CLI client
+* [**scripts**](scripts): scripts
+    * [**deploy**](scripts/deploy.sh): Deploy to Google Compute Engine
```
taknira: There are some caps vs not caps inconsistencies after ':' that could be unified - e.g.
```
[**impl**](impl): environment...
```
vs
```
[**deploy**](scripts/deploy.sh): Deploy...
```

taknira: It might be worth ordering these in the same order as they are listed in the repo, for ease of navigation/lookup of what's in the repo.  But you may want to keep this ordering that lists them in order of importance (I'm assuming), so this is completely up to you :)

taknira: Even though the docs, testdata and vendor dirs are pretty self explanatory, it might still be worth listing every top-level dir in the repo for completeness - seeing only 4 top-level dirs on this list, and then seeing 8 in the repo made me go and check what wasn't on this list.

taknira: There's also a top-level deploy dir - how does that differ from scripts/deploy?  It looks like scripts/deploy is actually scripts/deploy.sh, so should probably be *displayed* as such - currently it looks like deploy is a subdir of scripts.

gdbelvin: They should all be lower case. Fixed.

gdbelvin: It was in order of importance, maybe I can use **bold** to hint at what's important

gdbelvin: I've added more folders, but I'm not sure it makes sense to list everything and keep it in sync.

gdbelvin: Good catch. This was incorrect.

README.md
```diff
@@ -113,8 +113,38 @@ Key Transparency and its [Trillian](https://github.com/google/trillian) backend
...
+    * [**kubernetes**](deploy/kubernetes): kube deploy configs.
+    * [prometheus](deploy/prometheus): monitoring docker module.
+* [**docs**](docs): documentation.
+* [**impl**](impl): environment specific modules:
+    * [**authentication**](impl/authentication): authentication policy grpc interceptor.
+    * [**authorization**](impl/authorization): OAuth and fake auth grpc interceptor.
+    * [integration](impl/integration): environment specific integration tests.
+    * [**sql**](impl/sql): mysql implementations of storage modules.
+* [**scripts**](scripts): scripts
+    * [**deploy**](scripts/deploy.sh): deploy to Google Compute Engine.
```
taknira: This line still makes it seem like deploy is a subdir of scripts.  Maybe change it to:
```
* [**deploy.sh**](scripts/deploy.sh): deploy to Google Compute Engine.
```

PR #988, opened 2018-06-05
Place MapRoots into Log by index

gdbelvin:
Now that PREORDERED_LOGs are a thing, we can insert map roots into the
log in a deterministic way, and remove the sword of Damocles that was
previously there due to non-deterministic log sequencing

### DIFF
--- a/core/sequencer/sequencer.go
+++ b/core/sequencer/sequencer.go
@@ -316,8 +316,8 @@ func (s *Sequencer) createEpoch(ctx context.Context, d *domain.Domain, logClient
 	}
 
 	// Put SignedMapHead in an append only log.
-	if err := queueLogLeaf(ctx, logClient, setResp.GetMapRoot()); err != nil {
-		glog.Fatalf("queueLogLeaf(logID: %v, rev: %v): %v", d.LogID, mapRoot.Revision, err)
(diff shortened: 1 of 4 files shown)


core/sequencer/sequencer.go
```diff
@@ -329,9 +329,3 @@ func (s *Sequencer) createEpoch(ctx context.Context, d *domain.Domain, logClient
 	glog.Infof("CreatedEpoch: rev: %v, root: %x", mapRoot.Revision, mapRoot.RootHash)
 	return nil
 }
-
-// TODO(gdbelvin): Add leaf at a specific index. trillian#423
-func queueLogLeaf(ctx context.Context, logClient *tclient.LogClient, smr *tpb.SignedMapRoot) error {
-	// Queue the leaf and then wait until it has been sequenced and verified.
-	return logClient.AddLeaf(ctx, smr.MapRoot)
```
pavelkalinnikov: For a potential follow-up PR: Maybe `AddLeafAndWait` would be less a misleading name than `AddLeaf`? For a moment I thought "Why does the new code waits and the old one doesn't?", but then I realized the old one did wait.

pavelkalinnikov:
LGTM but let's wait until it passes CT tests.

core/sequencer/sequencer.go
gdbelvin: yep. I'll make that change in Trillian

PR #990, opened 2018-06-06
Remove duplicate GetLatestEpoch

gdbelvin:
Remove a duplicate, and inconsitent call to getLatestLogRoot.
Previously, GetLatestEpoch had a flow like this:

 * GetLatestLogRoot
 * Calculate LatestMapRevision
 * **GetLatestLogRoot** (which could be different!)
 * GetMapRevision

### DIFF
--- a/core/client/get_and_verify.go
+++ b/core/client/get_and_verify.go
@@ -77,7 +77,7 @@ func (c *Client) VerifiedGetLatestEpoch(ctx context.Context) (*types.LogRootV1, 
 		return nil, nil, err
 	}
 	if smr.Revision != wantRevision {
-		return nil, nil, fmt.Errorf("GetLatestEpoch(): smr.Revison: %v + 1 != slr.TreeSize: %v", smr.Revision, slr.TreeSize)
+		return nil, nil, fmt.Errorf("GetLatestEpoch(): smr.Revison: %v != slr.TreeSize-1: %v", smr.Revision, slr.TreeSize-1)
 	}
 	return slr, smr, nil
 }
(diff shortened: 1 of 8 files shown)


core/keyserver/epochs.go
```diff
@@ -192,11 +184,8 @@ func (s *Server) logProofs(ctx context.Context, d *domain.Domain, firstTreeSize
 		glog.Errorf("logProofs(): log.GetInclusionProof(%v, %v, %v): %v", d.LogID, epoch, secondTreeSize, err)
```
AlCutter: Do you need to update the error message to reflect the new func name?

core/client/get_and_verify.go
```diff
@@ -77,7 +77,7 @@ func (c *Client) VerifiedGetLatestEpoch(ctx context.Context) (*types.LogRootV1,
 		return nil, nil, err
 	}
 	if smr.Revision != wantRevision {
-		return nil, nil, fmt.Errorf("GetLatestEpoch() did not return latest map revision. Got MapRoot.Revison: %v, want: %v", smr.Revision, wantRevision)
+		return nil, nil, fmt.Errorf("GetLatestEpoch(): smr.Revison: %v + 1 != slr.TreeSize: %v", smr.Revision, slr.TreeSize)
```
AlCutter: Would the message `("... smr.Revision (%v) != slr.TreeSize-1 (%v)", smr.Revision, wantRevision)` better represent what's actually being tested?

core/keyserver/epochs.go

core/client/get_and_verify.go
gdbelvin: yep

gdbelvin:
PTAL

PR #991, opened 2018-06-12
Set tree display names

gdbelvin:
Make monitoring graphs easier to use by setting the trillian tree display name, which is used as the primary key.


Martin2112:
Can you fix the lint error.

gdbelvin:
PTAL

PR #992, opened 2018-06-14
Use flexible implementations with metric factory

gdbelvin:
Injects prometheus in the open source implementation, while
allowing for other monitoring implementations.

### DIFF
--- a/core/sequencer/sequencer.go
+++ b/core/sequencer/sequencer.go
@@ -43,7 +43,7 @@ const (
 var (
 	once               sync.Once
 	knownDomains       monitoring.Gauge
-	mutationsCTR       monitoring.Counter
+	mutationCount      monitoring.Counter
 	sequencingRuns     monitoring.Counter
 	sequencingFailures monitoring.Counter
 	sequencingLatency  monitoring.Histogram
@@ -54,8 +54,8 @@ func createMetrics(mf monitoring.MetricFactory) {
 		"known_domains",
 		"Set to 1 for known domains (whether this instance is master or not)",
 		domainIDLabel)
-	mutationsCTR = mf.NewCounter(
-		"mutations",
+	mutationCount = mf.NewCounter(
+		"mutation_count",
 		"Number of mutations the signer has processed",
 		domainIDLabel)
 	sequencingRuns = mf.NewCounter(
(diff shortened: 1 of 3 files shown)


core/sequencer/sequencer.go
```diff
@@ -27,42 +27,49 @@ import (
...
+	MaxBatchSize  = int32(1000)
+	domainIDLabel = "domainid"
 )
 
-// MaxBatchSize limits the number of mutations that will be processed per epoch.
-const MaxBatchSize = int32(1000)
+var (
+	once               sync.Once
+	knownDomains       monitoring.Gauge
+	mutationsCTR       monitoring.Counter
```
phad: Can this be `mutationsCount`?
To me, CTR == click-through rate

core/sequencer/sequencer.go
```diff
@@ -120,6 +130,7 @@ func (s *Sequencer) ListenForNewDomains(ctx context.Context, refresh time.Durati
 				return fmt.Errorf("admin.List(): %v", err)
 			}
 			for _, d := range domains {
+				knownDomains.Set(1.0, d.DomainID)
```
phad: does the Gauge metric only permit floats to be stored? not ints?

core/sequencer/sequencer.go
```diff
@@ -27,42 +27,49 @@ import (
...
-func init() {
-	prometheus.MustRegister(mutationsCTR)
-	prometheus.MustRegister(indexCTR)
-	prometheus.MustRegister(mapUpdateHist)
-	prometheus.MustRegister(createEpochHist)
+func createMetrics(mf monitoring.MetricFactory) {
+	knownDomains = mf.NewGauge(
+		"known_domains",
+		"Set to 1 for known domains (whether this instance is master or not)",
+		domainIDLabel)
```
phad: Is the final parameter 'domainid' defining a field within known_domains over which separate counts will be made?

core/sequencer/sequencer.go
gdbelvin: Done

core/sequencer/sequencer.go
gdbelvin: That's right.  However I can use `1` rather than `1.0` because, in Go, numbers are not typed.

core/sequencer/sequencer.go
gdbelvin: That's right.  My rough understanding is that this will produce a metric which contains a map from domain_id to this 1/0 count.

PR #993, opened 2018-06-19
Generic load test

gdbelvin:
To test multiple operation types, this PR refactors the hammer library a bit to measure arbitrary operations.  

 The new components:
- Calculating stats on a stream of results.
- Measuring the latency of a stream of requests mapped to a collection of workers. 
- Better speed control of the generated stream of requests.  (time.Ticker doesn't go faster than 10QPS)
- New operation types for 
   - Batch Write
   - Parallel Write 
   - Parallel Read
   - Parallel Audit History


core/client/hammer/benchmark.go
```diff
@@ -0,0 +1,58 @@
...
+				}
+			}(rh)
+		}
+		wg.Wait()
+		close(results)
+	}()
+	return results
+}
+
+func generateReport(ctx context.Context, reqs <-chan request, handlers []ReqHandler) *Stats {
```
daviddrysdale: Could do with a better name -- to me `generateReport` implies more of what `collectStats` is doing, and the running of various handler code is a bit of a surprise.

core/client/hammer/benchmark.go
```diff
@@ -0,0 +1,58 @@
...
+	results := startHandlers(ctx, reqs, handlers)
+	stats := collectStats(results)
+	stats.Print()
```
daviddrysdale: Feels slightly like this code belongs in the caller, so it can control the individual steps of:
 - run a bunch of requests through the handlers
 - collate the results of those requests into statistics
 - output the results somewhere.

core/client/hammer/benchmark.go
```diff
@@ -0,0 +1,58 @@
+// Copyright 2018 Google Inc. All Rights Reserved.
```
daviddrysdale: Could do with a better filename -- this isn't entirely benchmark-specific code.

core/client/hammer/benchmark_test.go
```diff
@@ -0,0 +1,105 @@
...
+import (
+	"context"
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestGenerateReport(t *testing.T) {
```
daviddrysdale: Testing a test tool might be overkill. 

(It's also testing two different things in combination, which might be better broken out)

cmd/keytransparency-client/cmd/hammer.go
```diff
@@ -75,8 +73,23 @@ var hammerCmd = &cobra.Command{
 			return err
 		}
 
-		h.Run(ctx, maxOperations, maxWorkers, ramp)
-		return nil
+		return h.Run(ctx, maxWorkers, hammer.Config{
```
daviddrysdale: Maybe a `TODO: expose more of these options as flags` ?

core/client/hammer/benchmark.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"context"
+	"sync"
+	"time"
+)
+
+// ReqHandler executes a request.
+type ReqHandler func(ctx context.Context, req *request) error
+
+type request struct {
```
daviddrysdale: It would be nice to find a less generic name for this.  `userBatch` maybe?

core/client/hammer/benchmark.go
```diff
@@ -0,0 +1,58 @@
...
+	UserIDs   []string
+	BatchSize int
```
daviddrysdale: Is `BatchSize` different than `len(UserIDs)` ?  Would `PageSize` be a better name?

core/client/hammer/hammer.go
```diff
@@ -76,125 +93,159 @@ func New(ctx context.Context, dial DialFunc, callOptions CallOptions,
...
 }
 
-func (h *Hammer) worker(ctx context.Context, jobs <-chan int, outputs chan<- error) {
-	client, err := client.NewFromConfig(h.ktCli, h.config)
-	if err != nil {
-		glog.Errorf("NewFromConfig(): %v", err)
-		return
+// batchWriteOp queues many user mutations, waits, and then verifies them all.
+func (w *worker) writeOp(ctx context.Context, req *request) error {
+	users := make([]*tpb.User, 0, req.BatchSize)
```
daviddrysdale: `len(req.UserIDs)` not `req.BatchSize` ?

core/client/hammer/hammer.go
```diff
@@ -76,125 +93,159 @@ func New(ctx context.Context, dial DialFunc, callOptions CallOptions,
...
-	cctx, cancel := context.WithTimeout(ctx, h.timeout)
-	defer cancel()
-
-	u := &tpb.User{
-		DomainId:       h.config.DomainId,
-		AppId:          h.appID,
-		UserId:         userID,
-		PublicKeyData:  []byte("publickey"),
-		AuthorizedKeys: h.authorizedKeys,
+// readOp simulates a read operation, typically performed during conversation setup.
```
daviddrysdale: "simulates many user read operations" ?

core/client/hammer/hammer.go
```diff
@@ -76,125 +93,159 @@ func New(ctx context.Context, dial DialFunc, callOptions CallOptions,
...
+	client *client.Client
+}
 
-	// Collect outputs until output closes.
-	for err := range outputs {
-		totalRequests++
+func (h *Hammer) newWorkers(n int) ([]worker, error) {
+	workers := make([]worker, n)
+	for i := range workers {
+		// Give each worker it's own client.
```
daviddrysdale: s/it's/its/

core/client/hammer/hammer.go
```diff
@@ -76,125 +93,159 @@ func New(ctx context.Context, dial DialFunc, callOptions CallOptions,
...
+			}
+			inflightReqs <- request{
+				UserIDs:   userIDs,
+				BatchSize: batch,
 			}
 		}
 	}()
-	return jobs
+	return inflightReqs
 }
```
daviddrysdale: I like that there's no concurrency-related stuff from here onwards in the file, just a bunch of linear functions -- it makes it much easier to think about the concurrency when it's concentrated in a smaller number of places.

core/client/hammer/hammer.go
```diff
@@ -76,125 +93,159 @@ func New(ctx context.Context, dial DialFunc, callOptions CallOptions,
...
-			h.workers++
-			wg.Add(1)
-			go func() {
-				defer wg.Done()
-				h.worker(ctx, jobs, outputs)
-			}()
-			<-rampTicker.C
-		}
-		rampTicker.Stop()
+// workers should roughly be (goal QPS / timeout seconds)
```
daviddrysdale: Comment out of date (no ramping any more)?

core/client/hammer/hammer.go
```diff
@@ -76,125 +93,159 @@ func New(ctx context.Context, dial DialFunc, callOptions CallOptions,
...
+	generateReport(ctx, requests, handlers)
 
-	return outputs
+	return nil
 }
 
-// gen produces a channel of integers from 0 up to count-1 and then closes.
-func gen(ctx context.Context, count int) <-chan int {
-	jobs := make(chan int)
+func genRequests(ctx context.Context, qps, batch, count int, duration time.Duration) <-chan request {
```
daviddrysdale: Maybe `requestGenerator` rather than `genRequests` to give a hint that it hasn't generated all of the requests by the time it returns.

core/client/hammer/hammer.go
```diff
@@ -76,125 +93,159 @@ func New(ctx context.Context, dial DialFunc, callOptions CallOptions,
...
-		close(outputs)
-	}()
+	// Batch Write users
+	log.Printf("Batch Write")
+	requests := genRequests(ctx, c.BatchWriteQPS, c.BatchWriteSize, c.BatchWriteCount, c.Duration)
+	handlers := make([]ReqHandler, 0, len(workers))
+	for _, w := range workers {
+		handlers = append(handlers, w.writeOp)
+	}
+	generateReport(ctx, requests, handlers)
```
daviddrysdale: Maybe this should be `executeAllRequests()` or somesuch, to hint that there's goroutine stuff going on under the covers?

core/client/hammer/hammer.go
```diff
@@ -76,125 +93,159 @@ func New(ctx context.Context, dial DialFunc, callOptions CallOptions,
...
+	// Read users
+	log.Printf("User Read")
+	requests = genRequests(ctx, c.ReadQPS, 1, c.ReadCount, c.Duration)
+	handlers = make([]ReqHandler, 0, len(workers))
+	for _, w := range workers {
+		handlers = append(handlers, w.readOp)
+	}
+	generateReport(ctx, requests, handlers)
+
+	// History
```
daviddrysdale: Running 4 distinct phases one after another feels a bit limiting -- it might be nice to be able to select which phase(s) to run, and possibly to allow parallel execution of them.

Idle thought: a `--test_styles=batch,read,audit` command line option that the selected phases in parallel?  Then you could do the current setup with `hammer --style=batch; hammer --style=write; hammer --style=read; hammer --style=audit`

core/client/hammer/hammer.go
gdbelvin: done

core/client/hammer/hammer.go
gdbelvin: Yes, thank you.

core/client/hammer/hammer.go
gdbelvin: done

core/client/hammer/hammer.go
gdbelvin: 👍

core/client/hammer/hammer.go
gdbelvin: Yep. Removed.

core/client/hammer/benchmark.go
gdbelvin: now that `collectStats` is gone, this function can go away too.

core/client/hammer/benchmark.go
gdbelvin: Removed

core/client/hammer/benchmark_test.go
gdbelvin: Also removed.

cmd/keytransparency-client/cmd/hammer.go
gdbelvin: Exposed

core/client/hammer/benchmark.go
gdbelvin: This feels like the same level of genericness as `ReqHandler`

core/client/hammer/benchmark.go
gdbelvin: Good idea.

core/client/hammer/hammer.go
gdbelvin: done

core/client/hammer/hammer.go
gdbelvin: done

cmd/keytransparency-client/cmd/hammer.go
```diff
@@ -41,9 +40,8 @@ func init() {
 
 	RootCmd.AddCommand(hammerCmd)
 
-	hammerCmd.Flags().IntVar(&maxWorkers, "workers", 1, "Number of parallel workers")
+	hammerCmd.Flags().IntVar(&maxWorkers, "workers", 2000, "Number of parallel workers")
```
pavelkalinnikov: 2000 looks a bit excessive a default to me (esp. as a number of workers). Have you considered 100 or 200?

core/client/hammer/report.go
```diff
@@ -0,0 +1,90 @@
...
+
+// Stats hold statistics on the output of many operations.
+type Stats struct {
+	RequestCount int64
+	AvgTotal     time.Duration // Sum of individual latencies.
+	Total        time.Duration // Total wall time.
+	Fastest      time.Duration
+	Slowest      time.Duration
+	Average      time.Duration // Sum of latencies / count
+	QPS          float64       // Wall time / count
```
pavelkalinnikov: `s/count/RequestCount/` here and above?

core/client/hammer/report.go
```diff
@@ -0,0 +1,90 @@
...
+		ErrorDist: make(map[string]int),
+	}
+	st := time.Now()
+	for res := range results {
+		if res.Err != nil {
+			s.ErrorDist[res.Err.Error()]++
+			continue
+		}
+
+		s.RequestCount++
```
pavelkalinnikov: To me `RequestCount` looks more like a total #Ok+#Err counter rather than just #Ok. Rename to `OkCount`?

core/client/hammer/report.go
```diff
@@ -0,0 +1,90 @@
...
+		l := res.End.Sub(res.Start)
+		s.AvgTotal += l
+		s.Average = s.AvgTotal / time.Duration(s.RequestCount)
+		if l < s.Fastest || s.Fastest == 0 {
```
pavelkalinnikov: Is `collectStats` called only once per `Stats` instance? If so, maybe replace the `s.Fastest == 0` check with a local `isFirst` bool to avoid weird effects if suddenly `l == 0`.

cmd/keytransparency-client/cmd/hammer.go
gdbelvin: I lowered it to 1000. 

Now that the speed of the test is being controlled by `qps`, more workers don't hurt, and only ensure that there is enough parallelism to meet qps needs.

core/client/hammer/report.go
gdbelvin: Deleted file

core/client/hammer/report.go
gdbelvin: Deleted file

core/client/hammer/report.go
gdbelvin: Deleted file

core/client/hammer/benchmark.go
gdbelvin: hmm. well it was until we removed the benchmarking bit :-)

core/client/hammer/hammer.go
gdbelvin: Done

core/client/hammer/hammer.go
gdbelvin: done

core/client/hammer/hammer.go
gdbelvin: done

cmd/keytransparency-client/cmd/hammer.go
```diff
@@ -41,9 +44,11 @@ func init() {
 
 	RootCmd.AddCommand(hammerCmd)
 
-	hammerCmd.Flags().IntVar(&maxWorkers, "workers", 1, "Number of parallel workers")
+	hammerCmd.Flags().StringVar(&testTypes, "types", "batch,write,read,audit", "Types of stress tests to run")
```
pavelkalinnikov: optional: Add a `(comma-separated) `remark at the end.

cmd/keytransparency-client/cmd/hammer.go
```diff
@@ -41,9 +44,11 @@ func init() {
...
+	hammerCmd.Flags().IntVar(&qps, "qps", 100, "Numer of requests a second")
+	hammerCmd.Flags().IntVar(&pageSize, "batch", 10, "Number of things to do at once")
```
pavelkalinnikov: Is it possible to be more specific than "things"? If no word is suitable, maybe give an example of a "thing"?
E.g.: Max number of entries processed per one RPC.

cmd/keytransparency-client/cmd/hammer.go
```diff
@@ -41,9 +44,11 @@ func init() {
...
+	hammerCmd.Flags().IntVar(&maxWorkers, "workers", 1000, "Number of parallel workers")
```
pavelkalinnikov: Could you be a bit more specific on what the work of the workers is? E.g., RPC-sending workers.

core/client/hammer/hammer.go
```diff
@@ -76,125 +96,187 @@ func New(ctx context.Context, dial DialFunc, callOptions CallOptions,
...
 	}, nil
 }
 
-// Run runs operationCount with up to maxWorkers over ramp duration.
-func (h *Hammer) Run(ctx context.Context, operationCount, maxWorkers int, ramp time.Duration) {
-	jobs := gen(ctx, operationCount)
-	outputs := h.launchWorkers(ctx, jobs, maxWorkers, ramp)
-	h.printStats(1*time.Second, outputs)
-}
+// Run runs operationCount with up to maxWorkers.
```
pavelkalinnikov: numWorkers?
Is the "up to" bit really "exactly"? I couldn't find when less is possible.

core/client/hammer/hammer.go
```diff
@@ -76,125 +96,187 @@ func New(ctx context.Context, dial DialFunc, callOptions CallOptions,
...
-	wg.Wait()
+	return workers, nil
 }
 
-func (h *Hammer) worker(ctx context.Context, jobs <-chan int, outputs chan<- error) {
-	client, err := client.NewFromConfig(h.ktCli, h.config)
-	if err != nil {
-		glog.Errorf("NewFromConfig(): %v", err)
-		return
+// batchWriteOp queues many user mutations, waits, and then verifies them all.
```
pavelkalinnikov: writeOp?

core/client/hammer/hammer.go
```diff
@@ -76,125 +96,187 @@ func New(ctx context.Context, dial DialFunc, callOptions CallOptions,
...
+			PublicKeyData:  []byte("publickey"),
+			AuthorizedKeys: w.authorizedKeys,
+		})
+	}
+
+	mutations := make([]*entry.Mutation, 0, len(users))
+	for _, u := range users {
+		callOptions := w.callOptions(u.UserId)
+		cctx, cancel := context.WithTimeout(ctx, w.timeout)
+		defer cancel()
```
pavelkalinnikov: To double check, do you think it's okay to defer canceling until end of `writeOp` instead of before the next `for` iteration?

core/client/hammer/hammer.go
```diff
@@ -76,125 +96,187 @@ func New(ctx context.Context, dial DialFunc, callOptions CallOptions,
...
+		if err != nil {
+			fmt.Print("!")
+			return err
+		}
+		mutations = append(mutations, m)
+		if err := w.client.QueueMutation(cctx, m, w.signers, callOptions...); err != nil {
+			fmt.Print("!")
+			return err
+		}
+		fmt.Print("-")
```
pavelkalinnikov: Can this `Print` and other similar ones be interleaved because of having multiple concurrent workers? If so, perhaps a better approach would be to construct a string first and print it with a single call?

core/client/hammer/hammer.go
```diff
@@ -76,125 +96,187 @@ func New(ctx context.Context, dial DialFunc, callOptions CallOptions,
...
+		fmt.Print("\n")
+	}
+
+	return nil
 }
 
-// gen produces a channel of integers from 0 up to count-1 and then closes.
-func gen(ctx context.Context, count int) <-chan int {
-	jobs := make(chan int)
+func requestGenerator(ctx context.Context, qps, batch, count int, duration time.Duration) <-chan request {
```
pavelkalinnikov: Maybe `gen(erate)Requests`? Otherwise `request` can be read as a verb, which is slightly misleading.

core/client/hammer/benchmark.go
pavelkalinnikov: Maybe find a less generic name for both?

core/client/hammer/benchmark.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+func executeRequests(ctx context.Context, inflightReqs <-chan request, reqHandlers []ReqHandler) {
+	var wg sync.WaitGroup
+	for _, rh := range reqHandlers {
+		wg.Add(1)
+		go func(rh ReqHandler) {
+			defer wg.Done()
+			for req := range inflightReqs {
+				rh(ctx, &req)
```
golangci: Error return value of `rh` is not checked

core/client/hammer/hammer.go
```diff
@@ -76,125 +97,211 @@ func New(ctx context.Context, dial DialFunc, callOptions CallOptions,
...
+		requests := requestGenerator(ctx, c.HistoryQPS, c.HistoryPageSize, c.HistoryCount, c.Duration)
+		handlers := make([]ReqHandler, 0, len(workers))
+		for i := range workers {
+			handlers = append(handlers, workers[i].historyOp)
+		}
+		executeRequests(ctx, requests, handlers)
+		fmt.Print("\n")
+	}
+
+	pool.Stop()
```
golangci: Error return value of `pool.Stop` is not checked

core/client/hammer/hammer.go
gdbelvin: It was an old comment. Updated.

core/client/hammer/hammer.go
gdbelvin: Done.

core/client/hammer/hammer.go
gdbelvin: It's probably not ok because the parent function is long running.   Thanks for noticing. Fixed.

core/client/hammer/hammer.go
gdbelvin: These prints can be interleaved.  That said, I'm still searching for a better way to output status from many workers to the screen.

gdbelvin:
Updated. PTAL

core/client/hammer/hammer.go
```diff
@@ -76,125 +97,199 @@ func New(ctx context.Context, dial DialFunc, callOptions CallOptions,
...
 }
 
-// Run runs operationCount with up to maxWorkers over ramp duration.
-func (h *Hammer) Run(ctx context.Context, operationCount, maxWorkers int, ramp time.Duration) {
-	jobs := gen(ctx, operationCount)
-	outputs := h.launchWorkers(ctx, jobs, maxWorkers, ramp)
-	h.printStats(1*time.Second, outputs)
-}
+// Run runs a total of operationCount operations across numWorkers.
+// The number of workers should roughly be (goal QPS / timeout seconds)
```
pavelkalinnikov: It seems reasonable to replicate this "number of workers" remark in the corresponding `workers` flag description.

core/client/hammer/hammer.go
```diff
@@ -76,125 +97,199 @@ func New(ctx context.Context, dial DialFunc, callOptions CallOptions,
...
+		client, err := client.NewFromConfig(h.ktCli, h.domain)
 		if err != nil {
-			log.Printf("writeOp(): %v", err)
+			return nil, err
 		}
+
+		workers = append(workers, worker{
+			Hammer: h,
+			client: client,
+			bar:    cpb.New(100).Prefix(fmt.Sprintf("%v ", i)),
```
pavelkalinnikov: Do I understand correctly that in case of default 1000 workers you will have 1000 progress bars in the command line?

Have you considered doing just one? Or one for each type of hammering?

core/client/hammer/hammer.go
```diff
@@ -76,125 +97,199 @@ func New(ctx context.Context, dial DialFunc, callOptions CallOptions,
...
 }
 
-func (h *Hammer) worker(ctx context.Context, jobs <-chan int, outputs chan<- error) {
-	client, err := client.NewFromConfig(h.ktCli, h.config)
-	if err != nil {
-		glog.Errorf("NewFromConfig(): %v", err)
-		return
+// writeOp queues many user mutations, waits, and then verifies them all.
+func (w *worker) writeOp(ctx context.Context, req *opArg) error {
+	users := make([]*tpb.User, 0, len(req.UserIDs))
```
pavelkalinnikov: It appears that this `users` array is created only to be scanned below. Maybe merge the two loops?

```
for _, userID := range req.UserIDs {
  callOptions := w.callOptions(userID)
  u := &tpb.User{
    DomainId:       w.domain.DomainId,
    AppId:          w.appID,
    UserId:         userID,
    PublicKeyData:  []byte("publickey"),
    AuthorizedKeys: w.authorizedKeys,
  }
  ...
}
```

core/client/hammer/hammer.go
```diff
@@ -76,125 +97,199 @@ func New(ctx context.Context, dial DialFunc, callOptions CallOptions,
...
-		AppId:          h.appID,
-		UserId:         userID,
-		PublicKeyData:  []byte("publickey"),
-		AuthorizedKeys: h.authorizedKeys,
+// readOp simulates multiple read operations by a single client.
+// Typical conversation setup involves querying two userIDs: self and other.
+func (w *worker) readOp(ctx context.Context, req *opArg) error {
+	for _, userID := range req.UserIDs {
+		_, _, err := w.client.GetEntry(ctx, userID, w.appID)
+		if err != nil {
```
pavelkalinnikov: nit: Could merge this line with the previous one.

core/client/hammer/hammer.go
```diff
@@ -76,125 +97,199 @@ func New(ctx context.Context, dial DialFunc, callOptions CallOptions,
...
+		}
+		fmt.Print(".")
 	}
+	return nil
+}
 
-	callOptions := h.callOptions(userID)
-	_, err := client.Update(cctx, u, h.signers, callOptions...)
-	return err
+// auditHistoryOp simulates the daily check-in.
```
pavelkalinnikov: historyOp?

core/client/hammer/hammer.go
```diff
@@ -76,125 +97,199 @@ func New(ctx context.Context, dial DialFunc, callOptions CallOptions,
...
+func (w *worker) historyOp(ctx context.Context, req *opArg) error {
+	for _, userID := range req.UserIDs {
+		_, _, err := w.client.PaginateHistory(ctx, userID, w.appID, 0, int64(req.PageSize))
+		if err != nil {
```
pavelkalinnikov: nit: Could merge this line with the previous one.

core/client/hammer/hammer.go
```diff
@@ -76,125 +97,199 @@ func New(ctx context.Context, dial DialFunc, callOptions CallOptions,
...
-	jobs := make(chan int)
+func genArgs(ctx context.Context, qps, batch, count int, duration time.Duration) <-chan opArg {
+	inflightReqs := make(chan opArg, qps)
 	go func() {
-		defer close(jobs)
+		cctx, cancel := context.WithTimeout(ctx, duration)
+		defer cancel()
+		defer close(inflightReqs)
+
+		rateLimiter := rate.NewLimiter(rate.Limit(qps), qps)
```
pavelkalinnikov: With these parameters of rate limiter, I can expect that QPS in the long run will be okay, but in the short run it can jump to the skies (e.g., all Q tokens can go away in 10 ms, and the remaining 990 ms the generator places no requests in order to keep the average qps below Q). Is this intended?

If you never want to burst above the specified QPS, I guess the parameters should be `rate.NewLimiter(rate.Limit(qps), 1)`? However, take a look at the comment next to `WaitN` invocation.

core/client/hammer/hammer.go
```diff
@@ -76,125 +97,199 @@ func New(ctx context.Context, dial DialFunc, callOptions CallOptions,
...
 		for i := 0; i < count; i++ {
-			select {
-			case <-ctx.Done():
+			userIDs := make([]string, 0, batch)
+			for j := 0; j < batch; j++ {
+				userIDs = append(userIDs, fmt.Sprintf("user_%v", i*batch+j))
+			}
+			if err := rateLimiter.WaitN(cctx, batch); err != nil {
```
pavelkalinnikov: Note that WaitN `... returns an error if n exceeds the Limiter's burst size`, i.e. in your case if `batch > qps`. For default ones listed in flags it works, but not for arbitrary params. If you wanted to make it work always, you would need to create the limiter like this: `rate.NewLimiter(rate.Limit(qps), batch)`.

core/client/hammer/hammer.go
```diff
@@ -76,125 +97,199 @@ func New(ctx context.Context, dial DialFunc, callOptions CallOptions,
...
+				log.Printf("Duration %v elapsed, stopping request generation", duration)
```
pavelkalinnikov: Elapsed duration is not the only reason why the rate limiter can return an error. Maybe just log the error?

core/client/hammer/hammer.go
```diff
@@ -76,125 +97,199 @@ func New(ctx context.Context, dial DialFunc, callOptions CallOptions,
...
+		handlers := make([]ReqHandler, 0, len(workers))
+		for i := range workers {
+			handlers = append(handlers, workers[i].historyOp)
+		}
+		executeRequests(ctx, args, handlers)
+		fmt.Print("\n")
+	}
 
-	return outputs
+	return nil //pool.Stop()
```
pavelkalinnikov: Please act on this commented code.

core/client/hammer/hammer.go
```diff
@@ -76,125 +97,199 @@ func New(ctx context.Context, dial DialFunc, callOptions CallOptions,
...
+			return err
+		}
+	*/
 
-// launchWorkers gradually adds workers to the worker pool over ramp up to workers.
-func (h *Hammer) launchWorkers(ctx context.Context, jobs <-chan int, workers int, ramp time.Duration) <-chan error {
-	outputs := make(chan error)
+	if ok := c.TestTypes["batch"]; ok {
+		// Batch Write users
+		log.Printf("Batch Write")
```
pavelkalinnikov: nit: `Printf` -> `Print`? Same below.

core/client/hammer/hammer.go
```diff
@@ -76,125 +97,199 @@ func New(ctx context.Context, dial DialFunc, callOptions CallOptions,
...
+		args := genArgs(ctx, c.BatchWriteQPS, c.BatchWriteSize, c.BatchWriteCount, c.Duration)
+		handlers := make([]ReqHandler, 0, len(workers))
+		for i := range workers {
+			handlers = append(handlers, workers[i].writeOp)
+		}
+		log.Printf("workers: %v", len(handlers))
+		executeRequests(ctx, args, handlers)
+		fmt.Print("\n")
```
pavelkalinnikov: What's the purpose of this new line? Could it be put together with "Batch Write" above? Note that the dots printed by `executeRequests` are likely to appear after this `Print`.

core/client/hammer/hammer.go
```diff
@@ -19,15 +19,16 @@ import (
...
 
-	"github.com/golang/glog"
 	"github.com/google/tink/go/tink"
+	"golang.org/x/time/rate"
 	"google.golang.org/grpc"
 
 	"github.com/google/keytransparency/core/client"
+	"github.com/google/keytransparency/core/mutator/entry"
 
+	cpb "github.com/cheggaaa/pb"
```
daviddrysdale: I don't think this is the canonical import path for this new dependency - should it be `gopkg.in/cheggaaa/pb.v1` instead?

cmd/keytransparency-client/cmd/hammer.go
```diff
@@ -41,9 +45,12 @@ func init() {
 
 	RootCmd.AddCommand(hammerCmd)
 
-	hammerCmd.Flags().IntVar(&maxWorkers, "workers", 1, "Number of parallel workers")
+	hammerCmd.Flags().DurationVar(&duration, "duration", 10*time.Minute, "Maximum time to run each test")
+	hammerCmd.Flags().StringVar(&testTypes, "types", "batch,write,read,audit", "Types of stress tests to run")
```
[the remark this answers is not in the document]
gdbelvin: done

cmd/keytransparency-client/cmd/hammer.go
```diff
@@ -41,9 +45,12 @@ func init() {
...
+	hammerCmd.Flags().IntVar(&qps, "qps", 100, "Numer of requests a second")
+	hammerCmd.Flags().IntVar(&pageSize, "batch", 10, "Number of things to do at once")
```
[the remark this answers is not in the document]
gdbelvin: done

cmd/keytransparency-client/cmd/root.go
```diff
@@ -74,7 +74,7 @@ func init() {
 	RootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.keytransparency.yaml)")
 
 	RootCmd.PersistentFlags().String("domain", "default", "Domain within the KT server")
-	RootCmd.PersistentFlags().String("kt-url", "35.224.99.110:443", "URL of Key Transparency server")
+	RootCmd.PersistentFlags().String("kt-url", "35.202.56.9:443", "URL of Key Transparency server")
```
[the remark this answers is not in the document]
gdbelvin: The IP is publicly available to facilitate testing.  Unfortunately, I haven't quite figured out how to get a super stable kubernetes IP.

core/client/hammer/benchmark.go
```diff
@@ -0,0 +1,46 @@
...
+	"context"
+	"sync"
+
+	"github.com/golang/glog"
+)
+
+// ReqHandler executes a request.
+type ReqHandler func(ctx context.Context, arg *opArg) error
+
+type opArg struct {
```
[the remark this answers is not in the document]
gdbelvin: done

core/client/hammer/benchmark.go
```diff
@@ -0,0 +1,46 @@
+// Copyright 2018 Google Inc. All Rights Reserved.
+//
```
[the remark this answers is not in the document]
gdbelvin: done

core/client/hammer/hammer.go
```diff
@@ -39,15 +40,35 @@ type DialFunc func(ctx context.Context, addr string, opts ...grpc.DialOption) (p
 // CallOptions returns PerRPCCredentials for the requested user.
 type CallOptions func(userID string) []grpc.CallOption
 
+// Config tells the hammer how fast to go.
```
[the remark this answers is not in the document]
gdbelvin: done

core/client/hammer/hammer.go
```diff
@@ -76,125 +97,199 @@ func New(ctx context.Context, dial DialFunc, callOptions CallOptions,
...
-}
+// Run runs a total of operationCount operations across numWorkers.
+// The number of workers should roughly be (goal QPS / timeout seconds)
+func (h *Hammer) Run(ctx context.Context, numWorkers int, c Config) error {
+	workers, err := h.newWorkers(numWorkers)
+	if err != nil {
+		return err
+	}
+
+	/*
```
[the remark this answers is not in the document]
gdbelvin: Deleted it. 
I was experimenting with the use of multiple progress bars, but found them to be ultimately unhelpful.

core/client/hammer/hammer.go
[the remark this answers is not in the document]
gdbelvin: done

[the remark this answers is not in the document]
gdbelvin: Yes. Thanks for catching that.

gdbelvin: done

core/client/hammer/hammer.go
gdbelvin: Yes, that's correct. However, I'm removing the progress bar since 1000 progress bars aren't that user friendly ;-)

core/client/hammer/hammer.go
gdbelvin: Splitting the loops allows the changes to be efficiently batched. 
If there was only one loop, we would be waiting for a new map revision with one one change in it.

core/client/hammer/hammer.go
gdbelvin: done

core/client/hammer/hammer.go
gdbelvin: done

core/client/hammer/hammer.go
gdbelvin: done

core/client/hammer/hammer.go
gdbelvin: switching to batch size makes more sense. Thanks for spotting this.

core/client/hammer/hammer.go
gdbelvin: done

core/client/hammer/hammer.go
gdbelvin: done

core/client/hammer/hammer.go
gdbelvin: Removed comment.

core/client/hammer/hammer.go
gdbelvin: done

core/client/hammer/hammer.go
gdbelvin: I've removed the progress bar and this import.

core/client/hammer/hammer.go
gdbelvin: The purpose of the new line is to separate the dots between the request types.

`executeRequests` blocks until the workers have finished. 
Are you sure that it will be an issue?

pavelkalinnikov: Oh, it blocks indeed, sorry.

Hm, so if I understand correctly, the hammer will do multiple types of requests, but not simultaneously: first all batch writes, then all single writes, then all reads. Right? In this case I don't see why we can't have a slightly simpler hammer that accepts only one operation type (instead of a comma-separated list), and just run this binary multiple times. (there are couple more advantages of that, see my previous review messages)

PR #995, opened 2018-06-20
Regenerate kubernetes configs with kompose

gdbelvin:
Also propogates the --auth-type=insecure-fake flag from
the kubernetes config to the docker-compose config.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -119,6 +119,7 @@ services:
       - --map-url=map-server:8090
       - --tls-key=/kt/server.key
       - --tls-cert=/kt/server.crt
+      - --auth-type=insecure-fake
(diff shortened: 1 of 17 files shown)


docker-compose.yml
```diff
@@ -119,6 +119,7 @@ services:
       - --map-url=map-server:8090
       - --tls-key=/kt/server.key
       - --tls-cert=/kt/server.crt
+      - --auth-type=insecure-fake
```
Martin2112: Should there be a scary warning somewhere?

gdbelvin: Not particularly as this is the deployment for testing

Martin2112: Not sure people who check this code out will know that.

PR #996, opened 2018-06-20
Export Prometheus metrics to StackDriver

### LINKED ISSUE #385: Add performance tracing
Enable gRPC tracing and expose over a remote interface to track the performance and latency of requests.   

Useful libraries
- https://godoc.org/golang.org/x/net/trace
- UnaryInterceptor
- expvar

The goal is to enable the following kinds of stats:
- RPC counters for each method, total, per second 
- Latency for each method, last 10 requests, longest 10 requests, etc. 
- Latency counters on backend method calls. 
- Uptime, memory, cpu usage

### LINKED ISSUE #942: Send logging to StackDriver

gdbelvin:
Closes #942 
Closes #385


gdbelvin:
Screenshot of the new stack driver dashboard :-)

screen shot 2018-06-20 at 18 15 19

deploy/kubernetes/sequencer-deployment.yaml
```diff
@@ -27,9 +27,19 @@ spec:
         - --alsologtostderr
         - --v=5
         image: us.gcr.io/key-transparency/keytransparency-sequencer:latest
+        livenessProbe:
```
AlCutter: Would this capture what you're trying to do a bit more concisely?
 ```
livenessProbe:
          httpGet:
            path: /metrics
            port: 8081
```

File: deploy/kubernetes/prometheus-to-sd-service.yaml [hunk 2]
```diff
@@ -0,0 +1,19 @@
+apiVersion: v1
```
AlCutter: Could you help me understand the reason for the service (and corresponding `ports:` on the deploy?
My understanding is the prom-to-sd is active (i.e. scrapes metrics and then pushes to stackdriver APIs), is that not the case, or is there another reason you need the service?

gdbelvin: Woops, prometheus-to-sd job doesn't need a port.

deploy/kubernetes/sequencer-deployment.yaml
gdbelvin: Yes, that's probably the more efficient, kubernets specific, way of doing it. 
However, in an effort to only maintain one set of deployment configs, I'm auto generating the kubernetes configs with [`kompose`](https://github.com/google/keytransparency/blob/master/deploy/kubernetes/gen.go) from the `docker-compose.yml` file, and docker-compose uses a `cmd` format for liveness checks.

deploy/kubernetes/prometheus-to-sd-deployment.yaml
```diff
@@ -0,0 +1,38 @@
...
+    metadata:
+      creationTimestamp: null
+      labels:
+        io.kompose.service: prometheus-to-sd
+    spec:
+      containers:
+      - command:
+        - /monitor
+        - --stackdriver-prefix=custom.googleapis.com
+        - --source=logserver:http://log-server:8091/metrics
```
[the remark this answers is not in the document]
gdbelvin: Thanks for spotting this.  You make an excellent point. 

docker-compose can't support the "sidecar" / multiple containers-per-pod model that kubernets supports (and prometheus-to-sd) assumes. As long as there's only one replica (the current state) it should be fine. 

I'll make a note of it in the docker-compose file, and do the proper sidecar setup in kubernetes.

deploy/kubernetes/sequencer-deployment.yaml
AlCutter: Makes sense, though I suspect you're going to have to diverge at some point for a more nuanced kubernetes config, but no need to burden yourself with that now!

deploy/kubernetes/prometheus-to-sd-deployment.yaml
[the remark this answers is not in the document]
AlCutter: oke doke, LGTM otherwise - I'll approve now.

deploy/kubernetes/sequencer-deployment.yaml
gdbelvin: Yep. I'll be diverging now. 

docker-compose doesn't support the side-car strategy that prometheus-to-sd requires.

AlCutter: :)

AlCutter: Will you be changing the `livenessProbe` too then?

deploy/kubernetes/sequencer-service.yaml
```diff
@@ -13,6 +13,9 @@ spec:
   - name: "8080"
     port: 8080
     targetPort: 8080
+  - name: "8081"
```
AlCutter: Do you need to export the metrics port on the service too?

deploy/kubernetes/sequencer-deployment.yaml
gdbelvin: done

deploy/kubernetes/sequencer-service.yaml
gdbelvin: I think I did?  `sequencer-deployment.yaml` has a new `- containerPort: 8081` line.

PR #997, opened 2018-06-20
Serve metrics on http addr

gdbelvin:


cmd/keytransparency-sequencer/server.go
```diff
@@ -16,6 +16,7 @@ package main
 
 import (
 	"flag"
+	"log"
```
Martin2112: Should be glog, which is already imported?

gdbelvin: done

PR #998, opened 2018-06-21
Add errors to sequencer failed metric.

gdbelvin:
Makes monitoring easier by collecting statistics on each error string.


gdbelvin:
Is this a good idea?

Martin2112:
Probably better to use status codes as strings can change over time.

PR #999, opened 2018-06-21
Add batch-size flag and metric

gdbelvin:
Control the maximum number of mutations the Key Transparency sequencer processes per map revision.

Also adds visibility into the number of mutations the sequencer is attempting to process during each map revision.


gdbelvin:
This change seems to have resolved the errors:

- SetLeaves(): CalculateRoot(): Error 1040: Too many connections
- SetLeaves(): CalculateRoot(): Error 1062: Duplicate entry '3584314497260755511-@-25483' for key 'PRIMARY'"

PR #1006, opened 2018-06-26
Fix queue deadlock

gdbelvin:
Use a slightly simpler loop for sending multiple batches.

The previous implementation used a more chanel with a small buffer.
This worked as long as the select loop never chooses min.C when it
could have selected the more channel.

### DIFF
--- a/impl/sql/mutationstorage/queue.go
+++ b/impl/sql/mutationstorage/queue.go
@@ -111,6 +111,7 @@ func (r *Receiver) run(ctx context.Context, last time.Time) {
 	for {
 		select {
 		case <-r.ticker.C:
+			// We will be overdue for an epoch soon.
 			if _, err := r.sendMultiBatch(ctx, 1, int(r.opts.MaxBatchSize)); err != nil {
 				glog.Errorf("minTick: sendMultiBatch(): %v", err)
 			}
@@ -132,6 +133,9 @@ func (r *Receiver) run(ctx context.Context, last time.Time) {
 func (r *Receiver) sendMultiBatch(ctx context.Context, minMsgs, maxMsgs int) (int, error) {
 	var total int
 	var err error
+	if minMsgs > maxMsgs {
+		minMsgs = maxMsgs
+	}
 	for sent := maxMsgs; sent >= maxMsgs; {
 		sent, err = r.sendBatch(ctx, minMsgs, maxMsgs)
 		if err != nil {
--- a/impl/sql/mutationstorage/queue.go
+++ b/impl/sql/mutationstorage/queue.go
@@ -51,7 +51,6 @@ func (m *Mutations) NewReceiver(ctx context.Context, last time.Time, domainID st
 		store:       m,
 		domainID:    domainID,
 		opts:        rOpts,
-		more:        make(chan time.Time, 1),
 		ticker:      time.NewTicker(rOpts.Period),
 		maxTicker:   time.NewTicker(rOpts.MaxPeriod),
 		done:        make(chan interface{}),
@@ -69,7 +68,6 @@ type Receiver struct {
 	domainID    string
 	recieveFunc mutator.ReceiveFunc
 	opts        mutator.ReceiverOptions
-	more        chan time.Time
 	ticker      *time.Ticker
 	maxTicker   *time.Ticker
 	done        chan interface{}
@@ -84,11 +82,14 @@ func (r *Receiver) Close() {
 
 // FlushN verifies that a minimum of n items are available to send, and sends them.
 func (r *Receiver) FlushN(ctx context.Context, n int) error {
-	sent := r.sendBatch(ctx, int32(n), r.opts.MaxBatchSize)
-	if sent < int32(n) {
+	sent, err := r.sendMultiBatch(ctx, n, int(r.opts.MaxBatchSize))
+	if err != nil {
+		return err
+	}
+	if sent < n {
 		// We could retry at this point, but because this queue is mysql based
 		// and deterministic, we know that waiting won't be very useful.
-		return fmt.Errorf("sendBatch(): %v, want >= %v", sent, n)
+		return fmt.Errorf("sendMultiBatch(): %v, want >= %v", sent, n)
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


impl/sql/mutationstorage/queue_test.go
```diff
@@ -131,48 +131,48 @@ func TestQueueSendBatch(t *testing.T) {
...
-			batchSize: 3,
-		},
-		{
-			description: "read rest",
-			updates: []*pb.EntryUpdate{
-				genUpdate(4),
-				genUpdate(5),
+			description: "read half and rest",
+			updates: [][]*pb.EntryUpdate{
+				[]*pb.EntryUpdate{
```
golangci: File is not gofmt-ed with -s

impl/sql/mutationstorage/queue.go
```diff
@@ -101,57 +102,70 @@ func (r *Receiver) run(ctx context.Context, last time.Time) {
 	}
 
 	if time.Since(last) > (r.opts.MaxPeriod - r.opts.Period) {
-		r.sendBatch(ctx, 0, r.opts.MaxBatchSize) // We will be overdue for an epoch soon.
+		if _, err := r.sendMultiBatch(ctx, 0, int(r.opts.MaxBatchSize)); err != nil {
+			glog.Errorf("firstTick: sendMultiBatch(): %v", err)
```
Martin2112: This error doesn't get handled. Happens several times below. I assume it's intentional but looks suspicious.

impl/sql/mutationstorage/queue.go
```diff
@@ -101,57 +102,70 @@ func (r *Receiver) run(ctx context.Context, last time.Time) {
...
+// This helps the queue catch up when there is high traffic, rather than waiting for the next
+// mintick to occur.
+func (r *Receiver) sendMultiBatch(ctx context.Context, minMsgs, maxMsgs int) (int, error) {
+	var total int
+	var err error
+	for sent := maxMsgs; sent >= maxMsgs; {
+		sent, err = r.sendBatch(ctx, minMsgs, maxMsgs)
+		total += sent
+		if err != nil {
+			return total, err
```
Martin2112: If it still works returning `0, err` then I'd do it. If it would break then don't do that :-).

impl/sql/mutationstorage/queue.go
```diff
@@ -101,57 +102,70 @@ func (r *Receiver) run(ctx context.Context, last time.Time) {
...
+	return total, nil
 }
 
 // sendBatch sends up to batchSize items to the receiver. Returns the number of sent items.
 // If the number of available items is < minBatch, 0 items are sent.
 // If the number of available items is > maxBatch only maxBatch items are sent.
-func (r *Receiver) sendBatch(ctx context.Context, minBatch, maxBatch int32) int32 {
-	ms, err := r.store.readQueue(ctx, r.domainID, maxBatch)
+func (r *Receiver) sendBatch(ctx context.Context, minBatch, maxBatch int) (int, error) {
+	ms, err := r.store.readQueue(ctx, r.domainID, int32(maxBatch))
```
Martin2112: Might be clearer to rename `ms` to `batch`.

impl/sql/mutationstorage/queue.go
```diff
@@ -101,57 +102,70 @@ func (r *Receiver) run(ctx context.Context, last time.Time) {
...
+	}
+}
+
+// sendMultiBatch will send multiple batches if the number of available messages > maxMsgs.
+// This helps the queue catch up when there is high traffic, rather than waiting for the next
+// mintick to occur.
+func (r *Receiver) sendMultiBatch(ctx context.Context, minMsgs, maxMsgs int) (int, error) {
+	var total int
+	var err error
+	for sent := maxMsgs; sent >= maxMsgs; {
```
Martin2112: If someone dumps a very large number of messages then it looks like this doesn't exit until they're all done. You might want to bound the number of batches per run.

impl/sql/mutationstorage/queue.go
```diff
@@ -101,57 +102,70 @@ func (r *Receiver) run(ctx context.Context, last time.Time) {
...
 		}
-		if count >= r.opts.MaxBatchSize {
-			// Continue sending until we drop below batch size.
-			r.more <- time.Now()
+	}
+}
+
+// sendMultiBatch will send multiple batches if the number of available messages > maxMsgs.
+// This helps the queue catch up when there is high traffic, rather than waiting for the next
+// mintick to occur.
```
phad: Please document what the returned int signifies.

impl/sql/mutationstorage/queue.go
```diff
@@ -101,57 +102,70 @@ func (r *Receiver) run(ctx context.Context, last time.Time) {
...
+func (r *Receiver) sendMultiBatch(ctx context.Context, minMsgs, maxMsgs int) (int, error) {
+	var total int
+	var err error
+	for sent := maxMsgs; sent >= maxMsgs; {
+		sent, err = r.sendBatch(ctx, minMsgs, maxMsgs)
+		total += sent
```
phad: It's a bit optimistic inc'ing total here I think .. if err != nil then you can't count this particular value of sent in your total.

impl/sql/mutationstorage/queue.go
```diff
@@ -101,57 +102,70 @@ func (r *Receiver) run(ctx context.Context, last time.Time) {
...
-	if int32(len(ms)) < minBatch {
-		return 0
+	if len(ms) < minBatch {
+		return 0, nil
 	}
 
 	if err := r.recieveFunc(ms); err != nil {
-		glog.Errorf("queue.SendBatch(): %v", err)
-		return 0
+		return 0, fmt.Errorf("queue.SendBatch(): %v", err)
```
phad: do you want to note in the error message that it's r.receiveFunc() that returned the non-nil err?

since you're in func queue.sendBatch() wouldn't you normally expect the caller to log a message saying that queue.sendBatch() had errored?

impl/sql/mutationstorage/queue.go
```diff
@@ -101,57 +102,70 @@ func (r *Receiver) run(ctx context.Context, last time.Time) {
...
 	}
 	// TODO(gbelvin): Do we need finer grained errors?
 	// We could put an ack'ed field in a QueueMessage object.
 	// But I don't think we need that level of granularity -- yet?
 
 	// Delete old messages.
 	if err := r.store.deleteMessages(ctx, r.domainID, ms); err != nil {
-		glog.Errorf("deleteQueueMessages(%v, len(ms): %v): %v", r.domainID, len(ms), err)
+		return 0, fmt.Errorf("deleteQueueMessages(%v, len(ms): %v): %v", r.domainID, len(ms), err)
```
phad: nit: it's deleteMessages now, not deleteQueueMessages.

impl/sql/mutationstorage/queue_test.go
```diff
@@ -131,48 +131,48 @@ func TestQueueSendBatch(t *testing.T) {
...
-			updates:     nil,
+			updates:     [][]*pb.EntryUpdate{{}},
 			batchSize:   10,
 		},
 	} {
-		wg := new(sync.WaitGroup)
-		wg.Add(1)
+		var wg sync.WaitGroup
+		wg.Add(len(tc.updates))
+		var i int
```
phad: Are you sure this should be init'd to zero outside the t.Run loop? It looks dubious, and maybe fragile if tests get omitted via a test filter.

I'd expect var i int (or just i :=0) to be just before r, ok := m.NewReceiver within the t.Run loop.

impl/sql/mutationstorage/queue.go
gdbelvin: It is indeed intentional. Run is at the top of it's call stack, so there's not really any other place to go from here.

impl/sql/mutationstorage/queue.go
gdbelvin: cool

impl/sql/mutationstorage/queue.go
gdbelvin: done

impl/sql/mutationstorage/queue.go
gdbelvin: I *think* I actually want this to run as long as there a lot of messages to process. The rest of the code in this file implements waiting semantics.

impl/sql/mutationstorage/queue.go
gdbelvin: done

impl/sql/mutationstorage/queue.go
gdbelvin: done

impl/sql/mutationstorage/queue.go
gdbelvin: yep

impl/sql/mutationstorage/queue.go
gdbelvin: thanks

impl/sql/mutationstorage/queue_test.go
gdbelvin: woops, yes, this should be inside t.Run

gdbelvin:
PTAL

impl/sql/mutationstorage/queue.go
Martin2112: If there's 10,000,000 though? You could swamp the rest of the system. Though this might not be the right place to do that coordination.

PR #1013, opened 2018-07-06
Add golangci config

gdbelvin:
In the process also fix

* Capitalized errors
* Extra conversions
* Unchecked errors

### DIFF
--- a/core/crypto/vrf/p256/p256.go
+++ b/core/crypto/vrf/p256/p256.go
@@ -74,7 +74,8 @@ func GenerateKey() (vrf.PrivateKey, vrf.PublicKey) {
 		return nil, nil
 	}
 
-	return &PrivateKey{key}, &PublicKey{&key.PublicKey}
+	return &PrivateKey{PrivateKey: key},
+		&PublicKey{PublicKey: &key.PublicKey}
 }
 
 // H1 hashes m to a curve point
@@ -243,7 +244,9 @@ func NewVRFSigner(key *ecdsa.PrivateKey) (vrf.PrivateKey, error) {
 	if !curve.IsOnCurve(key.X, key.Y) {
 		return nil, ErrPointNotOnCurve
 	}
-	return &PrivateKey{key}, nil
+	return &PrivateKey{
+		PrivateKey: key,
+	}, nil
 }
 
 // Public returns the corresponding public key as bytes.
@@ -259,7 +262,9 @@ func NewVRFVerifier(pubkey *ecdsa.PublicKey) (vrf.PublicKey, error) {
 	if !curve.IsOnCurve(pubkey.X, pubkey.Y) {
 		return nil, ErrPointNotOnCurve
 	}
-	return &PublicKey{pubkey}, nil
+	return &PublicKey{
+		PublicKey: pubkey,
+	}, nil
 }
 
 // NewVRFSignerFromPEM creates a vrf private key from a PEM data structure.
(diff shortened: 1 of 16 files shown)


core/crypto/vrf/p256/p256.go
```diff
@@ -74,7 +74,8 @@ func GenerateKey() (vrf.PrivateKey, vrf.PublicKey) {
 		return nil, nil
 	}
 
-	return &PrivateKey{key}, &PublicKey{&key.PublicKey}
+	return &PrivateKey{PrivateKey: key},
+		&PublicKey{PublicKey: &key.PublicKey}
```
taknira: These probably don't need to be on separate lines.  Even with the field names added, the return statement isn't long enough to warrant a line split.

core/crypto/vrf/p256/p256.go
```diff
@@ -243,7 +244,9 @@ func NewVRFSigner(key *ecdsa.PrivateKey) (vrf.PrivateKey, error) {
 	if !curve.IsOnCurve(key.X, key.Y) {
 		return nil, ErrPointNotOnCurve
 	}
-	return &PrivateKey{key}, nil
+	return &PrivateKey{
+		PrivateKey: key,
+	}, nil
```
taknira: Similarly here, the whole return statement can probably be on one line - the PrivateKey struct is only little!

return &PrivateKey{PrivateKey: key}, nil

is still pretty readable.

core/crypto/vrf/p256/p256.go
```diff
@@ -259,7 +262,9 @@ func NewVRFVerifier(pubkey *ecdsa.PublicKey) (vrf.PublicKey, error) {
 	if !curve.IsOnCurve(pubkey.X, pubkey.Y) {
 		return nil, ErrPointNotOnCurve
 	}
-	return &PublicKey{pubkey}, nil
+	return &PublicKey{
+		PublicKey: pubkey,
+	}, nil
```
taknira: And here.

core/integration/client_tests.go
```diff
@@ -65,7 +65,9 @@ var (
 
 // TestEmptyGetAndUpdate verifies set/get semantics.
 func TestEmptyGetAndUpdate(ctx context.Context, env *Env, t *testing.T) {
-	signature.RegisterStandardKeyTypes()
+	if _, err := signature.RegisterStandardKeyTypes(); err != nil {
+		t.Errorf("RegisterStandaredKeyTypes(): %v", err)
```
taknira: If RegisterStandardKeyTypes() fails, do you intend to continue with the rest of the test?  Just checking, as t.Errorf will log the error but then let the rest of the test continue.

taknira: Typo in the logged function name - remove the 'e' from 'Standared'

impl/authentication/fake.go
```diff
@@ -41,7 +41,9 @@ func FakeAuthFunc(ctx context.Context) (context.Context, error) {
 
 // GetFakeCredential returns fake PerRPCCredentials
 func GetFakeCredential(userID string) credentials.PerRPCCredentials {
-	return fakeCredential{userID}
+	return fakeCredential{
+		userID: userID,
+	}
```
taknira: Again, probably doesn't need to be split across multiple lines.

core/crypto/vrf/p256/p256.go
gdbelvin: done

core/crypto/vrf/p256/p256.go
gdbelvin: done

core/crypto/vrf/p256/p256.go
gdbelvin: done

core/integration/client_tests.go
gdbelvin: nope, it should be fatal

gdbelvin: fixed

PR #1018, opened 2018-07-12
Restrict integration.Env to only grpc clients

gdbelvin:
This PR removes `Env.Receiver` which was previously being used for fine-grained control over what mutations were part of a batch. The presence of `Receiver` in `Env` prevented integration tests from being run against live servers.

This PR replace `Receiver` with 
a) Real sequencing behavior when the exact batch composition is not being tested.
b) Calls to the new`sequencer.Server.CreateEpcoh` API to create new epochs directly when exact batching is needed.

### DIFF
--- a/core/integration/client_tests.go
+++ b/core/integration/client_tests.go
@@ -287,7 +287,7 @@ func (env *Env) setupHistory(ctx context.Context, domain *pb.Domain, userID stri
 				DomainId: domain.DomainId,
 				Messages: []*pb.EntryUpdate{msg.EntryUpdate},
 			}); err != nil {
-				return fmt.Errorf("create empty epoch: %v", err)
+				return fmt.Errorf("create epoch: %v", err)
 			}
 		} else {
 			// Create an empty epoch.
--- a/core/integration/client_tests.go
+++ b/core/integration/client_tests.go
@@ -251,6 +251,13 @@ func (env *Env) setupHistory(ctx context.Context, domain *pb.Domain, userID stri
 	// Log TreeSize:  2  3  4  5  6  7  8  9  10 11 ...
 	// Note that profile 5 is submitted twice by the user to test that
 	// filtering case.
+
+	// Create an empty epoch.
+	if _, err := env.Sequencer.CreateEpoch(ctx, &spb.CreateEpochRequest{
+		DomainId: domain.DomainId,
+	}); err != nil {
+		return fmt.Errorf("create empty epoch: %v", err)
+	}
 	for _, p := range [][]byte{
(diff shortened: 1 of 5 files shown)


core/integration/client_tests.go
```diff
@@ -274,21 +278,23 @@ func (env *Env) setupHistory(ctx context.Context, domain *pb.Domain, userID stri
...
+				return fmt.Errorf("SerializeAndSign(): %v", err)
 			}
 
-			if _, err := env.Client.WaitForUserUpdate(cctx, m); err != nil {
-				return fmt.Errorf("WaitForUserUpdate(%v): %v, want nil", m, err)
+			if _, err := env.Sequencer.CreateEpoch(ctx, &spb.CreateEpochRequest{
+				DomainId: domain.DomainId,
+				Messages: []*pb.EntryUpdate{msg.EntryUpdate},
+			}); err != nil {
+				return fmt.Errorf("create empty epoch: %v", err)
```
jtoohill: This epoch isn't empty, maybe "CreateEpoch(): %v"

core/integration/client_tests.go
```diff
@@ -274,21 +278,23 @@ func (env *Env) setupHistory(ctx context.Context, domain *pb.Domain, userID stri
...
 			}
 		} else {
 			// Create an empty epoch.
-			if err := env.Receiver.FlushN(ctx, 0); err != nil {
-				return fmt.Errorf("FlushN(0): %v", err)
+			if _, err := env.Sequencer.CreateEpoch(ctx, &spb.CreateEpochRequest{
+				DomainId: domain.DomainId,
+			}); err != nil {
+				return fmt.Errorf("create empty epoch: %v", err)
```
jtoohill: ditto

core/integration/monitor_tests.go
```diff
@@ -97,24 +99,27 @@ func TestMonitor(ctx context.Context, env *Env, t *testing.T) {
...
-		if err := env.Receiver.FlushN(ctx, len(e.userUpdates)); err != nil {
-			t.Fatalf("FlushN(%v): %v", len(e.userUpdates), err)
-		}
-		if err := env.Client.WaitForRevision(ctx, e.epoch); err != nil {
-			t.Fatalf("WaitForRevision(): %v", err)
+		if _, err := env.Sequencer.CreateEpoch(ctx, &spb.CreateEpochRequest{
+			DomainId: env.Domain.DomainId,
+			Messages: msgs,
+		}); err != nil {
+			t.Fatalf("create empty epoch: %v", err)
```
jtoohill: ditto

PR #1019, opened 2018-07-15
Cleanup domains after integration test

gdbelvin:
At the end of an integration test, soft-delete the domain and the corresponding Trillian trees. 

TODO: 
- Run doman and tree garbage collection.

### DIFF
--- a/core/adminserver/admin_server.go
+++ b/core/adminserver/admin_server.go
@@ -346,21 +346,5 @@ func (s *Server) DeleteDomain(ctx context.Context, in *pb.DeleteDomainRequest) (
 
 // UndeleteDomain reactivates a deleted domain - provided that UndeleteDomain is called sufficiently soon after DeleteDomain.
 func (s *Server) UndeleteDomain(ctx context.Context, in *pb.UndeleteDomainRequest) (*google_protobuf.Empty, error) {
-	d, err := s.GetDomain(ctx, &pb.GetDomainRequest{DomainId: in.GetDomainId(), ShowDeleted: true})
-	if err != nil {
-		return nil, err
-	}
-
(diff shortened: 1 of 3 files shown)


Martin2112:
Can we downgrade code coverage severity so it doesn't look like the build has failed?

core/adminserver/admin_server.go
```diff
@@ -325,16 +325,42 @@ func (s *Server) initialize(ctx context.Context, logTree, mapTree *tpb.Tree) err
...
+		return nil, err
+	}
+
 	if err := s.domains.SetDelete(ctx, in.GetDomainId(), false); err != nil {
 		return nil, err
 	}
+
+	_, delLogErr := s.logAdmin.UndeleteTree(ctx, &tpb.UndeleteTreeRequest{TreeId: d.Log.TreeId})
+	_, delMapErr := s.mapAdmin.UndeleteTree(ctx, &tpb.UndeleteTreeRequest{TreeId: d.Map.TreeId})
+	if delLogErr != nil || delMapErr != nil {
```
Martin2112: One of these undeletions could fail, resulting in a weird state. Also SetDelete false is called before the trees are undeleted so they could both fail to undelete. Does that also create a bad state?

gdbelvin: If they failed, it would create a bad state. Should I try to undelete at that point?

gdbelvin:
The code coverage warning is just making sure that code coverage doesn't diminish. So while I agree that it is annoying, I like having it to motivate more code coverage :-)

Martin2112:
The code coverage thing is going to delay reviews because people will assume the build isn't working. I think it's a bad idea to make it fail the overall status.

PR #1022, opened 2018-07-23
Remove service-key

gdbelvin:
Fixes a bug where the client was not sending oauth user credentials.

Post #975, client creds do not need to be passed with each RPC, and as
such, an authenticated, yet anonymous mode of authentication is no
longer needed, we can just use unauthenticated calls for the public api.

If we need to bring this back at a later point for spam reasons, we can.
In the meantime, deleted code is best code.


davidbstein:
I don't have a great working knowledge of what's needed for fake user creds, but this LGTM and works on my machine.

PR #1028, opened 2018-08-22
Rewrite VRF test to use generated testdata

jtoohill:
Changes VRF verification tests to use common generated testdata. This makes it easier for other projects that want to implement VRF verification to keep up-to-date with keytransparency's VRF implementation by using the same testdata.


core/crypto/vrf/p256/p256_test.go
```diff
@@ -179,40 +186,37 @@ func TestVRF(t *testing.T) {
 }
 
 func TestProofToHash(t *testing.T) {
```
gdbelvin: Please add a note about how to regenerate the test data.

jtoohill: Done.

PR #1032, opened 2018-09-05
Add CODEOWNERS file

gdbelvin:
A `CODEOWNERS` file will enable automatic PR reviewer assignment.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,16 @@
+# See https://help.github.com/articles/about-codeowners/
+# for more info about CODEOWNERS file
+
+# It uses the same pattern rule for gitignore file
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -0,0 +1,16 @@
...
+
+# Default
+*				@gdbelvin
+
+/cmd				@gdbelvin
+/core				@gdbelvin
+/core/crypto			@gdbelvin
+/deploy				@gdbelvin
+/docs				@gdbelvin
+/impl				@gdbelvin
```
RJPercival: Why not let the default case cover these?

gdbelvin: reminding myself of the right syntax, but that isn't a very good reason. Removed.

PR #1036, opened 2018-09-17
Simplify s[:] to s where s is a slice

ludweeg:
Found using https://go-critic.github.io/overview#unslice-ref


Quasilyte:
@gdbelvin, it's not a good CI linter yet, more like an audit tool that requires manual use.
Some checks can be integrated into gometalinter and golangci-lint in future, but we're not there yet.

gdbelvin:
Well thanks for the contributions and tool improvements.

PR #1044, opened 2018-09-20
Retry-able Publish SignedMapRoot 

gdbelvin:
It is possible for the sequencer to die after line 165, after creating a new SignedMapRoot (SMR) and before publishing the SMR in the Log of Signed Map Roots (line 164).

This PR factors out the logic of publishing all known SMRs into a separate function which can be called  on a timer, in addition to on-demand.  

Part of the retry-able batches, #1037


core/sequencer/server.go
```diff
@@ -153,22 +153,65 @@ func (s *Server) CreateEpoch(ctx context.Context, in *spb.CreateEpochRequest) (*
...
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "VerifySignedMapRoot(): %v", err)
+		}
+		if err := logClient.AddSequencedLeaf(ctx, rawMapRoot.GetMapRoot(), int64(mapRoot.Revision)); err != nil {
+			glog.Errorf("AddSequencedLeaf(logID: %v, rev: %v): %v", logClient.LogID, mapRoot.Revision, err)
+			return nil, err
+		}
+
+		// TODO(gbelvin): Remove wait when batching boundaries are deterministic
+		if rev == latestMapRoot.Revision {
```
pavelkalinnikov: Could the be placed after the loop?

```
if logRoot.TreeSize - 1 <= latestMapRoot.Revision {
  ... wait ...  
}
```

core/sequencer/server.go
```diff
@@ -153,22 +153,65 @@ func (s *Server) CreateEpoch(ctx context.Context, in *spb.CreateEpochRequest) (*
 		return nil, status.Errorf(codes.Internal, "mutations.WriteBatch(): %v", err)
 	}
 
-	// TODO(gbelvin): Store and track trustedRoot.
-	trustedRoot := types.LogRootV1{} // Automatically trust the first observed log root.
+	mutationCount.Add(float64(len(msgs)), domainID)
+	glog.Infof("CreatedEpoch: rev: %v with %v mutations, root: %x", mapRoot.Revision, len(msgs), mapRoot.RootHash)
```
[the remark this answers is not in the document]
gdbelvin: `%v` works just as well doesn't it?

core/sequencer/server.go
gdbelvin: The intent is to wait for the last item to be sequenced. 
I can pull it out of the loop, but that would mess up a bunch of the variable scoping inside the for loop. I think this a bit cleaner. What do you think?

core/sequencer/server.go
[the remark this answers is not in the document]
pavelkalinnikov: It does, but `%d` is more specific. It's a matter of style really, not sure which is preferred by this repo. Do you not use `%d` in other places?

core/sequencer/server.go
pavelkalinnikov: Oh, just noticed it uses some local vars of the loop. Yeah, I think it's better to leave it as is.

pavelkalinnikov: OTOH, shouldn't `rawMapRoot` be equal to `latestMapRoot` on the last iteration?

core/sequencer/server.go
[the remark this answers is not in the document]
gdbelvin: I prefer `%v` in general because it leaves less to change later if the types change.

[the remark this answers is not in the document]
pavelkalinnikov: SGTM.

core/sequencer/server.go
gdbelvin: Good point. Fixed.

core/sequencer/server.go
```diff
@@ -195,22 +195,66 @@ func (s *Server) CreateEpoch(ctx context.Context, in *spb.CreateEpochRequest) (*
...
+	if err != nil {
+		return nil, status.Errorf(status.Code(err), "GetSignedMapRoot(%v): %v", mapClient.MapID, err)
+	}
+	latestMapRoot, err := mapClient.VerifySignedMapRoot(rootResp.GetMapRoot())
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "VerifySignedMapRoot(%v): %v", mapClient.MapID, err)
+	}
+
+	// Add all unpublished map roots to the log.
+	for rev := logRoot.TreeSize - 1; rev <= latestMapRoot.Revision; rev++ {
```
pavelkalinnikov: Does this loop always have at least 1 iteration? What will happen if the `latestMapRoot` stays the same between 2 `PublishBatch` runs?

gdbelvin: No, `TreeSize - 1` could == `latestMapRoot.Revision`. 
In that case, `WaitForInclusion` doesn't need to wait. It will just verify the inclusion proof of `latestMapRoot`

Issue #1047, opened 2018-09-25
Send remaining items in queue after a timeout


gdbelvin:
The current batching control uses a `min_batch` size to prevent overly small batches. 
This can result in queue items that take longer than needed before being incorporated into a batch. 

Replace `min_batch` size with a `max_delay` so we can establish an upper bound on item latency in the queue.

PR #1048, opened 2018-09-25
Scale mutation queue with multiple shards

gdbelvin:
Tables with an incrementing value in the lower bits of the primary key are notoriously hard to scale because all the writes will go to a single server.  This PR allows the queue to scale by adding write support for multiple queues. 

Depends on: read by high/low watermarks #1045 
Part of: retry-able mutations #1044


impl/sql/mutationstorage/queue.go
```diff
@@ -17,40 +17,107 @@ package mutationstorage
...
+	var shardIDs []int64
+	rows, err := m.db.QueryContext(ctx,
+		`SELECT ShardID from Shards WHERE DomainID = ? AND Enabled = ?;`,
+		domainID, true)
+	if err != nil {
+		return 0, err
+	}
+	for rows.Next() {
+		var shardID int64
+		rows.Scan(&shardID)
```
golangcibot: G104: Errors unhandled.

impl/sql/mutationstorage/queue.go
```diff
@@ -17,40 +17,107 @@ package mutationstorage
...
+		watermarks[shardID] = watermark
+	}
+	if err := rows.Err(); err != nil {
+		return nil, err
+	}
+	return watermarks, nil
+}
+
+// ReadQueue reads all mutations that are still in the queue up to batchSize.
+func (m *Mutations) ReadQueue(ctx context.Context, domainID string, shardID, low, high int64) ([]*mutator.QueueMessage, error) {
```
golangcibot: line is 128 characters

impl/sql/mutationstorage/queue_test.go
```diff
@@ -19,12 +19,58 @@ import (
...
 )
 
+func TestRandShard(t *testing.T) {
+	ctx := context.Background()
+	db := newDB(t)
+	m, err := New(db)
+	if err != nil {
+		t.Fatalf("Failed to create Mutations: %v", err)
+	}
+	domainID := "foo"
```
golangcibot: string `foo` has 3 occurrences, make it a constant

impl/sql/mutationstorage/queue.go
```diff
@@ -29,29 +30,102 @@ import (
...
+	var shardIDs []int64
+	rows, err := m.db.QueryContext(ctx,
+		`SELECT ShardID from Shards WHERE DomainID = ? AND Enabled = ?;`,
+		domainID, true)
+	if err != nil {
+		return 0, err
+	}
+	for rows.Next() {
+		var shardID int64
+		rows.Scan(&shardID)
```
golangcibot: G104: Errors unhandled.

impl/sql/mutationstorage/queue.go
```diff
@@ -29,29 +30,104 @@ import (
...
+			`INSERT INTO Shards (DomainID, ShardID, Enabled)  Values(?, ?, ?);`,
+			domainID, shardID, true); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// randShard returns a random, enabled shard for domainID.
+func (m *Mutations) randShard(ctx context.Context, domainID string) (int64, error) {
```
golangcibot: 47-73 lines are duplicate of `impl/sql/mutationstorage/queue.go:92-118`

impl/sql/mutationstorage/queue.go
```diff
@@ -29,29 +30,104 @@ import (
...
 		return err
 	}
 	// TODO(gbelvin): Implement retry with backoff for retryable errors if
 	// we get timestamp contention.
-	return m.send(ctx, domainID, mData, time.Now())
+	return m.send(ctx, domainID, shardID, mData, time.Now())
+}
+
+// randomShard returns a random shard from the list of active shards for domainID.
+func (m *Mutations) randomShard(ctx context.Context, domainID string) (int64, error) {
```
golangcibot: 92-118 lines are duplicate of `impl/sql/mutationstorage/queue.go:47-73`

gdbelvin:
Updated and nice and green. PTAL. 
Happy to walk through the code with you

impl/sql/mutationstorage/queue.go
```diff
@@ -29,29 +30,74 @@ import (
...
 		return err
 	}
 
 	var maxTime int64
 	if err := tx.QueryRowContext(ctx,
-		`SELECT COALESCE(MAX(Time), 0) FROM Queue WHERE DomainID = ?;`,
-		domainID).Scan(&maxTime); err != nil {
+		`SELECT COALESCE(MAX(Time), 0) FROM Queue WHERE DomainID = ? AND ShardID = ?;`,
+		domainID, shardID).Scan(&maxTime); err != nil {
 		if rollbackErr := tx.Rollback(); rollbackErr != nil {
```
pavelkalinnikov: nit (could be a TODO + follow-up PR): We could `defer` a function right after `BeginTx` which would call `Rollback()` in case of an error. This would allow you avoid copy-paste. 

How about?
```
func ... (...) (ret error) {
  tx, err := ...
  ...
  defer func() {
    if ret != nil {
      if err := tx.Rollback(...); err != nil {
        ret = status.Errorf(...)
      }
    }
  }()
}
```
... and then in this line (and similar ones below) just `return err`.

impl/sql/mutationstorage/queue.go
```diff
@@ -80,27 +126,37 @@ func (m *Mutations) send(ctx context.Context, domainID string, mData []byte, ts
...
-// HighWatermark returns the highest timestamp in the mutations table.
-func (m *Mutations) HighWatermark(ctx context.Context, domainID string) (int64, error) {
-	var watermark int64
-	if err := m.db.QueryRowContext(ctx,
-		`SELECT Time FROM Queue WHERE DomainID = ? ORDER BY Time DESC LIMIT 1;`,
-		domainID).Scan(&watermark); err == sql.ErrNoRows {
-		return 0, nil
-	} else if err != nil {
-		return 0, err
+// HighWatermarks returns the highest timestamp in the mutations table.
```
pavelkalinnikov: Please account for shards in this comment.

impl/sql/mutationstorage/queue.go
```diff
@@ -29,29 +30,74 @@ import (
...
+func (m *Mutations) randShard(ctx context.Context, domainID string) (int64, error) {
+	// TODO(gbelvin): Cache these results.
+	var shardIDs []int64
+	rows, err := m.db.QueryContext(ctx,
+		`SELECT ShardID from Shards WHERE DomainID = ? AND Enabled = ?;`,
+		domainID, true)
+	if err != nil {
+		return 0, err
+	}
+	for rows.Next() {
```
pavelkalinnikov: You should probably `defer rows.Close()`, to cover the case when `rows.Scan` returns an error.

impl/sql/mutationstorage/queue.go
```diff
@@ -29,29 +30,74 @@ import (
...
 	tx, err := m.db.BeginTx(ctx, &sql.TxOptions{Isolation: sql.LevelSerializable})
 	if err != nil {
 		return err
 	}
 
 	var maxTime int64
 	if err := tx.QueryRowContext(ctx,
-		`SELECT COALESCE(MAX(Time), 0) FROM Queue WHERE DomainID = ?;`,
-		domainID).Scan(&maxTime); err != nil {
+		`SELECT COALESCE(MAX(Time), 0) FROM Queue WHERE DomainID = ? AND ShardID = ?;`,
```
pavelkalinnikov: Can `Time` be nil?

impl/sql/mutationstorage/queue.go
```diff
@@ -80,27 +126,37 @@ func (m *Mutations) send(ctx context.Context, domainID string, mData []byte, ts
...
-		`SELECT Time FROM Queue WHERE DomainID = ? ORDER BY Time DESC LIMIT 1;`,
-		domainID).Scan(&watermark); err == sql.ErrNoRows {
-		return 0, nil
-	} else if err != nil {
-		return 0, err
+// HighWatermarks returns the highest timestamp in the mutations table.
+func (m *Mutations) HighWatermarks(ctx context.Context, domainID string) (map[int64]int64, error) {
+	watermarks := make(map[int64]int64)
+	rows, err := m.db.QueryContext(ctx,
+		`SELECT ShardID, Max(Time) FROM Queue WHERE DomainID = ? GROUP BY ShardID;`,
```
pavelkalinnikov: One of the functions above uses `COALESCE` to cover `Time == NULL` case. Is it needed here as well, or not needed above?

impl/sql/mutationstorage/queue.go
```diff
@@ -80,27 +126,37 @@ func (m *Mutations) send(ctx context.Context, domainID string, mData []byte, ts
...
+	if err := rows.Err(); err != nil {
+		return nil, err
 	}
-	return watermark, nil
+	return watermarks, nil
 }
 
-// ReadQueue reads all mutations that are still in the queue up to batchSize.
-func (m *Mutations) ReadQueue(ctx context.Context, domainID string, low, high int64) ([]*mutator.QueueMessage, error) {
+// ReadQueue reads all mutations that are still in the queue.
```
pavelkalinnikov: nit: Please comment it more accurately. E.g.:
```
// ReadQueue reads all mutations between low and high watermarks from the
// specified queue shard.
```

impl/sql/mutationstorage/queue_test.go
```diff
@@ -19,25 +19,73 @@ import (
...
+	ctx := context.Background()
+	db := newDB(t)
+	m, err := New(db)
+	if err != nil {
+		t.Fatalf("Failed to create Mutations: %v", err)
+	}
+
+	for _, tc := range []struct {
+		desc       string
+		send       bool
```
pavelkalinnikov: Consider making this `send []int64` and trying some other cases (e.g. when len=1).

impl/sql/mutationstorage/queue_test.go
```diff
@@ -19,25 +19,73 @@ import (
...
+		wantShards map[int64]bool
+	}{
+		{desc: "no rows", wantCode: codes.NotFound, wantShards: map[int64]bool{}},
+		{desc: "second", send: true, wantShards: map[int64]bool{
+			1: true,
+			2: true,
+			3: true,
+		}},
+	} {
+		if tc.send {
```
pavelkalinnikov: ... and:
```
if tc.send != nil {
  AddShards(..., tc.send...)
}
```

impl/sql/mutationstorage/queue_test.go
```diff
@@ -19,25 +19,73 @@ import (
...
+			if err := m.AddShards(ctx, domainID, 1, 2, 3); err != nil {
+				t.Fatalf("AddShards(): %v", err)
+			}
+		}
+		shards := make(map[int64]bool)
+		for i := 0; i < 20; i++ {
```
pavelkalinnikov: 20 could be, say, `10 * len(tc.send)`

impl/sql/mutationstorage/queue_test.go
```diff
@@ -65,31 +113,66 @@ func TestWatermark(t *testing.T) {
...
-				t.Fatalf("send(): %v", err)
+		for shardID, ts := range tc.send {
+			if err := m.send(ctx, domainID, shardID, []byte("mutation"), ts); err != nil {
+				t.Fatalf("send(%v, %v): %v", shardID, ts, err)
 			}
 		}
-		high, err := m.HighWatermark(ctx, domainID)
+		highs, err := m.HighWatermarks(ctx, domainID)
 		if err != nil {
 			t.Fatalf("HighWatermark(): %v", err)
```
pavelkalinnikov: nit: `HighWatermarks`
Same 3 lines below.

impl/sql/mutationstorage/queue_test.go
```diff
@@ -65,31 +113,66 @@ func TestWatermark(t *testing.T) {
...
-		{desc: "second", send: true, ts: ts2, want: ts2.UnixNano()},
+		{desc: "no rows", want: map[int64]int64{}},
+		{
+			desc: "first",
+			send: map[int64]time.Time{1: ts1},
+			want: map[int64]int64{1: ts1.UnixNano()},
+		},
+		{
+			desc: "second",
+			// Highwatermarks in each shard proceed independently.
```
pavelkalinnikov: nit: `High watermarks`

core/adminserver/admin_server.go
```diff
@@ -260,6 +270,13 @@ func (s *Server) CreateDomain(ctx context.Context, in *pb.CreateDomainRequest) (
 	}); err != nil {
 		return nil, fmt.Errorf("adminserver: domains.Write(): %v", err)
 	}
+
+	// Create shards for queue.
+	shardIDs := []int64{1}
```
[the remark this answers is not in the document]
gdbelvin: Created issue #1048

core/adminserver/admin_server_test.go
```diff
@@ -85,6 +85,12 @@ func (e *miniEnv) Close() {
 	e.stopMockServer()
 }
 
+type fakeQueueAdmin struct{}
+
+func (*fakeQueueAdmin) AddShards(ctx context.Context, domainID string, shardIDs ...int64) error {
```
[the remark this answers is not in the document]
gdbelvin: Yep, that would simplify things a bit.

core/sequencer/sequencer_api.proto
```diff
@@ -36,8 +36,9 @@ message MapMetadata {
       // of logged items MUST be monotonically increasing.
       int64 highest_watermark = 2;
    }
-   // source defines the range of inputs used for this map revision.
-   SourceSlice source = 1;
+   reserved 1;
```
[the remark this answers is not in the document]
gdbelvin: Fortunately we're early enough that we're happy to break compatibility. 
This PR is also introducing backwards-incompatible database schema changes as well.

core/sequencer/sequencer_api.proto
```diff
@@ -36,8 +36,9 @@ message MapMetadata {
...
+   // sources defines the ranges of inputs used for this map revision for each slice.
```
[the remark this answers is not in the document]
gdbelvin: sources is a map from log source IDs to the (low, high] range of primary keys in each slice that were used to construct this map revision.

core/sequencer/server.go
```diff
@@ -70,11 +70,10 @@ func createMetrics(mf monitoring.MetricFactory) {
...
 // Queue reads messages that haven't been deleted off the queue.
 type Queue interface {
-	// HighWatermark returns the highest primary key in the mutations table for DomainID.
-	HighWatermark(ctx context.Context, domainID string) (int64, error)
-	// ReadQueue returns the messages between (low, high] for domainID.
-	// TODO(gbelvin): Add paging API back in to support sharded reads.
-	ReadQueue(ctx context.Context, domainID string, low, high int64) ([]*mutator.QueueMessage, error)
+	// HighWatermark returns the highest timestamp in the mutations table.
+	HighWatermarks(ctx context.Context, domainID string) (map[int64]int64, error)
+	// Read returns up to batchSize messages for domainID.
```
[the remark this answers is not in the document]
gdbelvin: I suppose that this is no longer really a `Queue` because we will not be deleting messages from the queue ever.
The reason is that we want to keep these messages around in order to replay them into the map in case of failure.

I'll replace `Queue` with `Logs`. Would that be clearer?

core/sequencer/server.go
```diff
@@ -138,53 +137,55 @@ func (s *Server) RunBatch(ctx context.Context, in *spb.RunBatchRequest) (*empty.
...
+		}
 	}
 
-	msgs, err := s.readMessages(ctx, in.DomainId, metadata.GetSource())
+	msgs, err := s.readMessages(ctx, in.DomainId, sources)
 	if err != nil {
 		return nil, err
 	}
-	if int32(len(msgs)) < in.MinBatch {
+	if len(msgs) < int(in.MinBatch) {
```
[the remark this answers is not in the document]
gdbelvin: The reason we want `MinBatch != 0` is to avoid creating a lot of empty batches. Each new batch creates additional data for clients to download -- and we're running this function every 100ms or so because we want new messages to be picked up quickly.

The plan is to replace `MinBatch` with a comparison of watermarks and timestamps.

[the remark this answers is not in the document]
gdbelvin: The TODO is #1047

core/sequencer/server.go
```diff
@@ -138,53 +137,55 @@ func (s *Server) RunBatch(ctx context.Context, in *spb.RunBatchRequest) (*empty.
...
-		return nil, status.Errorf(codes.Internal, "ReadQueue(): %v", err)
-	}
-	msgs := make([]*ktpb.EntryUpdate, 0, len(batch))
-	for _, m := range batch {
-		msgs = append(msgs, &ktpb.EntryUpdate{
-			Mutation:  m.Mutation,
-			Committed: m.ExtraData,
-		})
+	sources map[int64]*spb.MapMetadata_SourceSlice) ([]*ktpb.EntryUpdate, error) {
+	msgs := make([]*ktpb.EntryUpdate, 0)
```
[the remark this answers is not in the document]
gdbelvin: If I remove it, vet complains. 
I think the len parameter is not required for maps, but it is for lists.

impl/sql/mutationstorage/mutations.go
```diff
@@ -50,9 +50,16 @@ var (
...
+		ShardID  BIGINT        NOT NULL,
 		Time     BIGINT        NOT NULL,
 		Mutation BLOB          NOT NULL,
-		PRIMARY KEY(DomainID, Time)
+		PRIMARY KEY(DomainID, ShardID, Time)
+	);`,
+		`CREATE TABLE IF NOT EXISTS Shards (
+		DomainID VARCHAR(30)   NOT NULL,
+		ShardID  BIGINT        NOT NULL,
+		Enabled  INTEGER       NOT NULL,
```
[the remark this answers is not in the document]
gdbelvin: Correct, right now it's always True.  #1063 Will need it, but that's a feature for the future. 
I'm not sure how easy it is to add columns to a table after the fact.  Do you think it's really best practice to remove it? I would agree about code, but databases are scarier.

impl/sql/mutationstorage/queue.go
```diff
@@ -29,29 +30,74 @@ import (
...
 )
 
+// AddShards creates and adds new shards for queue writing to a domain.
+func (m *Mutations) AddShards(ctx context.Context, domainID string, shardIDs ...int64) error {
+	glog.Infof("mutationstorage: AddShard(%v, %v)", domainID, shardIDs)
+	for _, shardID := range shardIDs {
+		if _, err := m.db.ExecContext(ctx,
+			`INSERT INTO Shards (DomainID, ShardID, Enabled)  Values(?, ?, ?);`,
+			domainID, shardID, true); err != nil {
+			return err
```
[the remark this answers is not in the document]
gdbelvin: Great point. This could change to `INSERT IGNORE`. Unfortunately  `INSERT IGNORE` doesn't have the same syntax in SQLite as it does in MySQL...  

I'll add a TODO for this and to change the testing database off of SQLite

impl/sql/mutationstorage/queue.go
gdbelvin: Thanks for catching that

impl/sql/mutationstorage/queue.go
gdbelvin: Time is a `NOT NULL` column, but `Max(Time)` is `Nil` if there are no rows.

impl/sql/mutationstorage/queue.go
gdbelvin: I've seen this pattern -- On the one hand it's nice because we get to use `defer`, but I find it overly complex and hard to follow myself. What do you think?

impl/sql/mutationstorage/queue.go
gdbelvin: This one doesn't need it because we're grouping by `ShardID` and therefore don't have a case where `Max` will operate on no rows. -- If there are no rows for a shard, it just won't be returned at all. 

I test for this case in `TestHighwatermarks`

impl/sql/mutationstorage/queue_test.go
gdbelvin: good idea

gdbelvin:
PTAL

core/adminserver/admin_server.go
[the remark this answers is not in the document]
pavelkalinnikov: You probably meant #1063
(putting it to create a link between github issues)

core/sequencer/sequencer_api.proto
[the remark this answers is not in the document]
pavelkalinnikov: Yeah, that was my point. Why using `reserved 1` instead of just dropping / reassigning the tag to the new field?

impl/sql/mutationstorage/queue.go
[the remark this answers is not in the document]
pavelkalinnikov: SGTM.

impl/sql/mutationstorage/queue.go
pavelkalinnikov: Ack.

impl/sql/mutationstorage/queue_test.go
```diff
@@ -37,24 +37,22 @@ func TestRandShard(t *testing.T) {
...
 	for _, tc := range []struct {
 		desc       string
-		send       bool
+		send       []int64
 		wantCode   codes.Code
 		wantShards map[int64]bool
 	}{
 		{desc: "no rows", wantCode: codes.NotFound, wantShards: map[int64]bool{}},
-		{desc: "second", send: true, wantShards: map[int64]bool{
+		{desc: "second", send: []int64{1, 2, 3}, wantShards: map[int64]bool{
```
pavelkalinnikov: Can you add another case for single shard?
Are there restrictions on shard IDs themselves (e.g. they should be positive)?

impl/sql/mutationstorage/queue_test.go
```diff
@@ -51,45 +97,80 @@ func TestSend(t *testing.T) {
...
 		}
-		high, err := m.HighWatermark(ctx, domainID)
+		highs, err := m.HighWatermarks(ctx, domainID)
 		if err != nil {
 			t.Fatalf("HighWatermark(): %v", err)
 		}
-		if high != tc.want {
-			t.Errorf("HighWatermark(): %v, want > %v", high, tc.want)
+		if !cmp.Equal(highs, tc.want) {
+			t.Errorf("HighWatermark(): %v, want %v", highs, tc.want)
```
pavelkalinnikov: HighWatermarks

impl/sql/mutationstorage/queue.go
pavelkalinnikov: Dunno, I personally have the opposite feeling w.r.t. complexity - it's easier to take care about the rollback once next to where Tx is created, and forget about it below, instead of trying to cover all code paths manually. But if you feel strongly, then leave it as is.

impl/sql/mutationstorage/queue_test.go
gdbelvin: Added test.
shardIDs can be any unique number.

impl/sql/mutationstorage/queue.go
gdbelvin: You're right. The defer is better.

core/sequencer/sequencer_api.proto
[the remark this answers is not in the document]
gdbelvin: Good habits I suppose :-)

gdbelvin:
Thanks very much for the review!

PR #1053, opened 2018-10-03
Generate consistency proofs in test vectors

jtoohill:
The trusted log root will fall behind a bit to make consistency proofs
more interesting.

Also changes the testdata generator to only open and write files once,
rather than overwriting the same file multiple times.

### DIFF
--- a/cmd/gen-test-vectors/main.go
+++ b/cmd/gen-test-vectors/main.go
@@ -86,6 +86,9 @@ func GenerateTestVectors(ctx context.Context, env *integration.Env) error {
 	// Collect a list of valid GetEntryResponses
 	getEntryResps := make([]testdata.GetEntryResponseVector, 0)
 
+	// Start with an empty trusted log root
+	slr := &types.LogRootV1{}
+
 	for _, tc := range []struct {
 		desc           string
 		wantProfile    []byte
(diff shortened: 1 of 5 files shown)


cmd/gen-test-vectors/main.go
```diff
@@ -143,16 +146,19 @@ func GenerateTestVectors(ctx context.Context, env *integration.Env) error {
...
+			UserId:        tc.userID,
+			AppId:         appID,
+			FirstTreeSize: int64(slr.TreeSize),
 		})
 		if err != nil {
 			return fmt.Errorf("gen-test-vectors: GetEntry(): %v", err)
 		}
-		if _, _, err := env.Client.VerifyGetEntryResponse(ctx, env.Domain.DomainId, appID, tc.userID, types.LogRootV1{}, e); err != nil {
+		var newslr *types.LogRootV1
+		if _, newslr, err = env.Client.VerifyGetEntryResponse(ctx, env.Domain.DomainId, appID, tc.userID, *slr, e); err != nil {
```
golangcibot: line is 122 characters

cmd/gen-test-vectors/main.go
```diff
@@ -163,6 +169,12 @@ func GenerateTestVectors(ctx context.Context, env *integration.Env) error {
 			Resp:   e,
 		})
 
+		// Let the trusted log root fall behind by a few revisions
+		// to make consistency proofs more interesting.
+		if newslr.TreeSize%5 == 1 {
```
gdbelvin: nit: might be cleaner to compare with 0 -- unless you're using 1 for a reason in which case it would be nice to document it.

jtoohill: Documented that we're using 1 intentionally. (There aren't more than 5 test cases at the moment, so 0 would use the empty trusted log root which doesn't require consistency proofs.)

core/client/verify_test.go
```diff
@@ -58,12 +58,16 @@ func TestVerifyGetEntryResponse(t *testing.T) {
 		t.Fatalf("Unmarshal(): %v", err)
 	}
 
+	trusted := &types.LogRootV1{}
 	for _, tc := range getEntryResponses {
 		t.Run(tc.Desc, func(t *testing.T) {
-			trusted := types.LogRootV1{}
-			if _, _, err := v.VerifyGetEntryResponse(ctx, domainPB.DomainId, tc.AppID, tc.UserID, trusted, tc.Resp); err != nil {
+			var slr *types.LogRootV1
+			if _, slr, err = v.VerifyGetEntryResponse(ctx, domainPB.DomainId, tc.AppID, tc.UserID, *trusted, tc.Resp); err != nil {
```
golangcibot: line is 122 characters

PR #1054, opened 2018-10-03
Merge Tink API changes

jtoohill:
Follows API cleanup in Tink per google/tink#93.

It seems like Tink removed the capability to read/write encrypted keysets, so I've had to use insecure.KeysetHandle in a few places.

### DIFF
--- a/core/testutil/tink.go
+++ b/core/testutil/tink.go
@@ -20,8 +20,8 @@ import (
 	"fmt"
 
 	"github.com/golang/protobuf/proto"
-	"github.com/google/tink/go/insecure"
 	"github.com/google/tink/go/signature"
+	"github.com/google/tink/go/testkeysethandle"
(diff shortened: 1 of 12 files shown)


core/testutil/tink.go
```diff
@@ -117,9 +118,9 @@ func SignKeysetsFromPEMs(privPEMs ...string) []*tink.KeysetHandle {
 		}
 		keysetKey := PrivateKeyFromPEM(pem, uint32(i+1))
 		keyset := tink.CreateKeyset(uint32(i+1), []*tinkpb.Keyset_Key{keysetKey})
-		parsedHandle, err := tink.CleartextKeysetHandle().ParseKeyset(keyset)
+		parsedHandle, err := insecure.KeysetHandle(keyset)
```
gdbelvin: This should be `testkeysethandle.KeysetHandle`

PR #1058, opened 2018-10-05
caching gometalinter in travis

ductm54:


gdbelvin:
Thanks for contributing!  

Was the travis install step for `gometalinter` slow on your infrastructure?

ductm54:
> Thanks for contributing!
> 
> Was the travis install step for `gometalinter` slow on your infrastructure?

From what i read from  https://github.com/alecthomas/gometalinter/blob/master/main.go#L48
gometalinter --install going to DEPRECATED soon
So i propose a way to use binary release instead

gdbelvin:
This is a very interesting proposal. Unfortunately, it involves more complexity that we'd like to maintain.

Once Go's modules are more broadly adopted, tools will have repeatable builds and gometalinter might re-enable `go get`.  Let's wait a bit and see what happens...

PR #1060, opened 2018-10-08
Fix gcloud install and deploy script 

gdbelvin:
Update the gcloud install script. 

For some reason, the main build was erroring with 
```
gcloud compoenents install kubeclt was erroring with
ERROR: (gcloud.components.install) You cannot perform this action because this Cloud SDK installation is managed by an external package manager.
Please consider using a separate installation of the Cloud SDK created through the default mechanism described at: https://cloud.google.com/sdk/
```

- Install via `wget` and `./install.sh`. 
- Update the authentication keys while we're at it.


.travis.yml
```diff
@@ -18,9 +18,23 @@ addons:
     packages:
     - python3-pip
 
+before_install:
+  - if [ ! -d $HOME/gcloud/google-cloud-sdk ]; then
+        mkdir -p $HOME/gcloud &&
+        wget https://dl.google.com/dl/cloudsdk/release/google-cloud-sdk.tar.gz --directory-prefix=$HOME/gcloud &&
+        cd $HOME/gcloud &&
+        tar xzf google-cloud-sdk.tar.gz &&
+        printf '\ny\n\ny\ny\n' | ./google-cloud-sdk/install.sh &&
```
AlCutter: you might be able to use the `--quiet` flag here rather than piping canned answers in - that's a little brittle.

AlCutter: Also I'm not 100% certain about the interaction between the compound `&& ... && ... &&` and a pipe with a compound command on the other side...

.travis.yml
```diff
@@ -36,14 +50,9 @@ before_deploy:
...
-  - if [ ! -d "$HOME/google-cloud-sdk/bin" ]; then rm -rf $HOME/google-cloud-sdk;
-    curl https://sdk.cloud.google.com | bash; fi
-  - source $HOME/google-cloud-sdk/path.bash.inc
-  - gcloud --quiet components install kubectl
-  - gcloud --quiet version
-  - openssl aes-256-cbc -K $encrypted_a1eb99cfc21e_key -iv $encrypted_a1eb99cfc21e_iv -in travis_secrets.tar.gz.enc -out travis_secrets.tar.gz -d
-  - tar -xzf travis_secrets.tar.gz
-  - gcloud auth activate-service-account --key-file service_key.json
+  - openssl aes-256-cbc -K $encrypted_555d9b2948d2_key -iv $encrypted_555d9b2948d2_iv
+    -in client_secrets.json.enc -out client_secrets.json -d
```
AlCutter: just a thought - there's `/var/ramfs` if you wanted to be "sure" that this file wasn't accidentally backed up to the homedir cache or otherwise laying around.

.travis.yml
```diff
@@ -36,14 +50,9 @@ before_deploy:
...
+  - gcloud auth activate-service-account --key-file client_secrets.json
```
AlCutter: Hmm, alternatively, _maybe_ you could do something like:

```
openssl aes..... -d | gcloud auth activate-service-account --key-file /dev/stdin

```
?

.travis.yml
```diff
@@ -52,6 +61,6 @@ before_deploy:
 deploy:
   skip_cleanup: true
   provider: script
-  script: ./scripts/deploy.sh
+  script: "./scripts/deploy.sh"
```
AlCutter: Are the quotes necessary?

gdbelvin: They are probably gratuitous here.

.travis.yml
gdbelvin: Added `-q` and removed the `&&`'s

.travis.yml
gdbelvin: I think that will work...

gdbelvin:
Cleaned up the nits.  PTAL

PR #1062, opened 2018-10-09
Add ';' to separate if statements

gdbelvin:
The multiple statements inside the bash if block are being treated as one continuous command.
This PR separates them properly with semicolons.

Fixes #1060

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,11 +20,11 @@ addons:
 
 before_install:
   - if [ ! -d $HOME/gcloud/google-cloud-sdk ]; then
-        mkdir -p $HOME/gcloud
-        wget https://dl.google.com/dl/cloudsdk/release/google-cloud-sdk.tar.gz --directory-prefix=$HOME/gcloud
-        cd $HOME/gcloud
-        tar xzf google-cloud-sdk.tar.gz
-        ./google-cloud-sdk/install.sh -q
+        mkdir -p $HOME/gcloud; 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -20,11 +20,11 @@ addons:
 
 before_install:
   - if [ ! -d $HOME/gcloud/google-cloud-sdk ]; then
-        mkdir -p $HOME/gcloud
-        wget https://dl.google.com/dl/cloudsdk/release/google-cloud-sdk.tar.gz --directory-prefix=$HOME/gcloud
-        cd $HOME/gcloud
-        tar xzf google-cloud-sdk.tar.gz
-        ./google-cloud-sdk/install.sh -q
+        mkdir -p $HOME/gcloud;
```
AlCutter: I _think_ that a less punctuationful way of doing this is with a _literal block_ entry:

```yaml
  - |
    if [ -d ....]; ...
       blah
       blah
    fi
 - next thing
```

AlCutter:
Could you add a comment here outlining why it was ok to merge with failing travis?

gdbelvin:
Sure thing: While we were waiting for this PR to get merged, google/tink published breaking changes to their repo which are resolved in #1061. 

In order to avoid merging the two PRs together and polluting the commit stream, one of them needed to be merged while Travis was unhappy.

gdbelvin:
I think it was better to merge this one first because it was simpler, and because it was preventing #1061 from being verified by travis.  There were two breakages and no single PR could have made the build green, unless we merged the two PRs.

PR #1064, opened 2018-10-11
Rename *Queue* with *Logs*

gdbelvin:
The previous few PRs have been migrating Key Transparency away from a queue based model where things in the queue are deleted after they have been processed, to a log(s) based model, where once things go in the log, they stay there forever.  This change helps increase reliability by creating a journal that can be replayed into the map at any time if things go haywire for any reason. 

This PR clarifies variables and comments by switching the terminology from Queues and Shards to a group of logs and individual logIDs.

### DIFF
--- a/cmd/keytransparency-server/main.go
+++ b/cmd/keytransparency-server/main.go
@@ -120,7 +120,7 @@ func main() {
 	mapAdmin := trillian.NewTrillianAdminClient(mconn)
 
 	// Create gRPC server.
-	queue := mutator.MutationQueue(mutations)
+	queue := mutator.MutationLogs(mutations)
 	ksvr := keyserver.New(tlog, tmap, logAdmin, mapAdmin,
 		entry.New(), domains, queue, mutations)
 	grpcServer := grpc.NewServer(
--- a/cmd/keytransparency-server/main.go
+++ b/cmd/keytransparency-server/main.go
@@ -21,7 +21,6 @@ import (
 
 	"github.com/google/keytransparency/cmd/serverutil"
 	"github.com/google/keytransparency/core/keyserver"
-	"github.com/google/keytransparency/core/mutator"
 	"github.com/google/keytransparency/core/mutator/entry"
 	"github.com/google/keytransparency/impl/authentication"
 	"github.com/google/keytransparency/impl/authorization"
@@ -100,9 +99,9 @@ func main() {
 	if err != nil {
 		glog.Exitf("Failed to create domain storage: %v", err)
 	}
-	mutations, err := mutationstorage.New(sqldb)
+	logs, err := mutationstorage.New(sqldb)
 	if err != nil {
-		glog.Exitf("Failed to create mutations object: %v", err)
+		glog.Exitf("Failed to create logs object: %v", err)
 	}
 
 	// Connect to log and map server.
@@ -120,9 +119,8 @@ func main() {
 	mapAdmin := trillian.NewTrillianAdminClient(mconn)
 
 	// Create gRPC server.
-	queue := mutator.MutationLogs(mutations)
 	ksvr := keyserver.New(tlog, tmap, logAdmin, mapAdmin,
-		entry.New(), domains, queue, mutations)
+		entry.New(), domains, logs, logs)
 	grpcServer := grpc.NewServer(
 		grpc.Creds(creds),
 		grpc.StreamInterceptor(grpc_middleware.ChainStreamServer(
--- a/core/adminserver/admin_server.go
+++ b/core/adminserver/admin_server.go
@@ -75,9 +75,9 @@ var (
 	}
 )
 
-// LogsAdmin controls the lifecycle and scaling of mutation queues.
+// LogsAdmin controls the lifecycle and scaling of mutation logs.
 type LogsAdmin interface {
-	// AddLogs creates and adds new shards for queue writing to a domain.
+	// AddLogs creates and adds new shards for writing to a domain.
 	AddLogs(ctx context.Context, domainID string, logIDs ...int64) error
 }
 
@@ -271,8 +271,8 @@ func (s *Server) CreateDomain(ctx context.Context, in *pb.CreateDomainRequest) (
 		return nil, fmt.Errorf("adminserver: domains.Write(): %v", err)
 	}
 
-	// Create initial shards for queue.
-	// TODO(#1063): Additional shards can be added at a later point to support increased server load.
+	// Create initial logs for writing.
+	// TODO(#1063): Additional logs can be added at a later point to support increased server load.
 	shardIDs := []int64{1, 2}
 	if err := s.logsAdmin.AddLogs(ctx, in.GetDomainId(), shardIDs...); err != nil {
 		return nil, fmt.Errorf("adminserver: AddShards(%v): %v", shardIDs, err)
--- a/core/adminserver/admin_server.go
+++ b/core/adminserver/admin_server.go
@@ -75,21 +75,21 @@ var (
 	}
 )
 
-// QueueAdmin controls the lifecycle and scaling of mutation queues.
-type QueueAdmin interface {
-	// AddShards creates and adds new shards for queue writing to a domain.
-	AddShards(ctx context.Context, domainID string, shardIDs ...int64) error
+// LogsAdmin controls the lifecycle and scaling of mutation queues.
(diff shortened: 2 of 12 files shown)


core/sequencer/server_test.go
```diff
@@ -25,7 +25,7 @@ import (
 	tpb "github.com/google/trillian"
 )
 
-func queueMsg(t *testing.T, id int64, signer *tink.KeysetHandle) *ktpb.EntryUpdate {
+func logMsg(t *testing.T, id int64, signer *tink.KeysetHandle) *ktpb.EntryUpdate {
```
golangcibot: `logMsg` - `id` always receives `1`

core/sequencer/server.go
```diff
@@ -68,13 +68,13 @@ func createMetrics(mf monitoring.MetricFactory) {
...
-type Queue interface {
-	// HighWatermarks returns the highest primary key for each shard in the mutations table.
+// LogsReader reads messages in multiple log logs.
+type LogsReader interface {
+	// HighWatermarks returns the highest primary key for each log in the mutations table.
 	HighWatermarks(ctx context.Context, domainID string) (map[int64]int64, error)
-	// ReadQueue returns the messages under shardID in the (low, high] range.
-	// ReadQueue does NOT delete messages.
-	ReadQueue(ctx context.Context, domainID string, shardID, low, high int64) ([]*mutator.QueueMessage, error)
+	// ReadLog returns the messages under logID in the (low, high] range.
```
AlCutter: "... the messages in the (low, high] range stored in specified log."?

core/sequencer/server.go
```diff
@@ -103,11 +103,11 @@ func NewServer(
 		tlog:      tlog,
 		tmap:      tmap,
 		mutations: mutations,
-		queue:     queue,
+		logs:      logs,
 	}
 }
 
-// RunBatch reads mutations out of the queue and calls CreateEpoch.
+// RunBatch reads mutations out of the logs and calls CreateEpoch.
```
AlCutter: unrelated, but I think it'd help the reader if this method comment talked a bit about what and why rather than (or perhaps in addition to if you feel it's important) how.

impl/sql/mutationstorage/queue.go
```diff
@@ -58,40 +58,40 @@ func (m *Mutations) Send(ctx context.Context, domainID string, update *pb.EntryU
...
-// randShard returns a random, enabled shard for domainID.
-func (m *Mutations) randShard(ctx context.Context, domainID string) (int64, error) {
+// randLog returns a random, enabled log for domainID.
+func (m *Mutations) randLog(ctx context.Context, domainID string) (int64, error) {
 	// TODO(gbelvin): Cache these results.
-	var shardIDs []int64
+	var logIDs []int64
 	rows, err := m.db.QueryContext(ctx,
-		`SELECT ShardID from Shards WHERE DomainID = ? AND Enabled = ?;`,
+		`SELECT LogID from Logs WHERE DomainID = ? AND Enabled = ?;`,
```
AlCutter: Since `Enabled` is always `true`, could you pop it into the SQL itself rather than as a const parameter?

impl/sql/mutationstorage/queue.go
```diff
@@ -58,40 +58,40 @@ func (m *Mutations) Send(ctx context.Context, domainID string, update *pb.EntryU
...
-		shardIDs = append(shardIDs, shardID)
+		logIDs = append(logIDs, logID)
 	}
 	if err := rows.Err(); err != nil {
 		return 0, err
 	}
-	if len(shardIDs) == 0 {
-		return 0, status.Errorf(codes.NotFound, "No shard found for domain %v", domainID)
+	if len(logIDs) == 0 {
+		return 0, status.Errorf(codes.NotFound, "No log found for domain %v", domainID)
```
AlCutter: No/no

impl/sql/mutationstorage/queue.go
```diff
@@ -187,8 +187,8 @@ func readQueueMessages(rows *sql.Rows) ([]*mutator.QueueMessage, error) {
 	return results, nil
 }
 
-// DeleteMessages removes messages from the queue.
-func (m *Mutations) DeleteMessages(ctx context.Context, domainID string, mutations []*mutator.QueueMessage) error {
+// DeleteMessages removes messages from the log.
```
AlCutter: I look forward to seeing the implementation of this one :)

cmd/keytransparency-server/main.go
```diff
@@ -120,9 +119,8 @@ func main() {
 	mapAdmin := trillian.NewTrillianAdminClient(mconn)
 
 	// Create gRPC server.
-	queue := mutator.MutationQueue(mutations)
 	ksvr := keyserver.New(tlog, tmap, logAdmin, mapAdmin,
-		entry.New(), domains, queue, mutations)
+		entry.New(), domains, logs, logs)
```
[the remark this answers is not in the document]
gdbelvin: Indeed it does :-)

The beauty of go interfaces is that the same object can satisfy multiple interfaces.
The old `mutations` object is going to be deprecated in the next PR as well.

core/adminserver/admin_server.go
```diff
@@ -75,21 +75,21 @@ var (
...
 )
 
-// QueueAdmin controls the lifecycle and scaling of mutation queues.
-type QueueAdmin interface {
-	// AddShards creates and adds new shards for queue writing to a domain.
-	AddShards(ctx context.Context, domainID string, shardIDs ...int64) error
+// LogsAdmin controls the lifecycle and scaling of mutation logs.
+type LogsAdmin interface {
+	// AddLogs creates and adds new logs for writing to a domain.
+	AddLogs(ctx context.Context, domainID string, logIDs ...int64) error
```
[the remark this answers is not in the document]
gdbelvin: Good question. 

1. This is a bit of a convenience method. ie. whatever the caller would do, it would be nice for it to do it internally. 
2. Pavel pointed this out as well - and we concluded that retrying this method should be idempotent.

core/sequencer/server.go
```diff
@@ -68,13 +68,13 @@ func createMetrics(mf monitoring.MetricFactory) {
 		domainIDLabel)
 }
 
-// Queue reads messages that haven't been deleted off the queue.
-type Queue interface {
-	// HighWatermarks returns the highest primary key for each shard in the mutations table.
+// LogsReader reads messages in multiple log logs.
```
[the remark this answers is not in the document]
gdbelvin: used to be `queue logs`.  Sorry about not catching that in my own review.

core/sequencer/server.go
```diff
@@ -68,13 +68,13 @@ func createMetrics(mf monitoring.MetricFactory) {
...
+type LogsReader interface {
+	// HighWatermarks returns the highest primary key for each log in the mutations table.
```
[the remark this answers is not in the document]
gdbelvin: We've been deliberately vague here. 
The PK can either be a timestamp - the current, tabled based implementation - but it could also be an index in a Trillian Log when we move to that in the future. 

The difference to the caller, though, should be completely transparent.

impl/sql/mutationstorage/queue.go
gdbelvin: haha.  This reminds me that I need to delete this.

impl/sql/mutationstorage/queue.go
gdbelvin: I could, but I was worried about `True` vs `true` vs `TRUE` for the specific SQL dialect.

impl/sql/mutationstorage/queue.go
gdbelvin: #1067

gdbelvin:
PTAL

gdbelvin:
PTAL

cmd/keytransparency-server/main.go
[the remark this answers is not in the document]
AlCutter: Cool, that last one's the answer I was hoping for :)

core/sequencer/server.go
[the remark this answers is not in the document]
AlCutter: Ok, maybe "identifier" or similar, then?

impl/sql/mutationstorage/queue.go
AlCutter: are there SQL dialects where these are case sensitive?

PR #1067, opened 2018-10-12
Remove DeleteMessages API

gdbelvin:
Since we are no longer using a queue, but a log of messages, we
shouldn't be deleting messages.


AlCutter:
Looks like there might be some helper methods which are now unneeded:

```
impl/sql/mutationstorage/mutations.go:29:1:warning: deleteQueueExpr is unused (deadcode)
impl/sql/mutationstorage/mutations.go:37:2:warning: const deleteQueueExpr is unused (U1000) (megacheck)
impl/sql/mutationstorage/mutations.go:37:2:warning: unused variable or constant deleteQueueExpr (varcheck)
```

PR #1068, opened 2018-10-16
Limit maximum number of mutations with MaxBatch

gdbelvin:
Put a limit on the number of items to be included in a map revision. 
This is necessary because the initial number of items in the mutation logs may be huge. 

#897


impl/sql/mutationstorage/queue.go
```diff
@@ -125,27 +140,49 @@ func (m *Mutations) send(ctx context.Context, domainID string, logID int64, mDat
 	return tx.Commit()
 }
 
-// HighWatermarks returns the highest timestamp for each log in the mutations table.
-func (m *Mutations) HighWatermarks(ctx context.Context, domainID string) (map[int64]int64, error) {
+// HighWatermarks returns the highest watermark that is less than or equal to batchSize items
+// greater than the watermark for that log in starts, for each log for domainID.
+// The total size will be numLogs * batchSize.
+func (m *Mutations) HighWatermarks(ctx context.Context, domainID string, starts map[int64]int64, batchSize int32) (map[int64]int32, map[int64]int64, error) {
```
golangcibot: line is 157 characters

impl/sql/mutationstorage/queue.go
```diff
@@ -125,27 +140,49 @@ func (m *Mutations) send(ctx context.Context, domainID string, logID int64, mDat
...
+		counts[logID] = count
 	}
-	if err := rows.Err(); err != nil {
-		return nil, err
-	}
-	return watermarks, nil
+	return counts, watermarks, nil
+}
+
+// highWatarmark returns the highest watermark in logID that is less than or equal to batchSize items greater than start.
```
golangcibot: line is 121 characters

impl/sql/mutationstorage/queue.go
```diff
@@ -125,27 +140,49 @@ func (m *Mutations) send(ctx context.Context, domainID string, logID int64, mDat
...
+func (m *Mutations) highWatermark(ctx context.Context, domainID string, logID, start int64, batchSize int32) (int32, int64, error) {
```
golangcibot: line is 132 characters

core/sequencer/server.go
```diff
@@ -70,11 +71,14 @@ func createMetrics(mf monitoring.MetricFactory) {
...
-	HighWatermarks(ctx context.Context, domainID string) (map[int64]int64, error)
+	// HighWatermarks returns the highest primary key that is less than or equal to batchSize
+	// items greater than the watermark for that log in starts, for each log for domainID.
+	// The total size will be numLogs * batchSize.
+	HighWatermarks(ctx context.Context, domainID string,
+		startWatermarks map[int64]int64, batchSize int32) (map[int64]int32, map[int64]int64, error)
 	// ReadLog returns the messages in the (low, high] range stored in the specified log.
 	// ReadLog does NOT delete messages.
-	ReadLog(ctx context.Context, domainID string, logID, low, high int64) ([]*mutator.LogMessage, error)
+	ReadLog(ctx context.Context, domainID string, logID, low, high int64, batchSize, offset int32) ([]*mutator.LogMessage, error)
```
golangcibot: line is 126 characters

core/sequencer/server.go
```diff
@@ -151,35 +164,52 @@ func (s *Server) RunBatch(ctx context.Context, in *spb.RunBatchRequest) (*empty.
...
+	// Rate limit the creation of new batches.
+	///
+
+	var totalCount int32
+	for _, count := range counts {
+		totalCount += count
 	}
-	if len(msgs) < int(in.MinBatch) {
-		return &empty.Empty{}, nil
+	// TODO(#1057): If time since last map revision > max timeout, run batch.
```
pavelkalinnikov: Note: If we change batching strategy to consider per-log batch sizes, then the timeout should be checked on per-log basis as well.

impl/sql/mutationstorage/queue.go
```diff
@@ -125,37 +140,63 @@ func (m *Mutations) send(ctx context.Context, domainID string, logID int64, mDat
...
-	if err := rows.Err(); err != nil {
-		return nil, err
-	}
-	return watermarks, nil
+	return counts, watermarks, nil
+}
+
+// highWatarmark returns the highest watermark in logID that is less than or
+// equal to batchSize items greater than start.
+func (m *Mutations) highWatermark(ctx context.Context, domainID string, logID,
```
pavelkalinnikov: Following our in-person discussion, it would be clearer / more flexible to have only this call behind the interface. The "global" batching strategy can stay outside.

core/sequencer/server.go
```diff
@@ -70,11 +71,15 @@ func createMetrics(mf monitoring.MetricFactory) {
 
 // LogsReader reads messages in multiple logs.
 type LogsReader interface {
-	// HighWatermarks returns the highest primary key for each log in the mutations table.
-	HighWatermarks(ctx context.Context, domainID string) (map[int64]int64, error)
+	// HighWatermarks returns the highest primary key that is less than or equal to batchSize
+	// items greater than the watermark for that log in starts, for each log for domainID.
+	// The total size will be numLogs * batchSize.
+	HighWatermarks(ctx context.Context, domainID string,
+		startWatermarks map[int64]int64, batchSize int32) (map[int64]int32, map[int64]int64, error)
```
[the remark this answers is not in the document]
gdbelvin: I've defined `Watermarks` and `SourcesEntry`

core/sequencer/server.go
gdbelvin: Perfect.

core/sequencer/server.go
```diff
@@ -151,35 +164,52 @@ func (s *Server) RunBatch(ctx context.Context, in *spb.RunBatchRequest) (*empty.
...
+			MapMetadata: &spb.MapMetadata{Sources: sources},
+		})
 	}
 
-	return s.CreateEpoch(ctx, &spb.CreateEpochRequest{
-		DomainId:    in.DomainId,
-		Revision:    int64(latestMapRoot.Revision) + 1,
-		MapMetadata: &spb.MapMetadata{Sources: sources},
-	})
+	// TODO(#1056): If count items > max_batch, should we define the next batch immediately?
```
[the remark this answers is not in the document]
gdbelvin: This comment is a little unclear. The idea is to detect when we've just run a full batch. This should read `count == max_batch`

core/sequencer/server.go
```diff
@@ -380,3 +404,34 @@ func (s *Server) applyMutations(domainID string, mutatorFunc mutator.Func,
 	glog.V(2).Infof("applyMutations applied %v mutations to %v leaves", len(msgs), len(leaves))
 	return ret, nil
 }
+
+// HighWatermarks returns the total count and the highest watermark for each log.
+// batchSize is a limit on the total number of items represented by the returned watermarks.
+// TODO(gbelvin): Block until a minBatchSize has been reached or a timeout has occured.
```
golangcibot: `occured` is a misspelling of `occurred`

core/sequencer/server.go
```diff
@@ -131,55 +142,68 @@ func (s *Server) RunBatch(ctx context.Context, in *spb.RunBatchRequest) (*empty.
...
+	return &empty.Empty{}, nil
 }
 
-func (s *Server) readMessages(ctx context.Context, domainID string,
-	sources map[int64]*spb.MapMetadata_SourceSlice) ([]*ktpb.EntryUpdate, error) {
+const readBatchSize = int32(1000)
+
+// readMessages returns the full set of EntryUpdates defined by sources.
+// batchSize limits the number of messages to read from a log at one time.
+func (s *Server) readMessages(ctx context.Context, domainID string, sources SourcesEntry, batchSize int32) ([]*ktpb.EntryUpdate, error) {
```
golangcibot: line is 137 characters

core/sequencer/server.go
```diff
@@ -380,3 +404,34 @@ func (s *Server) applyMutations(domainID string, mutatorFunc mutator.Func,
 	glog.V(2).Infof("applyMutations applied %v mutations to %v leaves", len(msgs), len(leaves))
 	return ret, nil
 }
+
+// HighWatermarks returns the total count and the highest watermark for each log.
+// batchSize is a limit on the total number of items represented by the returned watermarks.
+// TODO(gbelvin): Block until a minBatchSize has been reached or a timeout has occured.
+func (s *Server) HighWatermarks(ctx context.Context, domainID string, starts Watermarks, maxBatch int32) (int32, Watermarks, error) {
```
golangcibot: line is 133 characters

core/sequencer/server_test.go
```diff
@@ -15,16 +15,112 @@
...
 	"github.com/google/tink/go/tink"
 
 	ktpb "github.com/google/keytransparency/core/api/v1/keytransparency_go_proto"
+	spb "github.com/google/keytransparency/core/sequencer/sequencer_go_proto"
 	tpb "github.com/google/trillian"
 )
 
+type fakeLogs map[int64][]mutator.LogMessage
+
+func (l fakeLogs) ReadLog(ctx context.Context, domainID string, logID, low, high int64, batchSize int32) ([]*mutator.LogMessage, error) {
```
golangcibot: line is 137 characters

core/sequencer/server_test.go
```diff
@@ -15,16 +15,112 @@
...
+	logIDs := make([]int64, 0, len(l))
+	for logID := range l {
+		logIDs = append(logIDs, logID)
+	}
+	// sort logsIDs for test repeatability.
+	sort.Slice(logIDs, func(i, j int) bool { return logIDs[i] < logIDs[j] })
+	return logIDs, nil
+}
+
+func (l fakeLogs) HighWatermark(ctx context.Context, domainID string, logID, start int64, batchSize int32) (int32, int64, error) {
```
golangcibot: line is 130 characters

core/sequencer/server_test.go
```diff
@@ -15,16 +15,112 @@
...
+		0: make([]mutator.LogMessage, 10),
+		1: make([]mutator.LogMessage, 20),
+	}}
+
+	for _, tc := range []struct {
+		sources   SourcesEntry
+		batchSize int32
+		want      int
+	}{
+		{batchSize: 1, want: 9, sources: SourcesEntry{0: &spb.MapMetadata_SourceSlice{LowestWatermark: 0, HighestWatermark: 9}}},
```
golangcibot: line is 123 characters

core/sequencer/server.go
```diff
@@ -380,3 +404,34 @@ func (s *Server) applyMutations(domainID string, mutatorFunc mutator.Func,
...
+		count, high, err := s.logs.HighWatermark(ctx, domainID, logID, start, maxBatch)
+		if err != nil {
+			return 0, nil, status.Errorf(codes.Internal,
+				"highWatermark(%v/%v, start: %v, batch:%v): %v",
+				domainID, logID, start, batchSize, err)
+		}
+		watermarks[logID] = high
+		total += count
+		maxBatch -= count
+		if batchSize <= 0 {
```
golangcibot: cannot convert 0 (untyped int constant) to github.com/google/trillian/monitoring.Gauge

core/sequencer/server.go
```diff
@@ -131,55 +143,69 @@ func (s *Server) RunBatch(ctx context.Context, in *spb.RunBatchRequest) (*empty.
...
+			batch, err := s.logs.ReadLog(ctx, domainID, logID, low, source.GetHighestWatermark(), batchSize)
+			if err != nil {
+				return nil, status.Errorf(codes.Internal, "ReadLog(): %v", err)
+			}
+			for _, m := range batch {
+				msgs = append(msgs, &ktpb.EntryUpdate{
+					Mutation:  m.Mutation,
+					Committed: m.ExtraData,
+				})
+				if m.ID > low {
```
golangcibot: empty branch

core/sequencer/server_test.go
```diff
@@ -15,16 +15,119 @@
...
+	for _, tc := range []struct {
+		starts    Watermarks
+		batchSize int32
+		count     int32
+		highs     Watermarks
+	}{
+		{batchSize: 30, starts: Watermarks{}, count: 28, highs: Watermarks{0: 9, 1: 19}},
+		{batchSize: 20, starts: Watermarks{}, count: 20, highs: Watermarks{0: 9, 1: 11}},
+		{batchSize: 20, starts: Watermarks{0: 9}, count: 19, highs: Watermarks{0: 9, 1: 19}},
+		{batchSize: 1, starts: Watermarks{1: 9}, count: 1, highs: Watermarks{0: 1, 1: 9}},       // Don't drop existing watermarks.
```
golangcibot: line is 125 characters

core/sequencer/server_test.go
```diff
@@ -15,16 +15,119 @@
...
+		{batchSize: 0, starts: Watermarks{3: 9}, count: 0, highs: Watermarks{0: 0, 1: 0, 3: 9}}, // Don't drop pre-existing watermarks.
```
golangcibot: line is 129 characters

core/mutator/entry/mutator.go
```diff
@@ -83,7 +83,7 @@ func (*Mutator) Mutate(oldValue, update proto.Message) (proto.Message, error) {
 			glog.Warningf("mutation is a replay of an old one")
 			return nil, mutator.ErrReplay
 		}
-		glog.Warningf("previous entry hash (%v) does not match the hash provided in this mutation (%v)", prevEntryHash[:], newEntry.GetPrevious())
+		glog.Warningf("previous entry hash (%x) does not match the hash provided in this mutation (%x)", prevEntryHash[:], newEntry.GetPrevious())
```
golangcibot: line is 140 characters

core/sequencer/server.go
```diff
@@ -131,55 +143,70 @@ func (s *Server) RunBatch(ctx context.Context, in *spb.RunBatchRequest) (*empty.
...
-			return nil, status.Errorf(codes.Internal, "ReadQueue(): %v", err)
-		}
-		for _, m := range batch {
-			msgs = append(msgs, &ktpb.EntryUpdate{
-				Mutation:  m.Mutation,
-				Committed: m.ExtraData,
-			})
+		low := source.GetLowestWatermark()
+		count := readBatchSize
+		for count == readBatchSize {
```
pavelkalinnikov: I'm not following this line. Should this be `count < readBatchSize`, and `count := 0` above?

core/sequencer/server.go
```diff
@@ -131,55 +143,70 @@ func (s *Server) RunBatch(ctx context.Context, in *spb.RunBatchRequest) (*empty.
...
+			for _, m := range batch {
+				msgs = append(msgs, &ktpb.EntryUpdate{
+					Mutation:  m.Mutation,
+					Committed: m.ExtraData,
+				})
+				if m.ID > low {
+					low = m.ID
+				}
+			}
+			count = int32(len(batch))
```
pavelkalinnikov: nit: Maybe you could restructure it a bit, so that this line is merged into `for` definition?

core/sequencer/server.go
```diff
@@ -380,3 +410,44 @@ func (s *Server) applyMutations(domainID string, mutatorFunc mutator.Func,
...
+
+// HighWatermarks returns the total count and the highest watermark for each log.
+// batchSize is a limit on the total number of items represented by the returned watermarks.
+// TODO(gbelvin): Block until a minBatchSize has been reached or a timeout has occurred.
+func (s *Server) HighWatermarks(ctx context.Context, domainID string, starts Watermarks,
+	maxBatch int32) (int32, Watermarks, error) {
+	watermarks := make(Watermarks)
+	var total int32
+
+	// Ensure that we do not lose track of watermarks, even if they are no longer in the
```
pavelkalinnikov: Hm, this sounds dangerous to me. What if you add/delete shards as it goes? I think you should be able to do the work when some are missing from this list (just assume they stay on their current mark).

core/sequencer/server.go
```diff
@@ -380,3 +410,44 @@ func (s *Server) applyMutations(domainID string, mutatorFunc mutator.Func,
...
+	for _, logID := range logIDs {
+		start := starts[logID]
+		if maxBatch <= 0 {
+			watermarks[logID] = start
+			continue
+		}
+		count, high, err := s.logs.HighWatermark(ctx, domainID, logID, start, maxBatch)
+		if err != nil {
+			return 0, nil, status.Errorf(codes.Internal,
+				"highWatermark(%v/%v, start: %v, batch:%v): %v",
```
pavelkalinnikov: nit: HighWatermark

pavelkalinnikov: +space after colon

impl/sql/mutationstorage/queue.go
```diff
@@ -61,29 +61,44 @@ func (m *Mutations) Send(ctx context.Context, domainID string, update *pb.EntryU
 	return m.send(ctx, domainID, logID, mData, time.Now())
 }
 
-// randLog returns a random, enabled log for domainID.
-func (m *Mutations) randLog(ctx context.Context, domainID string) (int64, error) {
-	// TODO(gbelvin): Cache these results.
+// ListLogs returns a list of all logs for domainID, optionally filtered for writable logs.
+func (m *Mutations) ListLogs(ctx context.Context, domainID string, writable bool) ([]int64, error) {
+	var query string
+	if writable {
```
pavelkalinnikov: Maybe do:
```
query := "SELECT ..."
if writable {
  query += " WHERE ..."
}
```

impl/sql/mutationstorage/queue.go
```diff
@@ -61,29 +61,44 @@ func (m *Mutations) Send(ctx context.Context, domainID string, update *pb.EntryU
...
 	}
 	if len(logIDs) == 0 {
-		return 0, status.Errorf(codes.NotFound, "no log found for domain %v", domainID)
+		return nil, status.Errorf(codes.NotFound, "no log found for domain %v", domainID)
+	}
+	return logIDs, nil
+}
+
+// randLog returns a random, enabled log for domainID.
+func (m *Mutations) randLog(ctx context.Context, domainID string) (int64, error) {
```
pavelkalinnikov: Do you use this function anywhere? I see it only in tests.

impl/sql/mutationstorage/queue.go
```diff
@@ -125,37 +140,36 @@ func (m *Mutations) send(ctx context.Context, domainID string, logID int64, mDat
...
+// equal to batchSize items greater than start.
+func (m *Mutations) HighWatermark(ctx context.Context, domainID string, logID,
+	start int64, batchSize int32) (int32, int64, error) {
+	var count int32
+	var high int64
+	if err := m.db.QueryRowContext(ctx,
+		`SELECT COUNT(*), COALESCE(MAX(T1.Time), ?) FROM 
+		(
+			SELECT Q.Time FROM Queue as Q
+			WHERE Q.DomainID = ? AND Q.LogID = ? AND Q.Time > ?
```
pavelkalinnikov: Wait, do you store time? I'm not sure MySQL time is reliably ordered. Would auto_increment key be better?

The assumption that **must** be strongly supported by your storage is that once watermark == x, there will be no new messages with a lower value. I don't think it works with time just the way it works in Spanner.

pavelkalinnikov: Ah, found in the code that you query for max existing timestamp before creating a new one, so it should be correct. I'm not sure though how this would compare to auto_increment key in performance, perhaps needs some testing.

pavelkalinnikov: On the third hand, if you already read max(Timestamp) in a transaction then why can't it be just LeafIndex instead of Timestamp? You would then read it, +1, and write.

core/sequencer/server.go
```diff
@@ -68,13 +68,25 @@ func createMetrics(mf monitoring.MetricFactory) {
 		domainIDLabel)
 }
 
+// SourcesEntry is a map of SourceSlices by logID.
```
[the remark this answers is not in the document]
gdbelvin: The canonical name is now `logID`. There was a rename PR recently that might not have caught all the changes.   Despite the rename, `logID` and `shardID` are functionally equivalent.

core/sequencer/server.go
```diff
@@ -68,13 +68,25 @@ func createMetrics(mf monitoring.MetricFactory) {
...
 // LogsReader reads messages in multiple logs.
 type LogsReader interface {
-	// HighWatermarks returns the highest primary key for each log in the mutations table.
-	HighWatermarks(ctx context.Context, domainID string) (map[int64]int64, error)
-	// ReadLog returns the messages in the (low, high] range stored in the specified log.
-	// ReadLog does NOT delete messages.
-	ReadLog(ctx context.Context, domainID string, logID, low, high int64) ([]*mutator.LogMessage, error)
+	// HighWatermarks returns the number of items (not to exceed batchSize) between start and
+	// the highest primary key in the log identified by logID.
+	HighWatermark(ctx context.Context, domainID string, logID, start int64, batchSize int32) (int32, int64, error)
```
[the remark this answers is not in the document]
gdbelvin: It returns batchSize items _after_ `start`.  `start` is not included.  I'll adjust the function comment. does that make sense?

core/sequencer/server.go
```diff
@@ -131,55 +143,69 @@ func (s *Server) RunBatch(ctx context.Context, in *spb.RunBatchRequest) (*empty.
...
+			MapMetadata: &spb.MapMetadata{Sources: sources},
+		})
 	}
 
-	return s.CreateEpoch(ctx, &spb.CreateEpochRequest{
-		DomainId:    in.DomainId,
-		Revision:    int64(latestMapRoot.Revision) + 1,
-		MapMetadata: &spb.MapMetadata{Sources: sources},
-	})
+	// TODO(#1056): If count items == max_batch, should we define the next batch immediately?
```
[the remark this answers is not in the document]
gdbelvin: Indeed. - For future PRs.

core/sequencer/server.go
gdbelvin: This loops until less than `batchSize` items are returned. 
`count == readBatchSize` indicates that there are more items to be read.

core/sequencer/server.go
gdbelvin: I did try that, but it's not very readable because `ReadLog` would need to move into the last statement in the `for` loop.  `batch` is not visible in the `for` loop scope.

core/sequencer/server.go
gdbelvin: This struct currently serves two purposes - this distinction wasn't obvious to me until I traced down a subtle bug. 

1.  **Recreating a Map Revision** 
`sources` defines the ranges of inputs that constitute a map. These inputs are enough to recreate each map revision independently. As a result, if a log stays on it's current watermark, it can be dropped from this list, which is aesthetically desirable. 
2.  **Defining the *next* batch** 
When the sequencer takes a look around to figure out if any logs have grown since the last map revision, it needs the high watermarks of *all* logs that were considered for processing in the last map revision, even if those logs did not move.   ie. the sequencer needs the correct value of `start` for each potential source log.

impl/sql/mutationstorage/queue.go
gdbelvin: It's good security practice to **not** dynamically modify query strings in order to avoid any possibility of SQL injection.  I believe some DB libraries can even enforce that the query string is a compile time constant.

impl/sql/mutationstorage/queue.go
gdbelvin: `randLog` is used by `Send`

gdbelvin:
Updated the comments - PTAL

core/sequencer/server_test.go
```diff
@@ -15,16 +15,121 @@
...
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/keytransparency/core/mutator"
 	"github.com/google/keytransparency/core/mutator/entry"
 	"github.com/google/tink/go/signature"
 	"github.com/google/tink/go/tink"
 
 	ktpb "github.com/google/keytransparency/core/api/v1/keytransparency_go_proto"
+	spb "github.com/google/keytransparency/core/sequencer/sequencer_go_proto"
```
golangcibot: File is not `goimports`-ed

gdbelvin: I'm not sure what's happening with `goimports` - seems like it suddenly decided to not respect import groupings.

core/sequencer/server.go
```diff
@@ -68,13 +68,27 @@ func createMetrics(mf monitoring.MetricFactory) {
...
 // LogsReader reads messages in multiple logs.
 type LogsReader interface {
-	// HighWatermarks returns the highest primary key for each log in the mutations table.
-	HighWatermarks(ctx context.Context, domainID string) (map[int64]int64, error)
-	// ReadLog returns the messages in the (low, high] range stored in the specified log.
-	// ReadLog does NOT delete messages.
-	ReadLog(ctx context.Context, domainID string, logID, low, high int64) ([]*mutator.LogMessage, error)
+	// highwatermarks returns the number of items and the highest primary
+	// key up to batchsize items after start (exclusive)
+	HighWatermark(ctx context.Context, domainID string, logID, start int64, batchSize int32) (int32, int64, error)
```
pavelkalinnikov: I feel like giving names to the output variables here (at least in the interface) would help to avoid any confusion later.

core/sequencer/server.go
```diff
@@ -131,55 +145,68 @@ func (s *Server) RunBatch(ctx context.Context, in *spb.RunBatchRequest) (*empty.
...
-	// reading the items to be processed twice.  Once, here in RunBatch and
-	// again in CreateEpoch.
-	sources := make(map[int64]*spb.MapMetadata_SourceSlice)
-	for sliceID, high := range highs {
-		sources[sliceID] = &spb.MapMetadata_SourceSlice{
-			LowestWatermark:  lastMeta.Sources[sliceID].GetHighestWatermark(),
+	sources := make(SourcesEntry)
+	for logID, high := range highs {
+		sources[logID] = &spb.MapMetadata_SourceSlice{
+			LowestWatermark:  lastMeta.Sources[logID].GetHighestWatermark(),
```
pavelkalinnikov: Is it the same as `startWatermarks[logID]`? Could be shortened.

core/sequencer/server.go
```diff
@@ -131,55 +145,68 @@ func (s *Server) RunBatch(ctx context.Context, in *spb.RunBatchRequest) (*empty.
...
-	}
-	if len(msgs) < int(in.MinBatch) {
-		return &empty.Empty{}, nil
+	//
+	// Rate limit the creation of new batches.
+	///
+
+	// TODO(#1057): If time since last map revision > max timeout, run batch.
+	// TODO(#1047): If time since oldest queue item > max latency has elapsed, run batch.
+	// If count items >= min_batch, run batch.
```
pavelkalinnikov: These 3 lines say "run batch", but `RunBatch` is the current function. Maybe reword a bit, or rename the function?

core/sequencer/server.go
```diff
@@ -131,55 +145,68 @@ func (s *Server) RunBatch(ctx context.Context, in *spb.RunBatchRequest) (*empty.
...
+			batch, err := s.logs.ReadLog(ctx, domainID, logID, low, source.GetHighestWatermark(), batchSize)
+			if err != nil {
+				return nil, status.Errorf(codes.Internal, "ReadLog(): %v", err)
+			}
+			for _, m := range batch {
+				msgs = append(msgs, &ktpb.EntryUpdate{
+					Mutation:  m.Mutation,
+					Committed: m.ExtraData,
+				})
+				if m.ID > low {
```
pavelkalinnikov: I have a feeling that `m.ID` will always be `>= low`, assuming `ReadLog` returns the keys in the sorted order (which is a reasonable thing to ask for, given the specifics of the query itself) so is this `if` needed?

core/sequencer/server.go
```diff
@@ -131,55 +145,68 @@ func (s *Server) RunBatch(ctx context.Context, in *spb.RunBatchRequest) (*empty.
...
-			return nil, status.Errorf(codes.Internal, "ReadQueue(): %v", err)
-		}
-		for _, m := range batch {
-			msgs = append(msgs, &ktpb.EntryUpdate{
-				Mutation:  m.Mutation,
-				Committed: m.ExtraData,
-			})
+		low := source.GetLowestWatermark()
+		// Loop until less than batchSize items are returned.
+		for count := readBatchSize; count == readBatchSize; {
```
pavelkalinnikov: I still don't get the relation between `readBatchSize` and `batchSize`. `readBatchSize` seems to me like a default value for `batchSize`.
Should you use `batchSize` here instead?

core/sequencer/server.go
```diff
@@ -131,55 +145,68 @@ func (s *Server) RunBatch(ctx context.Context, in *spb.RunBatchRequest) (*empty.
...
+			batch, err := s.logs.ReadLog(ctx, domainID, logID, low, source.GetHighestWatermark(), batchSize)
```
pavelkalinnikov: nit: Maybe store `source.GetHighestWatermark()` in a `high` variable, next to `low` above? It is also used below in a log message.

core/sequencer/server.go
```diff
@@ -131,55 +145,68 @@ func (s *Server) RunBatch(ctx context.Context, in *spb.RunBatchRequest) (*empty.
...
-		if err != nil {
-			return nil, status.Errorf(codes.Internal, "ReadQueue(): %v", err)
-		}
-		for _, m := range batch {
-			msgs = append(msgs, &ktpb.EntryUpdate{
-				Mutation:  m.Mutation,
-				Committed: m.ExtraData,
-			})
+		low := source.GetLowestWatermark()
+		// Loop until less than batchSize items are returned.
```
pavelkalinnikov: Not sure I get why the stopping condition is `count == batchSize`. What if the first log has `batchSize / 2`, and all the others have `batchSize * 100`?

core/sequencer/server.go
```diff
@@ -380,3 +410,46 @@ func (s *Server) applyMutations(domainID string, mutatorFunc mutator.Func,
 	glog.V(2).Infof("applyMutations applied %v mutations to %v leaves", len(msgs), len(leaves))
 	return ret, nil
 }
+
+// HighWatermarks returns the total count and the highest watermark for each log.
```
pavelkalinnikov: nit: "total count" is misleading - my first thought was total count of entries in each log. Maybe should be something like "total count of entries between the passed in low and the returned high watermarks"? Maybe swapping order of the returned values could help to make the comment shorter.

core/sequencer/server.go
```diff
@@ -380,3 +410,46 @@ func (s *Server) applyMutations(domainID string, mutatorFunc mutator.Func,
...
+// batchSize is a limit on the total number of items represented by the returned watermarks.
```
pavelkalinnikov: nit: s/batchSize/maxBatch/, or rename the variable.

core/sequencer/server.go
```diff
@@ -68,13 +68,27 @@ func createMetrics(mf monitoring.MetricFactory) {
...
+	HighWatermark(ctx context.Context, domainID string, logID, start int64, batchSize int32) (int32, int64, error)
+
+	// ListLogs returns the logIDs associated with domainID that have their write bit set.
+	// or all logIDs associated with domainID if writable is false.
+	ListLogs(ctx context.Context, domainID string, writable bool) ([]int64, error)
+
+	// ReadLog returns the lowest messages in the (low, high] range stored in the
+	// specified log, up to batchSize.  Paginate by setting low to the
+	// highest LogMessage returned in the previous page.
+	ReadLog(ctx context.Context, domainID string, logID, low, high int64,
```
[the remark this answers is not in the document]
gdbelvin: I do admit that it is odd an error prone - so I'm happy to revisit it. 

The rationale is that when reading from the log, we typically want to read ALL the available log items, which implies an inclusive read on `high` or returning `high + 1`.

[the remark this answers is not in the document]
gdbelvin: Oh, good point about 0.  For timestamps this is not an issue, but it will be for logs. Thanks for pointing that out.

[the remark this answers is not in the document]
gdbelvin: Should we make this change in a new PR?

core/sequencer/server.go
gdbelvin: `define batch`

core/sequencer/server.go
gdbelvin: yes, you're correct. I included the `if` out of an abundance of caution and correctness :-)

core/sequencer/server.go
gdbelvin: yes. This was left over from a refactor.

core/sequencer/server.go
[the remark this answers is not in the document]
pavelkalinnikov: I think so, yeah.

core/sequencer/server.go
gdbelvin: This is the looping condition *per log*. 
If the first log has `batchSize / 2`, that means there are no more batches in log 1 to process, we'll proceed to log 2, where we'll retrieve the first `batchSize` items.  Seeing that `count == batchSize`, we'll proceed to fetch the next `batchSize` items until there are less than `batchSize` items left in log 2 to fetch, at which point we'll move on to log 3 etc.

core/sequencer/server.go
gdbelvin: Yes, good catch.

core/sequencer/server.go
gdbelvin: hows "total count across all logs"

gdbelvin:
Updated with nits. PTAL

core/sequencer/server.go
```diff
@@ -68,13 +68,28 @@ func createMetrics(mf monitoring.MetricFactory) {
...
+type Watermarks map[int64]int64
+
 // LogsReader reads messages in multiple logs.
 type LogsReader interface {
-	// HighWatermarks returns the highest primary key for each log in the mutations table.
-	HighWatermarks(ctx context.Context, directoryID string) (map[int64]int64, error)
-	// ReadLog returns the messages in the (low, high] range stored in the specified log.
-	// ReadLog does NOT delete messages.
-	ReadLog(ctx context.Context, directoryID string, logID, low, high int64) ([]*mutator.LogMessage, error)
+	// Highwatermark returns the number of items and the highest primary
```
pavelkalinnikov: nit: HighWatermark

core/sequencer/server.go
```diff
@@ -68,13 +68,28 @@ func createMetrics(mf monitoring.MetricFactory) {
...
+	// key up to batchsize items after start (exclusive).
```
pavelkalinnikov: nit: batchSize

core/sequencer/server_test.go
```diff
@@ -15,16 +15,120 @@
...
 	"github.com/google/tink/go/signature"
 	"github.com/google/tink/go/tink"
-
-	ktpb "github.com/google/keytransparency/core/api/v1/keytransparency_go_proto"
 	tpb "github.com/google/trillian"
 )
 
+type fakeLogs map[int64][]mutator.LogMessage
+
+func (l fakeLogs) ReadLog(ctx context.Context, domainID string, logID, low, high int64,
```
pavelkalinnikov: Rename `domainID` to `directoryID` here and throughout this PR?

core/sequencer/server.go
```diff
@@ -191,7 +218,8 @@ func (s *Server) CreateEpoch(ctx context.Context, in *spb.CreateEpochRequest) (*
 	if in.MapMetadata.GetSources() == nil {
 		return nil, status.Errorf(codes.InvalidArgument, "missing map metadata")
 	}
-	msgs, err := s.readMessages(ctx, in.DirectoryId, in.MapMetadata.GetSources())
+	readBatchSize := int32(1000)
```
pavelkalinnikov: Maybe TODO making `readBatchSize` configurable?

core/sequencer/server.go
```diff
@@ -214,6 +242,7 @@ func (s *Server) CreateEpoch(ctx context.Context, in *spb.CreateEpochRequest) (*
 	}
 	glog.V(2).Infof("CreateEpoch: %v mutations, %v indexes", len(msgs), len(indexes))
 
+	// TODO(gbelvin): fetch map leaves at a specific revision
```
pavelkalinnikov: nit: **F**etch; period

core/sequencer/server_test.go
```diff
@@ -15,16 +15,120 @@
...
+
+func (l fakeLogs) HighWatermark(ctx context.Context, domainID string, logID, start int64,
+	batchSize int32) (int32, int64, error) {
+	high := start + int64(batchSize)
+	if high > int64(len(l[logID]))-1 {
+		high = int64(len(l[logID])) - 1
+	}
+	count := int32(high - start)
+	return count, high, nil
+
```
pavelkalinnikov: Remove blank line

PR #1076, opened 2018-10-22
Update the comment on EntryUpdate.committed field

therealdrake:
Updating comment to reference the correct field name.


core/api/v1/keytransparency.proto
```diff
@@ -43,7 +43,7 @@ message Committed {
 message EntryUpdate {
   // mutation authorizes the change to entry.
   Entry mutation = 2;
-  // commitment contains the data committed to in update.commitment.
+  // commitment contains the data committed to in mutation.commitment.
```
jtoohill:
```suggestion
  // committed contains the data committed to in mutation.commitment.
```

PR #1080, opened 2018-10-23
Remove unused UpdateEntryRequest.first_tree_size

gdbelvin:
`QueueEntryUpdate` used to return all the proof data from `GetEntry` which includes a consistency proof between `first_tree_size` and the current log root.  But since we are no longer returning this, this field is unused.


core/api/v1/keytransparency.proto
```diff
@@ -174,9 +174,8 @@ message UpdateEntryRequest {
   string user_id = 1;
   // app_id is the identifier for the application.
   string app_id = 2;
-  // first_tree_size is the tree_size of the currently trusted log root.
-  // Omitting this field will omit the log consistency proof from the response.
-  int64 first_tree_size = 3;
+  // Used to be first_tree_size. QueueEntryUpdate no longer returns any data.
```
jtoohill: I prefer to just put all reserved tags at the end of the message definition with a generic comment like:
```suggestion
  // Deprecated tag numbers, do not reuse.
```
Anyone interested in the history of what a tag number used to be can look in the commit history.

gdbelvin: Is it a strong preference?  It might be more friendly to leave a useful comment for at least a small period of time.

jtoohill: Nope, not a strong one. Approving as-is.

Issue #1077, opened 2018-10-23
Simplify domain/app/user tuple


jtoohill:
In real deployments the domain name can be the "domain", and each domain can host multiple trees.

The AppID+UserID path structure should be application-defined rather than baked into KT, since some apps might need more or less hierarchy.

PR #1082, opened 2018-10-25
Remove AppID from Key Transparency (#1077)

jtoohill:
Applications can define their own structure for UserID if needed.

Changes authz policies to be per-domain only; supporting finer-grained authz per UserID prefix can be added later if needed.

Also removes all PGP-specific logic for the "pgp" AppID. The application or frontend app server that uses Key Transparency can enforce its own constraints on key formats and serialization.


core/client/client.go
```diff
@@ -148,14 +148,14 @@ func (c *Client) updateTrusted(newTrusted *types.LogRootV1) {
 }
 
 // GetEntry returns an entry if it exists, and nil if it does not.
-func (c *Client) GetEntry(ctx context.Context, userID, appID string, opts ...grpc.CallOption) ([]byte, *types.LogRootV1, error) {
-	e, slr, err := c.VerifiedGetEntry(ctx, appID, userID)
+func (c *Client) GetEntry(ctx context.Context, userID string, opts ...grpc.CallOption) ([]byte, *types.LogRootV1, error) {
```
golangcibot: line is 122 characters

core/client/client.go
```diff
@@ -148,14 +148,14 @@ func (c *Client) updateTrusted(newTrusted *types.LogRootV1) {
...
+	e, slr, err := c.VerifiedGetEntry(ctx, userID)
 	return e.GetCommitted().GetData(), slr, err
 }
 
 // PaginateHistory iteratively calls ListHistory to satisfy the start and end requirements.
 // Returns a list of map roots and profiles at each revision.
-func (c *Client) PaginateHistory(ctx context.Context, appID, userID string, start, end int64) (map[uint64]*types.MapRootV1, map[uint64][]byte, error) {
+func (c *Client) PaginateHistory(ctx context.Context, userID string, start, end int64) (map[uint64]*types.MapRootV1, map[uint64][]byte, error) {
```
golangcibot: line is 144 characters

core/client/get_and_verify.go
```diff
@@ -109,13 +108,12 @@ func (c *Client) VerifiedGetEpoch(ctx context.Context, epoch int64) (*types.LogR
 }
 
 // VerifiedListHistory performs one list history operation, verifies and returns the results.
-func (c *Client) VerifiedListHistory(ctx context.Context, appID, userID string, start int64, count int32) (map[*types.MapRootV1][]byte, int64, error) {
+func (c *Client) VerifiedListHistory(ctx context.Context, userID string, start int64, count int32) (map[*types.MapRootV1][]byte, int64, error) {
```
golangcibot: line is 144 characters

core/api/v1/keytransparency.proto
```diff
@@ -310,23 +305,23 @@ service KeyTransparency {
   // Entries contain signed commitments to a profile, which is also returned.
   // TODO(gbelvin): Replace with GetUser
   rpc GetEntry(GetEntryRequest) returns (GetEntryResponse) {
-    option (google.api.http) = { get: "/v1/directories/{directory_id}/apps/{app_id}/users/{user_id}" };
+    option (google.api.http) = { get: "/v1/directories/{directory_id}/users/{user_id}" };
```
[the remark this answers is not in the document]
jtoohill: Since {user_id} is the trailing resource ID, it can capture a value that contains forward slashes. I think your suggestion would put the string "users/alice@domain.com" into user_id instead of just "alice@domain.com".

I moved the /history path segment earlier for ListEntryHistory. I dropped :queue from QueueEntryUpdate, since just using the post HTTP verb is enough to distingush it.

core/api/v1/keytransparency.proto
```diff
@@ -264,7 +264,7 @@ message ListMutationsResponse {
 //    that user, and signatures indicating who made the last change.
 //
 // Each User also has a collection of historical values for user:
-// `/v1/directories/*/users/*/history`
+// `/v1/history/directories/*/users/*`
```
gdbelvin: do you think history belongs underneath directories? Would love to hear your thoughts on pro/cons/

core/api/v1/keytransparency.proto
```diff
@@ -313,15 +313,15 @@ service KeyTransparency {
...
+    option (google.api.http) = { get: "/v1/history/directories/{directory_id}/users/{user_id}" };
   }
 
   // QueueUserUpdate enqueues an update to a user's profile.
   //
   // Clients should poll GetEntry until the update appears, and retry if no update appears after a timeout.
   rpc QueueEntryUpdate(UpdateEntryRequest) returns (google.protobuf.Empty) {
     option (google.api.http) = {
-      post: "/v1/directories/{directory_id}/users/{user_id}:queue"
+      post: "/v1/directories/{directory_id}/users/{user_id}"
```
gdbelvin: The `post` doesn't really look like one of the standard methods in 
https://cloud.google.com/apis/design/standard_methods, which is why it had the custom verb per
https://cloud.google.com/apis/design/custom_methods. What do you think?

core/api/v1/keytransparency.proto
jtoohill: I don't see much difference between the two, but I'm also not a resource-oriented design expert.

What about a third option - that :listHistory is a custom verb on the directory+user resource?

gdbelvin: I'm not an expert either. Taking a hint from Google Drive's [API](https://developers.google.com/drive/api/v3/reference/revisions), revisions seems like it should come after `user_id`, which, as you say is screwing us a bit in that it means we can't have `user_id` as the last URI segement. 

* Get a user's current public key
`/v1/directories/{directory_id}/users/{user_id}`
* Get a user's public key at a specific revision
`/v1/directories/{directory_id}/users/{user_id}/revisions/{revision_id}/`
* List a user's public keys through time. 
`/v1/directories/{directory_id}/users/{user_id}/revisions/`

gdbelvin:
This is worth another brainstorming chat. I'm trying to keep the following principles in mind:

1. List collections should end on a plural resourceID. ie. ListRevisions should end on `.../revisions`. 
2. There's a missing API for fetching a user at a particular revision. ie. `.../revisions/{revision_id}`
3. We will probably want to have multiple path segments as part of `user_id`. 
   a. formats
   b. identifiers
   c. kinds

I wonder if we can square some of these circles by pre-determining the number of segments that can go into `user_id`. ie. 
`/v1/directories/{directory_id}/users/{user_id=*/*/*}/revisions/{revision_id}`
concretely: 
`/v1/directories/{directory_id}/users/e164/{e164 number}/keys/revisions/{revision_id}`

jtoohill:
Good points Gary. Rather than pre-determining the number of segments in `user_id`, I'm leaning towards not allowing forward slashes in `user_id` and keeping the current URI structure (minus `app_id` of course).

We can make the change from /history to /revisions in a follow-up PR if desired.

core/crypto/vrf/vrf.go
```diff
@@ -41,12 +41,10 @@ type PublicKey interface {
 	ProofToHash(m, proof []byte) (index [32]byte, err error)
 }
 
-// UniqueID computes a unique string for a directory, userID and appID combo.
-func UniqueID(userID, appID string) []byte {
+// UniqueID computes a unique string for a userID.
+func UniqueID(userID string) []byte {
 	b := new(bytes.Buffer)
```
gdbelvin: We can delete `UniqueID` now.  It's only purpose was to create a canonical string for the VRF function. 
VRF callers can pass in `userID` directly in place of `UniqueID`

PR #1084, opened 2018-10-25
Golang cleanup in admin server.

therealdrake:
Various golang cleanup:
-Move multi-line object creation outside of function arguments
-Fix import alias naming to be consistent
-Simplify logic where possible
-Use the New func in test code instead of creating Server object directly.


core/adminserver/admin_server.go
```diff
@@ -195,11 +188,9 @@ func treeConfig(treeTemplate *tpb.CreateTreeRequest, privKey *any.Any, directory
 	}
 
 	config.Tree.Description = fmt.Sprintf("KT directory %s", directoryID)
-	maxDisplayNameLen := 20
```
gdbelvin: I try to avoid magic numbers. Perhaps this should be moved into a package constant?

core/adminserver/admin_server_test.go
```diff
@@ -64,14 +64,8 @@ func newMiniEnv(ctx context.Context, t *testing.T) (*miniEnv, error) {
 	if err != nil {
 		return nil, fmt.Errorf("error starting fake server: %v", err)
 	}
-	srv := &Server{
```
gdbelvin: Be careful here, I don't think I'm actually setting all the variables in server - there's a few I don't need for this test. 

`New` can work, but you'll need to pass `nil` for a few items.

core/adminserver/admin_server_test.go
```diff
@@ -53,9 +54,8 @@ type miniEnv struct {
 
 func newMiniEnv(ctx context.Context, t *testing.T) (*miniEnv, error) {
 	fakeDirectories := fake.NewDirectoryStorage()
-	if err := fakeDirectories.Write(ctx, &directory.Directory{
-		DirectoryID: "existingdirectory",
-	}); err != nil {
+	dir := &directory.Directory{DirectoryID: "existingdirectory"}
```
gdbelvin: You've detected a pattern of mine: passing arguments as an immediate.
Is there a more general pattern of not doing this that I should be aware of? 
If so, there's a lot of other code sites that could also use this change. (for other PRs)

gdbelvin:
Also, we typically try to have meaningful text in the PR description box if you wouldn't mind adding a list of the kinds of changes you made.

core/adminserver/admin_server_test.go
therealdrake: I prefer this because I find it generally more readable. My usual argument for this is..
If: You need to multi-line when constructing arguments in the function call.
Then: It must be the same number of lines to construct the input object before the function call.

core/adminserver/admin_server_test.go
```diff
@@ -64,14 +64,8 @@ func newMiniEnv(ctx context.Context, t *testing.T) (*miniEnv, error) {
...
 	}
-	srv := &Server{
-		tlog:        s.LogClient,
-		tmap:        s.MapClient,
-		logAdmin:    s.AdminClient,
-		mapAdmin:    s.AdminClient,
-		directories: fakeDirectories,
-		keygen:      vrfKeyGen,
-	}
+	srv := New(s.LogClient, s.MapClient, s.AdminClient, s.AdminClient, fakeDirectories, vrfKeyGen)
```
golangcibot: cannot use vrfKeyGen (value of type func(ctx context.Context, spec *github.com/google/trillian/crypto/keyspb.Specification) (github.com/golang/protobuf/proto.Message, error)) as LogsAdmin value in argument to New: missing method AddLogs

PR #1091, opened 2018-10-31
Clean up GetEntry API as GetUser

jtoohill:
Epoch already defines the grouping of fields needed to verify a signed map head within a log. We can define a sibling message to group the fields needed to verify a leaf within a signed map head.

This will allow us to define a BatchGetUser API later on (#588) which returns a single Epoch and multiple MapLeafEntries.


core/client/gobindclient/client.go
```diff
@@ -116,18 +116,18 @@ func AddKtServer(ktURL string, insecureTLS bool, ktTLSCertPEM []byte, directoryI
...
 		return nil, fmt.Errorf("A connection to %v does not exists. Please call AddKtServer first", ktURL)
 	}
 
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
-	entry, _, err := client.GetEntry(ctx, userID)
+	entry, _, err := client.GetUser(ctx, userID)
 	if err != nil {
-		return nil, fmt.Errorf("GetEntry failed: %v", err)
+		return nil, fmt.Errorf("GetUser failed: %v", err)
```
golangcibot: error strings should not be capitalized or end with punctuation or a newline

core/api/v1/keytransparency.proto
```diff
@@ -92,59 +92,37 @@ message MapperMetadata {
...
   // directory_id identifies the directory in which the user lives.
-  string directory_id = 4;
-  // user_id is the user identifier, the format of which is defined by the
-  // application.
-  string user_id = 1;
-  // Used to be app_id. Applications can define their own hierarchy for user_id
-  // if desired.
-  reserved 2;
+  string directory_id = 1;
+  // user_id is the user identifier, the format for which is defined by the application.
```
gdbelvin: Double check line length here. I think the previous formatter split this line.

core/api/v1/keytransparency.proto
```diff
@@ -92,59 +92,37 @@ message MapperMetadata {
...
 
-// GetEntryResponse returns a requested user entry.
-message GetEntryResponse {
-  //
-  // Privacy layer hides user_id and profile data until requested.
-  //
-
-  // vrf_proof is the proof for VRF on user_id.
+// Leaf entry for a user.
+message MapLeafEntry {
```
gdbelvin: s/MapLeafEntry/MapLeaf/ ?

core/api/v1/keytransparency.proto
```diff
@@ -92,59 +92,37 @@ message MapperMetadata {
...
+  // vrf_proof is the proof for the VRF on user_id.
   bytes vrf_proof = 1;
+  // map_inclusion is an inclusion proof for the map leaf in an accompanying
+  // trillian.SignedMapRoot. If the leaf is non-empty, its leaf.leaf_value
+  // stores a serialized Entry proto.
+  trillian.MapLeafInclusion map_inclusion = 2;
+  // committed contains the data and nonce used to make a cryptographic
```
gdbelvin: FYI committed is serialized inside of map_inclusion.map_leaf.extra_data, and is deserialized here for convenience.

PR #1102, opened 2018-11-08
 Split Epoch into MapRoot and LatestLogRoot 

### LINKED ISSUE #1100: Define Primitives
- Split Epoch

gdbelvin:
Fixes #1100

Epoch contains two pieces of information that update independently:
- The latest log root which contains a consistency proof.
  API end points will only return one of these.
- A map root which contains an inclusion proof to the lastest log root.
  API end points may return multiple map roots, one for each revision
  requested


core/api/v1/keytransparency.proto
```diff
@@ -209,21 +209,36 @@ message GetLatestEpochRequest {
...
-  repeated bytes log_consistency = 4;
-  // log_inclusion proves that map_root is part of log_root at
-  // index=map_root.MapRevision.
-  repeated bytes log_inclusion = 5;
+  // map_root contains the map root and its inclusion in the log.
+  MapRoot map_root = 5;
+  // latest_log_root contains the most recent log root and its consistency
+  // proof to the client's last seen log root.
+  LogRoot latest_log_root = 6;
+  reserved 2, 3, 4;
```
jtoohill: `// Deprecated tag numbers, do not reuse.`

PR #1104, opened 2018-11-08
API: Rename Epoch to Revision

gdbelvin:
`Revision` is more consistent with `trilian.Map.Revision`. 

Key Transparency places each Trillian Map Revision into a Log of Map Revisions.  This MapRoot, LogInclusion pair is a KT `Revision`.


core/client/client.go
```diff
@@ -181,37 +181,37 @@ func (c *Client) PaginateHistory(ctx context.Context, userID string, start, end
...
+	sort.Sort(revisions)
 
 	// Compress profiles that are equal through time.  All
 	// nil profiles before the first profile are ignored.
 	var prevData []byte
-	var prevEpoch uint64
+	var prevRevision uint64
 	ret := make(map[uint64][]byte)
-	for i, e := range epochs {
+	for i, e := range revisions {
```
therealdrake: r instead of e?

core/client/client.go
```diff
@@ -181,37 +181,37 @@ func (c *Client) PaginateHistory(ctx context.Context, userID string, start, end
...
 		// Verify that the roots are contiguous
-		if i != 0 && e != prevEpoch+1 {
-			glog.Errorf("Non contiguous history. Got epoch %v, want %v", e, prevEpoch+1)
+		if i != 0 && e != prevRevision+1 {
```
therealdrake: r instead of e?

core/keyserver/epochs_test.go
```diff
@@ -185,7 +185,7 @@ func TestListMutations(t *testing.T) {
 
 			resp, err := e.srv.ListMutations(ctx, &pb.ListMutationsRequest{
 				DirectoryId: directoryID,
-				Epoch:       epoch,
+				Revision:       revision,
```
therealdrake: spacing seems off

Issue #1101, opened 2018-11-08
Refactor ListEntryHistory as ListUserRevisions


jtoohill:
- [ ] s/ListEntryHistory/ListUserRevisions/
- [ ] Deduplicate proofs in response (depends on #1100)
- [ ] Specify start/end revisions
- [ ] Pagination tokens

PR #1110, opened 2018-11-09
Add ListUserRevisions API

jtoohill:
This will eventually replace the ListEntryHistory API (#1101)


core/api/v1/keytransparency.proto
```diff
@@ -372,6 +414,15 @@ service KeyTransparency {
       get: "/v1/directories/{directory_id}/users/{user_id}/history"
     };
   }
+  // ListUserRevisions returns a list of historic leaf values for a user.
+  //
+  // Clients verify their account history by observing correct values for their
+  // account over time.
+  rpc ListUserRevisions(ListUserRevisionsRequest) returns (ListUserRevisionsResponse) {
+    option (google.api.http) = {
+      get: "/v1/directories/{directory_id}/users/{user_id}/revisions"
```
therealdrake: needs to be a post I think

core/keyserver/keyserver.go
```diff
@@ -273,6 +273,12 @@ func (s *Server) ListEntryHistory(ctx context.Context, in *pb.ListEntryHistoryRe
 	}, nil
 }
 
+// ListUserRevisions returns a list of revisions covering a period of time.
+func (s *Server) ListUserRevisions(ctx context.Context, in *pb.ListUserRevisionsRequest) (
```
therealdrake: the multi-lining here seems weird. Could you move the opening parens down to the line below?

core/api/v1/keytransparency.proto
jtoohill: Done.

core/keyserver/keyserver.go
jtoohill: If you don't leave the hanging open parens, Go thinks you forgot the `{` at the start of the function definition (doesn't compile).

PR #1111, opened 2018-11-09
API: BatchListUserRevisions

gdbelvin:
Proto defs for #1109


core/api/v1/keytransparency.proto
```diff
@@ -129,6 +129,14 @@ message MapLeaf {
   Committed committed = 3;
 }
 
+// BatchMapRevision contains a map leaf at a speific revision.
```
therealdrake: "a set of map leaves"?

core/api/v1/keytransparency.proto
```diff
@@ -187,6 +195,35 @@ message ListEntryHistoryResponse {
...
 }
 
+// BatchListUserRevisionsRequest contains a list of users and a range of revisions.
+message BatchListUserRevisionsRequest {
+  // directory_id identifies the directory in which the user lives.
+  string directory_id = 1;
+  // user_ids are the user identifiers.
+  repeated string user_ids = 2;
+  // start is the starting revision.
+  int64 start = 3;
```
therealdrake: start_revision to match the non-batch request's field names

core/api/v1/keytransparency.proto
```diff
@@ -187,6 +195,35 @@ message ListEntryHistoryResponse {
   int64 next_start = 2;
 }
 
+// BatchListUserRevisionsRequest contains a list of users and a range of revisions.
+message BatchListUserRevisionsRequest {
+  // directory_id identifies the directory in which the user lives.
+  string directory_id = 1;
+  // user_ids are the user identifiers.
+  repeated string user_ids = 2;
```
therealdrake: what's the behavior if a user_id doesn't exist in the map? Does the API handle that gracefully using an empty leaf or is it an error case we should call out?

core/api/v1/keytransparency.proto
```diff
@@ -187,6 +195,35 @@ message ListEntryHistoryResponse {
...
+  // verified. Omitting this field will omit the log consistency proof from the
+  // response.
+  int64 last_verified_tree_size = 5;
+}
+
+// BatchListUserRevisionsResponse
+message BatchListUserRevisionsResponse {
+  // latest_log_root contains the lastest log root and its consistency proof.
+  LogRoot latest_log_root = 1;
+  // map_revisions represents the list of keys this user_id has contained over
```
therealdrake: is 'keys' the right word here? maybe something more generic like "leaf data"?

gdbelvin:
@jtoohill Should I mark this is as blocked on #1110 ?

jtoohill:
> @jtoohill Should I mark this is as blocked on #1110 ?

I don't think so - that change has been merged.

core/api/v1/keytransparency.proto
gdbelvin: Good question.  Queries for leaves that don't exist in the Trillian Map produce proofs to empty map leaves, not gRPC errors.

core/api/v1/keytransparency.proto
```diff
@@ -129,6 +129,14 @@ message MapLeaf {
   Committed committed = 3;
 }
 
+// BatchMapRevision contains a map leaf at a speific revision.
+message BatchMapRevision {
+  // map_root contains the map root and its inclusion in the log.
+  MapRoot map_root = 1;
+  // map_leaf contains a leaf and its inclusion proof to map_root.
+  repeated MapLeaf map_leaf = 2;
```
jtoohill: map_leaves

core/api/v1/keytransparency.proto
jtoohill: > `this user_id has`

the requested user_ids have

core/api/v1/keytransparency.proto
```diff
@@ -187,6 +195,35 @@ message ListEntryHistoryResponse {
...
+
+// BatchListUserRevisionsResponse
+message BatchListUserRevisionsResponse {
+  // latest_log_root contains the lastest log root and its consistency proof.
+  LogRoot latest_log_root = 1;
+  // map_revisions represents the list of keys this user_id has contained over
+  // time. At most page_size results will be returned.
+  repeated BatchMapRevision map_revisions = 2;
+  // next_page_token is a pagination token which will be set if more than
+  // page_size results are available. Clients can pass this value as
```
jtoohill: Should the count of results per page be measured in revisions, or revisions * users?

gdbelvin: That's a great question... I suppose we can answer that a couple of way:

a) In terms of API round trips to Trillian, we have 1 API call per revision because the Trillian API exposes a batch get map leaves at revision API. 
b) In terms of bandwidth and DB access the costs per API call are more similar to revisions * users. 

Since this is a *list* of revisions, I think page size should be measured in revisions. 
Costs are typically more of a quota kind of thing, which I think we should charge in terms of revisions * users. Furthermore, when we implement more efficient ways of querying ranges of revisions we can reduce the quota charge without messing with the API semantics.

gdbelvin:
Updated protos. PTAL

PR #1112, opened 2018-11-11
Define pagination token format for ListUserRevisions

jtoohill:
Needed for #1101


core/keyserver/readtoken.proto
```diff
@@ -18,10 +18,25 @@ package google.keytransparency.v1;
...
+// when listing revisions.
+message ListUserRevisionsToken {
+  // request is the query being paginated over, used for validation of
+  // subsequent requests. Fields that are allowed to change between requests
+  // (such as page_token or last_verified_tree_size) will not be validated and
+  // should be omitted for brevity.
+  ListUserRevisionsRequest request = 1;
+  // revisions_returned is a running tally of the number of revisions that have
+  // been returned across paginated requests in this query.
+  int64 revisions_returned = 2;
```
gdbelvin: Is `revisions_returned` more like `offset` or a statistic?

I found this article to be a nice review of various paging techniques and their tradeoffs: 
https://slack.engineering/evolving-api-pagination-at-slack-1c1f644f8e12

For our use case we can efficiently add `start` + `revisions_returned` to get our read cursor, but from a design perspective, we might want to train ourselves to use read cursors to start with.

jtoohill: I was thinking `revisions_returned` would be used like an `offset` in practice.

Would the read cursor be passed through to Trillian? If so, it seems like we'd need to implement batched+paged reads from Trillian first, which is an area I know nothing about yet.

gdbelvin: At the moment, Trillian only offers [`GetLeavesByRevision`](https://github.com/google/trillian/blob/master/trillian_map_api.proto#L125) so (for the moment), the keyserver is generating individual requests for each revision and passing them to Trillian. 

When Trillian implements `GetLeavesByRevisionRange` we can transition / add an opaque paging token to this proto that we can pass through. 

`revisions_returned` makes sense for now. The client logic should be exactly the same in both cases since this whole proto will be an opaque token.

PR #1114, opened 2018-11-14
Implement ListUserRevisions

### LINKED ISSUE #1101: Refactor ListEntryHistory as ListUserRevisions

jtoohill:
Fixes #1101. Still needs integration tests, and should eventually use a batch API in Trillian for efficiency.

### DIFF
--- a/core/keyserver/keyserver.go
+++ b/core/keyserver/keyserver.go
@@ -277,7 +277,7 @@ func (s *Server) ListEntryHistory(ctx context.Context, in *pb.ListEntryHistoryRe
 // ListUserRevisions returns a list of revisions covering a period of time.
 func (s *Server) ListUserRevisions(ctx context.Context, in *pb.ListUserRevisionsRequest) (
 	*pb.ListUserRevisionsResponse, error) {
-	offset := int64(0)
+	pageStart := in.StartRevision
 	lastVerified := in.GetLastVerifiedTreeSize()
 	if in.GetPageToken() != "" {
 		token := &rtpb.ListUserRevisionsToken{}
@@ -292,7 +292,7 @@ func (s *Server) ListUserRevisions(ctx context.Context, in *pb.ListUserRevisions
 		if !proto.Equal(in, token.Request) {
 			return nil, status.Errorf(codes.InvalidArgument, "Request fields changed during pagination")
 		}
-		offset = token.RevisionsReturned
+		pageStart += token.RevisionsReturned
 	}
 
 	// Lookup log and map info.
@@ -317,9 +317,9 @@ func (s *Server) ListUserRevisions(ctx context.Context, in *pb.ListUserRevisions
 		return nil, err
 	}
 
-	numRevisions, err := validateListUserRevisionsRequest(in, offset, newestRevision)
+	numRevisions, err := validateListUserRevisionsRequest(in, pageStart, newestRevision)
 	if err != nil {
-		glog.Errorf("validateListUserRevisionsRequest(%v, %v, %v): %v", in, offset, newestRevision, err)
+		glog.Errorf("validateListUserRevisionsRequest(%v, %v, %v): %v", in, pageStart, newestRevision, err)
 		return nil, status.Errorf(codes.InvalidArgument, "Invalid request")
 	}
 
@@ -327,7 +327,7 @@ func (s *Server) ListUserRevisions(ctx context.Context, in *pb.ListUserRevisions
 	// Get all revisions in the range [start + offset, start + offset + numRevisions].
 	revisions := make([]*pb.MapRevision, numRevisions)
 	for i := range revisions {
-		rev := in.StartRevision + offset + int64(i)
+		rev := pageStart + int64(i)
 		resp, err := s.getUserByRevision(ctx, sth, d, in.UserId, rev)
 		if err != nil {
 			glog.Errorf("getUser failed for revision %v: %v", rev, err)
@@ -341,24 +341,26 @@ func (s *Server) ListUserRevisions(ctx context.Context, in *pb.ListUserRevisions
 
 	// Add a page token to the response if more revisions can be fetched.
 	token := ""
-	if in.StartRevision+offset+numRevisions < in.EndRevision {
-		token, err = EncodeToken(&rtpb.ListUserRevisionsToken{
+	if pageStart+numRevisions < in.EndRevision {
+		tokenProto := &rtpb.ListUserRevisionsToken{
 			Request:           in,
-			RevisionsReturned: offset + numRevisions,
-		})
+			RevisionsReturned: (pageStart - in.StartRevision) + numRevisions,
+		}
+		token, err = EncodeToken(tokenProto)
 		if err != nil {
 			glog.Errorf("error encoding page token: %v", err)
 			return nil, status.Errorf(codes.Internal, "Error encoding pagination token")
 		}
 	}
-	return &pb.ListUserRevisionsResponse{
+	resp := &pb.ListUserRevisionsResponse{
 		LatestLogRoot: &pb.LogRoot{
 			LogRoot:        sth,
 			LogConsistency: consistencyProof.GetHashes(),
 		},
 		MapRevisions:  revisions,
 		NextPageToken: token,
-	}, nil
+	}
+	return resp, nil
 }
 
 // BatchListUserRevisions returns a list of revisions covering a period of time.
--- a/core/keyserver/keyserver.go
+++ b/core/keyserver/keyserver.go
@@ -313,7 +313,7 @@ func (s *Server) ListUserRevisions(ctx context.Context, in *pb.ListUserRevisions
 	}
 	currentRevision, err := mapRevisionFor(sth)
 	if err != nil {
-		glog.Errorf("latestRevision(log %v, sth%v): %v", d.LogID, sth, err)
+		glog.Errorf("latestRevision(log %v, sth %v): %v", d.LogID, sth, err)
 		return nil, err
 	}
 
@@ -339,21 +339,14 @@ func (s *Server) ListUserRevisions(ctx context.Context, in *pb.ListUserRevisions
 		}
 	}
 
-	resp := &pb.ListUserRevisionsResponse{
-		LatestLogRoot: &pb.LogRoot{
-			LogRoot:        sth,
-			LogConsistency: consistencyProof.GetHashes(),
-		},
-		MapRevisions: revisions,
-	}
-
 	// Add a page token to the response if more revisions can be fetched.
 	end := currentRevision
 	if in.EndRevision != 0 {
 		end = in.EndRevision
 	}
+	token := ""
 	if in.StartRevision+offset+numRevisions < end {
-		token, err := EncodeToken(&rtpb.ListUserRevisionsToken{
+		token, err = EncodeToken(&rtpb.ListUserRevisionsToken{
 			Request:           in,
 			RevisionsReturned: int64(offset + numRevisions),
 		})
@@ -361,9 +354,15 @@ func (s *Server) ListUserRevisions(ctx context.Context, in *pb.ListUserRevisions
 			glog.Errorf("error encoding page token: %v", err)
 			return nil, status.Errorf(codes.Internal, "Error encoding pagination token")
 		}
-		resp.NextPageToken = token
 	}
-	return resp, nil
+	return &pb.ListUserRevisionsResponse{
+		LatestLogRoot: &pb.LogRoot{
+			LogRoot:        sth,
+			LogConsistency: consistencyProof.GetHashes(),
+		},
+		MapRevisions:  revisions,
+		NextPageToken: token,
+	}, nil
 }
 
 // BatchListUserRevisions returns a list of revisions covering a period of time.
--- a/core/keyserver/keyserver.go
+++ b/core/keyserver/keyserver.go
@@ -31,6 +31,7 @@ import (
 	"google.golang.org/grpc/status"
 
 	pb "github.com/google/keytransparency/core/api/v1/keytransparency_go_proto"
+	rtpb "github.com/google/keytransparency/core/keyserver/readtoken_go_proto"
 	spb "github.com/google/keytransparency/core/sequencer/sequencer_go_proto"
 	tpb "github.com/google/trillian"
 )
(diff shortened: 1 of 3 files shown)


core/keyserver/keyserver.go
```diff
@@ -276,7 +277,93 @@ func (s *Server) ListEntryHistory(ctx context.Context, in *pb.ListEntryHistoryRe
...
+	lastVerified := in.GetLastVerifiedTreeSize()
+	if in.GetPageToken() != "" {
+		token := &rtpb.ListUserRevisionsToken{}
+		if err := DecodeToken(in.GetPageToken(), token); err != nil {
+			glog.Errorf("invalid page token %v: %v", in.GetPageToken(), err)
+			return nil, status.Errorf(codes.InvalidArgument, "Invalid page_token provided")
+		}
+		// last_verified_tree_size and page_token are allowed to change between paginated requests.
+		// Clear them here both for comparison and for encoding next_page_token in the response.
+		in.LastVerifiedTreeSize = 0
```
therealdrake: if you set this to 0 you lose this information which you need. I think you can pass an option to the proto comparison to ignore certain fields.

core/keyserver/keyserver.go
```diff
@@ -276,7 +277,93 @@ func (s *Server) ListEntryHistory(ctx context.Context, in *pb.ListEntryHistoryRe
...
+	}
+
+	// Fetch latest revision.
+	sth, consistencyProof, err := s.latestLogRootProof(ctx, d, lastVerified)
+	if err != nil {
+		return nil, err
+	}
+	currentRevision, err := mapRevisionFor(sth)
+	if err != nil {
+		glog.Errorf("latestRevision(log %v, sth%v): %v", d.LogID, sth, err)
```
therealdrake: add a space after sth

core/keyserver/keyserver.go
```diff
@@ -276,7 +277,93 @@ func (s *Server) ListEntryHistory(ctx context.Context, in *pb.ListEntryHistoryRe
...
+			return nil, status.Errorf(codes.InvalidArgument, "Invalid page_token provided")
+		}
+		// last_verified_tree_size and page_token are allowed to change between paginated requests.
+		// Clear them here both for comparison and for encoding next_page_token in the response.
+		in.LastVerifiedTreeSize = 0
+		in.PageToken = ""
+		if !proto.Equal(in, token.Request) {
+			return nil, status.Errorf(codes.InvalidArgument, "Request fields changed during pagination")
+		}
+		offset = token.RevisionsReturned
```
therealdrake: instead of tracking offset this way should we just set in.StartRevision += token.RevisionsReturned

core/keyserver/keyserver.go
```diff
@@ -276,7 +277,93 @@ func (s *Server) ListEntryHistory(ctx context.Context, in *pb.ListEntryHistoryRe
...
+	}
+
+	// Add a page token to the response if more revisions can be fetched.
+	end := currentRevision
+	if in.EndRevision != 0 {
+		end = in.EndRevision
+	}
+	if in.StartRevision+offset+numRevisions < end {
+		token, err := EncodeToken(&rtpb.ListUserRevisionsToken{
+			Request:           in,
```
therealdrake: same number of lines to create the token proto in a variable above

core/keyserver/keyserver.go
```diff
@@ -276,7 +277,93 @@ func (s *Server) ListEntryHistory(ctx context.Context, in *pb.ListEntryHistoryRe
...
+		glog.Errorf("adminstorage.Read(%v): %v", directoryID, err)
+		return nil, status.Errorf(codes.Internal, "Cannot fetch directory info")
+	}
+
+	// Fetch latest revision.
+	sth, consistencyProof, err := s.latestLogRootProof(ctx, d, lastVerified)
+	if err != nil {
+		return nil, err
+	}
+	currentRevision, err := mapRevisionFor(sth)
```
therealdrake: "lastRevision" or "endRevision" seems more correct wdyt?

therealdrake: add a consistency check here that currentRevision == in.EndRevision if in.EndRevision != 0?

core/keyserver/keyserver.go
```diff
@@ -276,7 +277,93 @@ func (s *Server) ListEntryHistory(ctx context.Context, in *pb.ListEntryHistoryRe
...
+			LogRoot:        sth,
+			LogConsistency: consistencyProof.GetHashes(),
+		},
+		MapRevisions: revisions,
+	}
+
+	// Add a page token to the response if more revisions can be fetched.
+	end := currentRevision
+	if in.EndRevision != 0 {
+		end = in.EndRevision
```
therealdrake: this should == currentRevision if it's not 0 right?

core/keyserver/keyserver.go
```diff
@@ -276,7 +277,93 @@ func (s *Server) ListEntryHistory(ctx context.Context, in *pb.ListEntryHistoryRe
...
+			return nil, status.Errorf(codes.Internal, "GetUser failed")
+		}
+		revisions[i] = &pb.MapRevision{
+			MapRoot: resp.GetRevision().GetMapRoot(),
+			MapLeaf: resp.GetLeaf(),
+		}
+	}
+
+	resp := &pb.ListUserRevisionsResponse{
+		LatestLogRoot: &pb.LogRoot{
```
therealdrake: does this fit on one line?

core/keyserver/keyserver_test.go
```diff
@@ -139,15 +139,19 @@ func TestLatestRevision(t *testing.T) {
...
 				e.s.Log.EXPECT().GetInclusionProof(gomock.Any(), gomock.Any()).
-					Return(&tpb.GetInclusionProofResponse{}, nil)
+					Return(&tpb.GetInclusionProofResponse{}, nil).Times(2)
 			}
 
 			_, err = e.srv.ListEntryHistory(ctx, &pb.ListEntryHistoryRequest{DirectoryId: directoryID})
 			if got, want := status.Code(err), tc.wantErr; got != want {
 				t.Errorf("ListEntryHistory(): %v, want %v", err, tc.wantErr)
 			}
+			_, err = e.srv.ListUserRevisions(ctx, &pb.ListUserRevisionsRequest{DirectoryId: directoryID})
```
therealdrake: why not validate the response?

core/keyserver/validate.go
```diff
@@ -100,6 +103,34 @@ func validateListEntryHistoryRequest(in *pb.ListEntryHistoryRequest, currentRevi
 	return nil
 }
 
+// validateListUserRevisionsRequest checks the bounds on start and end revisions and returns an appropriate number of
+// revisions to return for this request.
+func validateListUserRevisionsRequest(in *pb.ListUserRevisionsRequest, offset, currentRevision int64) (int64, error) {
```
therealdrake: can we not use int instead of int64 all throughout?  int will default to 64 bit on 64-bit machines.

core/keyserver/validate.go
```diff
@@ -100,6 +103,34 @@ func validateListEntryHistoryRequest(in *pb.ListEntryHistoryRequest, currentRevi
...
+	if in.EndRevision != 0 {
+		if in.EndRevision < in.StartRevision || in.EndRevision > currentRevision {
+			return 0, ErrInvalidEnd
+		}
+		currentRevision = in.EndRevision
+	}
+
+	revisions := int64(in.PageSize)
+	switch {
+	case in.PageSize < 0:
```
therealdrake: it seems like you could pretty safely treat ==0 and <0 as the same case

core/keyserver/validate.go
```diff
@@ -100,6 +103,34 @@ func validateListEntryHistoryRequest(in *pb.ListEntryHistoryRequest, currentRevi
...
+		return 0, fmt.Errorf("Invalid page size")
+	case in.PageSize == 0:
+		revisions = int64(defaultPageSize)
```
therealdrake: I would expect defaultPageSize to be maxPageSize. wdyt?

core/keyserver/keyserver.go
jtoohill: I store it in lastVerified and use that instead.

I also had a vague recollection that you can compare ignoring fields, but I think that's a contrib library (and maybe only in Java / C++).

core/keyserver/keyserver.go
jtoohill: Done.

core/keyserver/keyserver.go
jtoohill: We're fetching the newest revision, so I think "current" makes sense.

core/keyserver/keyserver.go
jtoohill: Moved into the return statement.

core/keyserver/validate.go
jtoohill: Using int64 eliminates most of the casting (which IMO makes it more readable), since the field type in the proto is int64.

core/keyserver/keyserver_test.go
jtoohill: There's no client code to perform validation yet - I'll add that plus an integration test in another PR.

core/keyserver/validate.go
jtoohill: 0 means the page size wasn't specified in the request so we can use a default, negative would be an invalid request.

core/keyserver/keyserver.go
jtoohill: Done.

core/keyserver/validate.go
jtoohill: Currently default page size is 16 entries (~2KB response data per request) and max is 2048 entries (~8MB response data per request). I think these are fine defaults to have for now, we can always change them or simplify later.

core/keyserver/validate.go
jtoohill: I don't think any proto scalar types map onto Golang's int: https://developers.google.com/protocol-buffers/docs/proto3#scalar

core/keyserver/keyserver.go
jtoohill: Changed validation to require in.EndRevision be set explicitly.

core/keyserver/keyserver.go
jtoohill: Changed validation to require in.EndRevision be set explicitly.

jtoohill: Renamed.

core/keyserver/keyserver.go
```diff
@@ -276,7 +277,88 @@ func (s *Server) ListEntryHistory(ctx context.Context, in *pb.ListEntryHistoryRe
...
+		}
+	}
+	return &pb.ListUserRevisionsResponse{
+		LatestLogRoot: &pb.LogRoot{
+			LogRoot:        sth,
+			LogConsistency: consistencyProof.GetHashes(),
+		},
+		MapRevisions:  revisions,
+		NextPageToken: token,
+	}, nil
```
therealdrake: create the proto on the line above for readability?

core/keyserver/keyserver.go
therealdrake: doesn't look done :/

core/keyserver/validate.go
```diff
@@ -100,6 +103,31 @@ func validateListEntryHistoryRequest(in *pb.ListEntryHistoryRequest, currentRevi
...
+	case in.PageSize > maxPageSize:
+		revisions = int64(maxPageSize)
+	}
+	if in.StartRevision+offset+revisions > in.EndRevision {
+		revisions = in.EndRevision - (in.StartRevision + offset) + 1
```
therealdrake: it seems like you mostly make use of in.StartRevision + offset rather than either of them alone. Should we just store them in a "pageStart" variable to reduce the number of terms?

therealdrake: I think it might simplify things if we just initialize revisions to be in.EndRevision - (in.StartRevision + offset) + 1 and then just check if the page size we want is is smaller than that in the switch statement.

core/keyserver/keyserver.go
therealdrake: fair enough. Should we just rename offset to "pageStart" and set it to in.StartRevision + token.RevisionsReturned? The start of the page seems to be what we want throughout the rest of the func anyway.

core/keyserver/keyserver.go
jtoohill: Done.

core/keyserver/validate.go
jtoohill: Done.

jtoohill: We'd have to introduce another temporary to store the page size (default or max), so it doesn't end up simplifying things. I think the flow of starting with `PageSize` revisions, then limiting the results if it goes beyond `EndRevision` makes sense.

core/keyserver/keyserver.go
jtoohill: Done.

core/keyserver/keyserver.go
jtoohill: Oops, actually done this time.

core/keyserver/validate.go
```diff
@@ -100,6 +103,31 @@ func validateListEntryHistoryRequest(in *pb.ListEntryHistoryRequest, currentRevi
...
+		return 0, ErrInvalidStart
+	}
+	if in.EndRevision < in.StartRevision || in.EndRevision > newestRevision {
+		return 0, ErrInvalidEnd
+	}
+
+	revisions := int64(in.PageSize)
+	switch {
+	case in.PageSize < 0:
+		return 0, fmt.Errorf("Invalid page size")
```
golangcibot: error strings should not be capitalized or end with punctuation or a newline

core/keyserver/keyserver_test.go
```diff
@@ -139,15 +139,22 @@ func TestLatestRevision(t *testing.T) {
...
+					Return(&tpb.GetInclusionProofResponse{}, nil).Times(2)
 			}
 
 			_, err = e.srv.ListEntryHistory(ctx, &pb.ListEntryHistoryRequest{DirectoryId: directoryID})
 			if got, want := status.Code(err), tc.wantErr; got != want {
 				t.Errorf("ListEntryHistory(): %v, want %v", err, tc.wantErr)
 			}
+			_, err = e.srv.ListUserRevisions(ctx, &pb.ListUserRevisionsRequest{
+				DirectoryId: directoryID,
+				EndRevision: tc.treeSize - 1,
```
golangcibot: Using the variable on range scope `tc` in function literal

core/keyserver/keyserver_test.go
```diff
@@ -139,15 +139,22 @@ func TestLatestRevision(t *testing.T) {
...
+			})
+			if got, want := status.Code(err), tc.wantErr; got != want {
```
golangcibot: Using the variable on range scope `tc` in function literal

core/keyserver/keyserver_test.go
```diff
@@ -139,15 +139,22 @@ func TestLatestRevision(t *testing.T) {
...
+				t.Errorf("ListUserRevisions(): %v, want %v", err, tc.wantErr)
```
golangcibot: Using the variable on range scope `tc` in function literal

core/keyserver/keyserver.go
```diff
@@ -276,7 +277,90 @@ func (s *Server) ListEntryHistory(ctx context.Context, in *pb.ListEntryHistoryRe
 // ListUserRevisions returns a list of revisions covering a period of time.
 func (s *Server) ListUserRevisions(ctx context.Context, in *pb.ListUserRevisionsRequest) (
 	*pb.ListUserRevisionsResponse, error) {
-	return nil, status.Error(codes.Unimplemented, "not implemented")
+	pageStart := in.StartRevision
+	lastVerified := in.GetLastVerifiedTreeSize()
```
gdbelvin: The `Get` functions are useful if the object before the `.` (`in` in this case) might be `nil`. 
Otherwise, in proto3, all objects have a default value. 

For consistency, consider always using `Get` or trying to save `Get` for cases where it might be needed.

core/keyserver/keyserver.go
```diff
@@ -276,7 +277,90 @@ func (s *Server) ListEntryHistory(ctx context.Context, in *pb.ListEntryHistoryRe
...
+			return nil, status.Errorf(codes.InvalidArgument, "Invalid page_token provided")
+		}
+		// last_verified_tree_size and page_token are allowed to change between paginated requests.
+		// Clear them here both for comparison and for encoding next_page_token in the response.
+		in.LastVerifiedTreeSize = 0
+		in.PageToken = ""
+		if !proto.Equal(in, token.Request) {
+			return nil, status.Errorf(codes.InvalidArgument, "Request fields changed during pagination")
+		}
+		pageStart += token.RevisionsReturned
```
gdbelvin: It would be slightly easier to read to keep pageStart with it's original meaning and do the addition at the call sites.

core/keyserver/keyserver.go
```diff
@@ -276,7 +277,90 @@ func (s *Server) ListEntryHistory(ctx context.Context, in *pb.ListEntryHistoryRe
...
+			MapLeaf: resp.GetLeaf(),
+		}
+	}
+
+	// Add a page token to the response if more revisions can be fetched.
+	token := ""
+	if pageStart+numRevisions < in.EndRevision {
+		tokenProto := &rtpb.ListUserRevisionsToken{
+			Request:           in,
+			RevisionsReturned: (pageStart - in.StartRevision) + numRevisions,
```
gdbelvin: with the original meaning of page start, this could become
`offset + numRevisions`

gdbelvin:
LGTM. Would you like me to merge this?

jtoohill:
Yep, should be ready.

PR #1117, opened 2018-11-20
Implement CreateRevision as a Beam Transform

gdbelvin:
* `CreateRevision` is an idempotent way to convert a list of mutations into a trillian map revision. 
* `Beam` is the "not yet ready", cross platform, Map Reduce [SDK](https://docs.google.com/document/d/1yj0_hxq2J1iestjFUUrm_BVQLsFxQiiqtcFhgodzIgM/edit).

This PR is structured so that projects that do not want to take a dependency on beam can build flag exclude `beam.go` with `go build --tag nobeam`. 

Files:
- `beam.go` contains the beam pipeline and the beam direct runner. 
- `beam_channels.go` contains the same pipeline implemented with go channels. 
- `beam_fns.go` contains the individual worker functions executed by both pipelines.

This PR does not include logic for executing the beam pipeline on a distributed platform such as [Google Cloud DataFlow](https://cloud.google.com/dataflow/). That will come later, after the Trillian Map can accept sharded write requests for the same revision.

TODO:
- [ ] Output metrics about the number of map leaves written, mutations read, errored mutations. 
- [ ] Make sure the pipeline doesn't panic when it encounters errors midway through.


core/sequencer/beam_test.go
```diff
@@ -0,0 +1,236 @@
...
+		{desc: "two", meta: &spb.MapMetadata{
+			Sources: map[int64]*spb.MapMetadata_SourceSlice{
+				1: {HighestWatermark: 2},
+			},
+		}},
+	} {
+		t.Run(tc.desc, func(t *testing.T) {
+			mw1 := &mapWrites{}
+			if err := createRevisionWithBeam(ctx, in.DirectoryId, in.Revision,
+				tc.meta, mw1, mr, lr); err != nil {
```
golangcibot: Using the variable on range scope `tc` in function literal

core/sequencer/beam_test.go
```diff
@@ -0,0 +1,236 @@
...
+				t.Errorf("createRevisionWithBeam(): %v", err)
+			}
+			mw2 := &mapWrites{}
+			if err := createRevisionWithChannels(ctx, in.DirectoryId, in.Revision,
+				tc.meta, mw2, mr, lr); err != nil {
```
golangcibot: Using the variable on range scope `tc` in function literal

core/sequencer/beam_test.go
```diff
@@ -0,0 +1,236 @@
...
+			{LeafValue: []byte("def")},
+		}},
+	} {
+
+		p := beam.NewPipeline()
+		s := p.Root()
+
+		// Split inputs and then recombine them.
+		leaves := beam.ParDo(s, func(in []*tpb.MapLeaf, emit func(*tpb.MapLeaf)) {
+			for _, l := range tc.in {
```
golangcibot: Using the variable on range scope `tc` in function literal

core/sequencer/beam_channels.go
```diff
@@ -0,0 +1,287 @@
...
+
+func goApply(rows <-chan joinRow) <-chan *tpb.MapLeaf {
+	c := make(chan *tpb.MapLeaf)
+	go func() {
+		for r := range rows {
+			var leafIndex int
+			var msgIndex int
+			applyMutation(r.index,
+				func(e **tpb.MapLeaf) bool {
+					if leafIndex < len(r.leaves) {
```
golangcibot: Using the variable on range scope `r` in function literal

core/sequencer/beam_channels.go
```diff
@@ -0,0 +1,287 @@
...
+						*e = r.leaves[leafIndex]
```
golangcibot: Using the variable on range scope `r` in function literal

core/sequencer/beam_channels.go
```diff
@@ -0,0 +1,287 @@
...
+						leafIndex++
+						return true
+					}
+					return false
+				},
+				func(e **ktpb.EntryUpdate) bool {
+					if msgIndex < len(r.msgs) {
```
golangcibot: Using the variable on range scope `r` in function literal

core/sequencer/server.go
```diff
@@ -100,6 +100,13 @@ type Batcher interface {
 	ReadBatch(ctx context.Context, directoryID string, rev int64) (*spb.MapMetadata, error)
 }
 
+// CreateRevFn creates a new map revision given an input batch
+type CreateRevFn func(ctx context.Context, dirID string, rev int64,
+	meta *spb.MapMetadata, mw MapWriter, mr MapReader, lr LogReader) error
+
+// createRev is the function used to create new revisions.
+var createRev CreateRevFn = createRevisionWithChannels
```
golangcibot: `createRev` is a global variable

core/sequencer/beam_channels.go
```diff
@@ -0,0 +1,287 @@
...
+		close(entries)
+	}()
+	return entries
+}
+
+func goMapLogItems(logItems <-chan *ktpb.EntryUpdate) <-chan indexMutation {
+	c := make(chan indexMutation)
+	go func() {
+		for e := range logItems {
+			mapLogItem(e, func(index []byte, update *ktpb.EntryUpdate) {
```
golangcibot: Error return value of `mapLogItem` is not checked

core/sequencer/beam_channels.go
```diff
@@ -0,0 +1,287 @@
...
+	}()
+	return c
+}
+
+func goReadMap(ctx context.Context, indexSets <-chan [][]byte, dirID string, mr MapReader) <-chan indexLeaf {
+	c := make(chan indexLeaf)
+	go func() {
+		// There should only be one index set.
+		for indexes := range indexSets {
+			mr.ReadMap(ctx, indexes, dirID, func(index []byte, leaf *tpb.MapLeaf) {
```
golangcibot: Error return value of `mr.ReadMap` is not checked

core/sequencer/beam_channels.go
```diff
@@ -0,0 +1,287 @@
...
+	return c
+}
+
+func goApply(rows <-chan joinRow) <-chan *tpb.MapLeaf {
+	c := make(chan *tpb.MapLeaf)
+	go func() {
+		for r := range rows {
+			var leafIndex int
+			var msgIndex int
+			applyMutation(r.index,
```
golangcibot: Error return value of `applyMutation` is not checked

core/sequencer/beam.go
```diff
@@ -0,0 +1,82 @@
...
+	"context"
+
+	"github.com/apache/beam/sdks/go/pkg/beam"
+	"github.com/apache/beam/sdks/go/pkg/beam/x/beamx"
+
+	spb "github.com/google/keytransparency/core/sequencer/sequencer_go_proto"
+	tpb "github.com/google/trillian"
+)
+
+func init() {
```
golangcibot: don't use `init` function

core/sequencer/beam_channels.go
```diff
@@ -0,0 +1,287 @@
...
+			ls := logSource{LogID: logID, Source: source}
+			glog.V(2).Infof("LogSource: %v", ls)
+			metaChan <- ls
+		})
+		close(metaChan)
+	}()
+	return metaChan
+}
+
+func goReadOneLog(ctx context.Context, logs <-chan logSource, dirID string, batchSize int32, lr LogReader) <-chan *ktpb.EntryUpdate {
```
golangcibot: line is 133 characters

core/sequencer/beam_channels.go
```diff
@@ -0,0 +1,287 @@
...
+		for l := range leaves {
+			accum = mergeFn.AddInput(accum, l)
+		}
+		c <- mergeFn.ExtractOutput(accum)
+		close(c)
+	}()
+	return c
+}
+
+func goWriteMap(ctx context.Context, leafSets <-chan []*tpb.MapLeaf, meta *spb.MapMetadata, dirID string, mw MapWriter) error {
```
golangcibot: line is 127 characters

core/sequencer/beam_test.go
```diff
@@ -0,0 +1,236 @@
...
+		})
+	}
+}
+
+type mapWrites struct {
+	Leaves []*tpb.MapLeaf
+	Called bool
+}
+
+func (m *mapWrites) WriteMap(ctx context.Context, leaves []*tpb.MapLeaf, meta *spb.MapMetadata, directoryID string) error {
```
golangcibot: line is 123 characters

core/sequencer/beam_test.go
```diff
@@ -0,0 +1,236 @@
...
+	return nil
+}
+
+func TestDontPanic(t *testing.T) {
+	ctx := context.Background()
+	p := beam.NewPipeline()
+	s := p.Root()
+
+	errDontPanic := errors.New("don't panic")
+	ints := beam.ParDo(s, func(i []byte, e func(int)) error {
```
golangcibot: `TestDontPanic$1` - `i` is unused

core/sequencer/beam_test.go
```diff
@@ -0,0 +1,236 @@
...
+			{LeafValue: []byte("abc")},
+			{LeafValue: []byte("def")},
+		}},
+	} {
+
+		p := beam.NewPipeline()
+		s := p.Root()
+
+		// Split inputs and then recombine them.
+		leaves := beam.ParDo(s, func(in []*tpb.MapLeaf, emit func(*tpb.MapLeaf)) {
```
golangcibot: `TestEmptyRevision$1` - `in` is unused

core/sequencer/beam_test.go
```diff
@@ -0,0 +1,239 @@
...
+
+		meta := beam.Create(scope, tc.meta)
+		// Read each logID in parallel.
+		sourceSlices := beam.ParDo(scope, splitMeta, meta) // KV<logID, source>
+		logItems := beam.ParDo(scope, s.ReadLog, sourceSlices,
+			beam.SideInput{Input: beam.Create(scope, directoryID)},
+			beam.SideInput{Input: beam.Create(scope, tc.batchSize)})
+		count := stats.Sum(scope, beam.DropKey(scope, stats.Count(scope, logItems)))
+		beam.ParDo0(scope, func(got int) error {
+			if got != tc.want {
```
golangcibot: Using the variable on range scope `tc` in function literal

core/sequencer/beam_test.go
```diff
@@ -0,0 +1,239 @@
...
+				return fmt.Errorf("readMessages(): len: %v, want %v", got, tc.want)
```
golangcibot: Using the variable on range scope `tc` in function literal

core/sequencer/beam_test.go
```diff
@@ -0,0 +1,239 @@
...
+	return nil
+}
+
+func TestDontPanic(t *testing.T) {
+	ctx := context.Background()
+	p := beam.NewPipeline()
+	s := p.Root()
+
+	errDontPanic := errors.New("don't panic")
+	ints := beam.ParDo(s, func(_ []byte, e func(int)) error {
```
golangcibot: `TestDontPanic$1` - `e` is unused

core/sequencer/beam_test.go
```diff
@@ -0,0 +1,239 @@
...
+
+		meta := beam.Create(scope, tc.meta)
+		// Read each logID in parallel.
+		sourceSlices := beam.ParDo(scope, splitMeta, meta) // KV<logID, source>
+		logItems := beam.ParDo(scope, s.ReadLog, sourceSlices,
+			beam.SideInput{Input: beam.Create(scope, directoryID)},
+			beam.SideInput{Input: beam.Create(scope, tc.batchSize)})
+		count := stats.Sum(scope, beam.DropKey(scope, stats.Count(scope, logItems)))
+		beam.ParDo0(scope, func(got int) error {
+			if got != tc.want {
```
golangcibot: Using the variable on range scope `tc` in function literal

core/sequencer/beam_test.go
```diff
@@ -0,0 +1,239 @@
...
+	return nil
+}
+
+func TestDontPanic(t *testing.T) {
+	ctx := context.Background()
+	p := beam.NewPipeline()
+	s := p.Root()
+
+	errDontPanic := errors.New("don't panic")
+	ints := beam.ParDo(s, func(_ []byte, e func(int)) error {
```
golangcibot: `TestDontPanic$1` - `e` is unused

pavelkalinnikov:
While I am reading this, couple of general things:

1. It would be nice to have a high-level picture/description of the pipeline.
2. If you build with `nobeam` tag, does it mean no beam dependency on any level; or it means you don't have to deploy a Beam-compatible system, but the build still depends on Beam on the package level, and KT provides some default implementation of its interfaces?

core/sequencer/beam.go
```diff
@@ -0,0 +1,82 @@
...
+	// Read each logID in parallel.
+	sourceSlices := beam.ParDo(scope, splitMeta, meta) // KV<logID, source>
+	logItems := beam.ParDo(scope, lr.ReadLog, sourceSlices,
+		beam.SideInput{Input: dirID},
+		beam.SideInput{Input: beam.Create(scope, batchSize)}) // *ktpb.EntryUpdate
+
+	keyedMutations := beam.ParDo(scope, mapLogItem, logItems) // KV<index, *ktpb.EntryUpdate>
+
+	// Read the map
+	indexes := beam.Combine(scope, &mergeIndexFn{}, beam.DropValue(scope, keyedMutations)) // []index
```
pavelkalinnikov: Does this line return a PCollection containing just one item -- the full list of indices? It can be millions, right?

core/sequencer/beam.go
```diff
@@ -0,0 +1,82 @@
...
+	mapLeaves := beam.ParDo(scope, mr.ReadMap, indexes,
```
pavelkalinnikov: And then this line does only one call to `ReadMap`, and puts the full list of indices into it?

pavelkalinnikov: As discussed in person, this is indeed the case, so KT instance that processes this one []index item is a bottleneck in the pipeline. There should be some sharding in place instead -- could be implemented with `beam.Partition`, for example.

The downside of `Partition` is that the number of shards has to be a "constant" that can't depend on the input `PCollection`, so we can't decrease the number of shards as we get less input. But it can be a flag, and maybe it's okay to restart the binary when we expect the load to change.

Also, I am not sure it's impossible to "block" on `PCollection` and read the value from it in the middle of the pipeline. If it is possible then we can use `Partition` with input-dependent number of shards.

pavelkalinnikov: Finally, note that we could implement our own version of `Partition` that takes `N` as a side input.

core/sequencer/beam.go
```diff
@@ -0,0 +1,82 @@
...
+		beam.SideInput{Input: dirID},
+		beam.SideInput{Input: beam.Create(scope, rev-1)}) // KV<index, *tpb.MapLeaf>
+
+	// Align MapLeaves with their mutations and apply mutations.
+	joined := beam.CoGroupByKey(scope, mapLeaves, keyedMutations) // []*tpb.MapLeaf, []*ktpb.EntryUpdate
```
pavelkalinnikov: Is CoGroupByKey implemented already? The doc linked in this PR description [mentions](https://docs.google.com/document/d/1yj0_hxq2J1iestjFUUrm_BVQLsFxQiiqtcFhgodzIgM/edit?disco=AAAABeSShNY) it was not in Jan.

core/sequencer/beam_fns.go
```diff
@@ -0,0 +1,223 @@
...
+	oldValue, err := entry.FromLeafValue(oldMapLeaf.LeafValue)
+	if err != nil {
+		glog.Warningf("entry.FromLeafValue(%v): %v", oldMapLeaf.LeafValue, err)
+		return err
+	}
+
+	var newEntry *ktpb.SignedEntry
+	var committed *ktpb.Committed
+	var msg *ktpb.EntryUpdate
+	for getMessage(&msg) {
```
pavelkalinnikov: The order of entries returned from this function seems not deterministic to me, which makes the output of this pipeline not idempotent.

pavelkalinnikov: Ah, so line 141 resolves the conflict, right? Sorry, seems like I just proved to myself that this code works, and could totally omit writing these comments :)

The only outstanding question is how okay is it to silently drop a mutation just because there is a conflicting one? Does this sort of thing need to be logged / monitored?

core/sequencer/beam_fns.go
```diff
@@ -0,0 +1,223 @@
...
+		// TODO(gbelvin): Modify mutation function to accept this function's signature.
+		newValue, err := entry.New().Mutate(oldValue, msg.Mutation)
+		if err != nil {
+			glog.Warningf("Mutate(): %v", err)
+			continue // Filter for valid mutations.
+		}
+		// The order of mutations needs to be associative we need to break ties without regard to the order of messages.
+		if bytes.Compare(newValue.Entry, newEntry.GetEntry()) > 0 {
```
pavelkalinnikov: I wonder if this conflict should be resolved in favor of the "latest" mutation is some sense. Couple of valid reasons for having conflicts in the first place can be:

1. Client submits a mutation and fails / disconnects, then shortly after it submits a new mutation.
2. Client submits a mutation and says "oups, wrong one", then shortly after it submits a new one.

In both cases there is a chance for both mutations getting to the same revision.

core/sequencer/beam_fns.go
```diff
@@ -0,0 +1,223 @@
...
+// increments of batchSize, emitting *ktpb.EntryUpdates as it goes.
+func (s *Server) ReadLog(ctx context.Context, logID int64, source *spb.MapMetadata_SourceSlice,
+	directoryID string, batchSize int32, emit func(*ktpb.EntryUpdate)) error {
+	low := source.GetLowestWatermark()
+	high := source.GetHighestWatermark()
+	// Loop until less than batchSize items are returned.
+	for count := batchSize; count == batchSize; {
+		batch, err := s.logs.ReadLog(ctx, directoryID, logID, low, high, batchSize)
+		if err != nil {
+			return status.Errorf(codes.Internal, "ReadLog(): %v", err)
```
pavelkalinnikov: I wonder how the pipeline reacts to errors returned from transforms? Does it abort the whole processing, skip the batch, restart it, or smth else?

core/sequencer/server.go
```diff
@@ -100,13 +100,41 @@ type Batcher interface {
...
+// MapWriter writes to the trillian Map.
+type MapWriter interface {
+	// WriteMap sends leaves to the map for directoryID.
+	WriteMap(ctx context.Context, leaves []*tpb.MapLeaf, meta *spb.MapMetadata, directoryID string) error
+}
+
+// MapReader reads a set of indexes from the map.
+type MapReader interface {
+	// ReadMap emits one map leaf for every index requested.
+	ReadMap(ctx context.Context, indexes [][]byte, directoryID string, rev int64,
```
pavelkalinnikov: Prereq: see "DoFn and other user functions" -> "(2) Struct with methods" and DoFn [godoc](https://godoc.org/github.com/apache/beam/sdks/go/pkg/beam#hdr-DoFns). Apparently, Beam API allows `DoFn` be a struct implementing their interface.

An alternative approach to yours is to rename `ReadMap` to `ProcessElement`, and drop `directoryID` and `rev` from this signature. Sending these 2 parameters via side input doesn't look reasonable to me as those are construction time parameters for the pipeline. Instead you can assume that `MapReader` implementations know about `directoryID` and `rev`.

If you do that then `MapReaderFn` could be a better name.

pavelkalinnikov: Same reasoning applies to all interfaces in this file.

core/sequencer/beam.go
```diff
@@ -0,0 +1,82 @@
...
+	tpb "github.com/google/trillian"
+)
+
+// CreateRevisionWithBeam is an implementation of CreateRevFn using beam.
+func CreateRevisionWithBeam(ctx context.Context, directoryID string, rev int64,
+	metaProto *spb.MapMetadata, batchSize int32,
+	mw MapWriter, mr MapReader, lr LogReader) error {
+
+	p := beam.NewPipeline()
+	scope := p.Root()
```
pavelkalinnikov: You only use this `scope` throughout the whole pipeline. Might be reasonable to `scope.Scope("blah")` for each logical step in your pipeline.

core/sequencer/beam_fns.go
```diff
@@ -0,0 +1,223 @@
...
+func splitMeta(meta *spb.MapMetadata, emit func(logID int64, source *spb.MapMetadata_SourceSlice)) {
+	for logID, source := range meta.Sources {
+		glog.Infof("emit(logID %v, source %v)", logID, source)
+		emit(logID, source)
+	}
+}
+
+// ReadLog reads from source.LowestWatermark to source.HighestWatermark in
+// increments of batchSize, emitting *ktpb.EntryUpdates as it goes.
+func (s *Server) ReadLog(ctx context.Context, logID int64, source *spb.MapMetadata_SourceSlice,
```
pavelkalinnikov: observation: You only need `s.logs`, not sure if it makes sense to make is `Server`'s method rather than passing `s.logs` in directly.
concern: `ReadLog` is supposed to be serialized and distributed among the pipeline workers, isn't it? Will it be able to distribute the `Server` object and/or `s.logs`?

core/sequencer/beam.go
pavelkalinnikov: ... unless Beam magically embeds function names into it.

core/sequencer/beam.go
```diff
@@ -0,0 +1,82 @@
...
+
+	"github.com/apache/beam/sdks/go/pkg/beam"
+	"github.com/apache/beam/sdks/go/pkg/beam/x/beamx"
+
+	spb "github.com/google/keytransparency/core/sequencer/sequencer_go_proto"
+	tpb "github.com/google/trillian"
+)
+
+// CreateRevisionWithBeam is an implementation of CreateRevFn using beam.
+func CreateRevisionWithBeam(ctx context.Context, directoryID string, rev int64,
```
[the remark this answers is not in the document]
gdbelvin: I'm building the end-to-end Key Transparency Sequencer flow as a series of retryable, idempotent steps.

This distributed step of the process occurs *after* we have committed to the metadata for a particular revision.

core/sequencer/beam.go
```diff
@@ -0,0 +1,82 @@
...
+	metaProto *spb.MapMetadata, batchSize int32,
+	mw MapWriter, mr MapReader, lr LogReader) error {
+
+	p := beam.NewPipeline()
```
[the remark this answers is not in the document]
gdbelvin: A single pipeline is used in the streaming case as well.  See this [example](https://github.com/apache/beam/blob/master/sdks/go/examples/streaming_wordcap/wordcap.go)

I'm thinking of having this function return the pipeline so that it can be used by a variety of runners.

core/sequencer/beam.go
gdbelvin: fair enough

pavelkalinnikov:
As discussed in person, it would make sense to break down this RP again, so that each pipeline transform goes into an individual PR, and is tested separately. It's probably a good idea to start from the end of the pipeline (which defines the output we want), and unwind it back towards the data source.

pavelkalinnikov:
Obsolete?

PR #1123, opened 2018-11-22
Wrap all Trillian Client logic.  

gdbelvin:
Considates creating Trillian Log and Map clients for specific domains behind 
* `MapClient(domainID string)`
* `LogClient(domainID string)`

And it creates verifying / wrapping functions for a few Trillian APIs.
These methods are good candidates for moving the underlying trillian thick clients.
* `SetAndVerifyLeaves()`
* `GetAndVerifyLatestMapRoot` to return `rawMapRoot`
* `GetAndVerifyMapRootByRevision` to return `rawMapRoot`


core/sequencer/trillian_client.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	mapRoot, err := c.VerifySignedMapRoot(rootResp.GetMapRoot())
+	if err != nil {
+		return nil, nil, status.Errorf(codes.Internal, "VerifySignedMapRoot(%v): %v", c.MapID, err)
+	}
+	return rootResp.GetMapRoot(), mapRoot, nil
+}
+
+// GetAndVerifyMapRootByRevision verifies and returns a specific map root.
+func (c *MapClient) GetAndVerifyMapRootByRevision(ctx context.Context, rev int64) (*tpb.SignedMapRoot, *types.MapRootV1, error) {
```
golangcibot: line is 129 characters

Martin2112:
I don't know all the motivation for this but it looks to me like there are too many interfaces and possibly something simpler will do the same job.

There are also two fairly unrelated things in this PR, which makes it harder to see what's going on.

Martin2112:
BTW it might be easier if you explain the thinking behind this. But I'm also not around for the next few days.

core/sequencer/trillian_client.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/google/keytransparency/core/directory"
+	"github.com/google/keytransparency/core/keyserver"
+
+	ktpb "github.com/google/keytransparency/core/api/v1/keytransparency_go_proto"
+	tpb "github.com/google/trillian"
+	tclient "github.com/google/trillian/client"
+)
+
+// TrillianFactory creates verifying clients for Trillian
+type TrillianFactory interface {
```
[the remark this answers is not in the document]
gdbelvin: I've removed the factory from this PR.

core/sequencer/trillian_client.go
```diff
@@ -0,0 +1,151 @@
...
+	MapClient(ctx context.Context, dirID string) (TrillianMap, error)
+	LogClient(ctx context.Context, dirID string) (TrillianLog, error)
+}
+
+// TrillianMap communicates with the Trilian map and verifies the responses.
+type TrillianMap interface {
```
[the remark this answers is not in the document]
gdbelvin: I've removed these interfaces  as well

core/sequencer/trillian_client.go
```diff
@@ -0,0 +1,151 @@
...
+
+// TrillianLog communicates with the Trillian log and verifies the responses.
+type TrillianLog interface {
+	WaitForInclusion(ctx context.Context, data []byte) error
+	UpdateRoot(ctx context.Context) (*types.LogRootV1, error)
+	AddSequencedLeaf(ctx context.Context, data []byte, index int64) error
+}
+
+// TrillianConns contains connections to Trillian. Implements TrillianFactory.
+type TrillianConns struct {
```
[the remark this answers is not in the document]
gdbelvin: Renamed to `Trillian` to make `Trillian.MapClient` easier on the eyes.

core/sequencer/trillian_client.go
```diff
@@ -0,0 +1,151 @@
...
+	directories directory.Storage
```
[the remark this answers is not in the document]
gdbelvin: `directories` is a list of KeyTransparency Directories. 
Each directory contains a Log TreeID and Map TreeID. 

The `MapClient(directoryID string)` function looks up the relevant directory in the `directories` object, fetches the public key for the TrillianMap, and returns a verifying `MapClient`

gdbelvin:
Related PR: google/trillian#1358

core/sequencer/trillian_client.go
```diff
@@ -0,0 +1,132 @@
...
+	mapAdmin    tpb.TrillianAdminClient
+	tmap        tpb.TrillianMapClient
+	tlog        tpb.TrillianLogClient
+}
+
+// MapClient returns a verifying MapClient
+func (t *Trillian) MapClient(ctx context.Context, dirID string) (*MapClient, error) {
+	directory, err := t.directories.Read(ctx, dirID, false)
+	if err != nil {
+		glog.Errorf("adminstorage.Read(%v): %v", dirID, err)
```
pavelkalinnikov: nit: `directories.Read` ?

core/sequencer/trillian_client.go
```diff
@@ -0,0 +1,132 @@
...
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "Cannot fetch map info for %v: %v", dirID, err)
+	}
+
+	c, err := tclient.NewMapClientFromTree(t.tmap, mapTree)
+	if err != nil {
+		return nil, err
+	}
+	return &MapClient{
+		MapClient: c,
```
pavelkalinnikov: nit: Could be a one-line `return`.

core/sequencer/trillian_client.go
```diff
@@ -0,0 +1,132 @@
...
+	}, nil
+}
+
+// LogClient returns a verifying LogClient.
+func (t *Trillian) LogClient(ctx context.Context, dirID string) (*tclient.LogClient, error) {
+	directory, err := t.directories.Read(ctx, dirID, false)
+	if err != nil {
+		glog.Errorf("adminstorage.Read(%v): %v", dirID, err)
```
pavelkalinnikov: nit: `directories.Read` ?

core/sequencer/trillian_client.go
```diff
@@ -0,0 +1,132 @@
...
+		return nil, status.Errorf(codes.Internal, "Cannot fetch directory info for %v", dirID)
+	}
+	logTree, err := t.logAdmin.GetTree(ctx, &tpb.GetTreeRequest{TreeId: directory.LogID})
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "Cannot fetch log info for %v: %v", dirID, err)
+	}
+
+	// Create verifying log and map clients.
```
pavelkalinnikov: Please remove maps from this sentence. Maybe even drop the comment, as the return line is self-describing?

core/sequencer/trillian_client.go
```diff
@@ -0,0 +1,132 @@
...
+	trustedRoot := types.LogRootV1{} // TODO(gbelvin): Store and track trustedRoot.
+	return tclient.NewFromTree(t.tlog, logTree, trustedRoot)
+}
+
+// MapClient implements TrillianMap.
```
pavelkalinnikov: Does `TrillianMap` interface exist? Couldn't find it.

core/sequencer/trillian_client.go
```diff
@@ -0,0 +1,132 @@
...
+// TODO(gbelvin): Move to Trillian Map client.
+type MapClient struct {
+	*tclient.MapClient
+}
+
+// SetLeaves creates a new map revision and returns its verified root.
```
pavelkalinnikov: The idea of parallel revision creation is in **not** using a method that creates a revision based on the complete set of mutations, but instead do multiple calls that are combined into one revision, right? I'd defer putting this method to Trillian API for now.

core/sequencer/trillian_client.go
```diff
@@ -0,0 +1,132 @@
...
+		return nil, err
+	}
+	return &MapClient{
+		MapClient: c,
+	}, nil
+}
+
+// LogClient returns a verifying LogClient.
+func (t *Trillian) LogClient(ctx context.Context, dirID string) (*tclient.LogClient, error) {
+	directory, err := t.directories.Read(ctx, dirID, false)
```
pavelkalinnikov: At least one of the functions above in this PR create both `LogClient` and `MapClient`. Both call `directories.Read`. Is performance not a concern in those use-cases?

core/sequencer/trillian_client.go
gdbelvin: It does in the next PR.  I'll remove it from this one.

core/sequencer/trillian_client.go
gdbelvin: That is indeed true. The `SetLeaves` API would need to change significantly.  This method might be more analogous to `FinalizeRevision`.

core/sequencer/trillian_client.go
gdbelvin: I expect that querying `directories.Read` to be much cheaper than `admin.GetTree`. 
Previously I was querying `GetTree` and then not using the results.

gdbelvin:
Updated with nits, PTAL

core/sequencer/trillian_client.go
pavelkalinnikov: Please resolve this before submitting.

PR #1124, opened 2018-11-23
Replace a Map of LogSources with a List of LogSources

gdbelvin:
A list of `[]*spb.MapMetadata_SourceSlice` is a lot easier to manage and iterate through than `map[int64]*spb.MapMetadata_SourceSlice`. 

* The list of sources going into a map does need to be one contiguous range per log.  One can imagine ingesting multiple, non-contiguous ranges from a single log too. 
* It has a defined order for iteration which is useful when we are paging. 
* Lists can be natively emitted by [`beam.CreateList`](https://godoc.org/github.com/apache/beam/sdks/go/pkg/beam#CreateList), unlike KeyValue pairs. 
* We can avoid creating new types `SourcesEntry` which are not protos and therefore will have trouble when it comes to serializing them for a map reduce.


core/sequencer/server_test.go
```diff
@@ -102,29 +102,53 @@ func TestHighWatermarks(t *testing.T) {
...
-			t.Errorf("HighWatermarks(): %v", err)
-		}
-		if count != tc.count {
-			t.Errorf("HighWatermarks(): count: %v, want %v", count, tc.count)
-		}
-		if !cmp.Equal(highs, tc.highs) {
-			t.Errorf("HighWatermarks(): %v, want %v", highs, tc.highs)
-		}
+		t.Run(tc.desc, func(t *testing.T) {
+			count, next, err := s.HighWatermarks(ctx, directoryID, &tc.last, tc.batchSize)
```
golangcibot: Using the variable on range scope `tc` in function literal

core/sequencer/server_test.go
```diff
@@ -102,29 +102,53 @@ func TestHighWatermarks(t *testing.T) {
...
+			if err != nil {
+				t.Fatalf("HighWatermarks(): %v", err)
+			}
+			if count != tc.count {
```
golangcibot: Using the variable on range scope `tc` in function literal

core/sequencer/server_test.go
```diff
@@ -102,29 +102,53 @@ func TestHighWatermarks(t *testing.T) {
...
+				t.Errorf("HighWatermarks(): count: %v, want %v", count, tc.count)
```
golangcibot: Using the variable on range scope `tc` in function literal

core/sequencer/server_test.go
jirfag: @gdbelvin maybe disable `scopelint` in `.golangci.yml`? It has such false-positives (https://github.com/kyoh86/scopelint/issues/4)

```yaml
linters:
  enable-all: true
  disable:
    - maligned
    - scopelint
```

gdbelvin: Thanks for the suggestion!

gdbelvin:
`beam.Run` also refuses to serialize `map[int64]*spb.MapMetadata_SourceSlice` because it is not a proto.

core/keyserver/paginator.go
```diff
@@ -100,37 +92,12 @@ func (s SourceMap) Next(rt *rtpb.ReadToken, lastRow *mutator.LogMessage) *rtpb.R
...
-	if i >= len(shardIDs)-1 {
+// NextShard returns the next shardID from the SourceList.
+// Returns false if there are no more shards or shardID is not in SourceList.
+func (s SourceList) NextShard(shardID int64) (int64, bool) {
+	if shardID >= int64(len(s))-1 {
 		// there are no more shards to iterate over.
 		return 0, false
 	}
-	return shardIDs[i+1], true
+	return shardID + 1, true
```
pavelkalinnikov: Just to double check: Should this be an index in the `SourceList` slice, or should this be the ID of the queue (which, I suppose, needs to be extracted from the `MapMetadata_SourceSlice`)?

core/keyserver/revisions.go
```diff
@@ -133,17 +133,18 @@ func (s *Server) ListMutations(ctx context.Context, in *pb.ListMutationsRequest)
...
-	rt, err := SourceMap(meta.Sources).ParseToken(in.PageToken)
+	rt, err := SourceList(meta.Sources).ParseToken(in.PageToken)
 	if err != nil {
 		return nil, status.Errorf(codes.InvalidArgument, "Failed parsing page_token: %v: %v", in.PageToken, err)
 	}
 
 	// Read PageSize + 1 messages from the log to see if there is another page.
 	high := meta.Sources[rt.ShardId].HighestWatermark
-	msgs, err := s.logs.ReadLog(ctx, d.DirectoryID, rt.ShardId, rt.LowWatermark, high, in.PageSize+1)
+	logID := meta.Sources[rt.ShardId].LogId
```
pavelkalinnikov: `ShardId` vs. `LogId` looks confusing to me. Is `ShardId` just meant to be index in the list of `Slices` that `meta` contains? So this ID makes sense only in combination with a single `meta` message. I'd recommend using a less "global" name for it then. Maybe `SliceIndex`? Although I realize that almost any name here can be confusing.

core/keyserver/paginator.go
gdbelvin: This PR has changed the meaning of this.  
It used to be `LogID` and now it is `index of source slice`

gdbelvin: I'll rename this shard_index for clarity.

core/keyserver/revisions.go
gdbelvin: Renamed to `SliceIndex`

core/sequencer/server.go
```diff
@@ -445,20 +436,29 @@ func (s *Server) HighWatermarks(ctx context.Context, directoryID string, starts
 	}
 	// TODO(gbelvin): Get HighWatermarks in parallel.
 	for _, logID := range logIDs {
-		start := starts[logID]
-		if batchSize <= 0 {
-			watermarks[logID] = start
-			continue
-		}
-		count, high, err := s.logs.HighWatermark(ctx, directoryID, logID, start, batchSize)
+		count, high, err := s.logs.HighWatermark(ctx, directoryID, logID, ends[logID], batchSize)
```
pavelkalinnikov: nit: Maybe `low := ends[logID]` above, and reuse it everywhere in this clause?

core/sequencer/server.go
```diff
@@ -445,20 +436,29 @@ func (s *Server) HighWatermarks(ctx context.Context, directoryID string, starts
...
 		if err != nil {
 			return 0, nil, status.Errorf(codes.Internal,
 				"HighWatermark(%v/%v, start: %v, batch: %v): %v",
-				directoryID, logID, start, batchSize, err)
+				directoryID, logID, ends[logID], batchSize, err)
 		}
-		watermarks[logID] = high
+		starts[logID] = ends[logID]
```
pavelkalinnikov: Then you could write this like:
`starts[logID], ends[logID] = low, high`

core/keyserver/paginator.go
```diff
@@ -60,77 +58,46 @@ func (s SourceMap) ParseToken(token string) (*rtpb.ReadToken, error) {
...
-	// Index of current shard.
-	i := sort.Search(len(shardIDs), func(i int) bool { return shardIDs[i] >= shardID })
-	if i == -1 {
-		// shardID isn't in SourceMap.
-		return 0, false
-	}
-	if i >= len(shardIDs)-1 {
+// NextShard returns the next shardID from the SourceList.
+// Returns false if there are no more shards or shardID is not in SourceList.
+func (s SourceList) NextShard(shardID int64) (int64, bool) {
```
pavelkalinnikov: Please rename `shardID` to `sliceIndex` (and update the comment) to be consistent with the recent change.

core/keyserver/paginator.go
```diff
@@ -60,77 +58,46 @@ func (s SourceMap) ParseToken(token string) (*rtpb.ReadToken, error) {
...
+	if shardID >= int64(len(s))-1 {
 		// there are no more shards to iterate over.
 		return 0, false
 	}
-	return shardIDs[i+1], true
+	return shardID + 1, true
```
pavelkalinnikov: Would it be okay to replace the whole body with the following?
`return sliceIndex+1, sliceIndex+1 < len(s)`

Presumably the caller checks the `bool` before using the returned index?

core/keyserver/paginator.go
pavelkalinnikov: Maybe rename the function itself as well?

pavelkalinnikov: Does this function need to be public? I don't see other use-cases. Might be actually better to drop it, and inline the `< len` check into `Next`.

pavelkalinnikov:
There are some other "shard" mentionings left from the previous code. Find&replace them with "slice index" to be consistent please.

core/keyserver/paginator.go
gdbelvin: Yes, I thought about that, but other people on our team feel strongly about returning `0` or `nil` on error conditions, so I left the if statement in place.

core/sequencer/server.go
gdbelvin: oooh, fancy :-)

core/keyserver/paginator.go
gdbelvin: I'll linline it. The function used to be more complex, but now it's simple enough to do this.

core/keyserver/paginator.go
```diff
@@ -60,77 +58,36 @@ func (s SourceMap) ParseToken(token string) (*rtpb.ReadToken, error) {
...
+			SliceIndex:   rt.SliceIndex,
 			LowWatermark: lastRow.ID,
 		}
 	}
 
 	// Advance to the next shard.
-	nextShard, ok := s.NextShard(rt.ShardId)
-	if !ok {
+	if rt.SliceIndex >= int64(len(s))-1 {
+		// there are no more shards to iterate over.
```
pavelkalinnikov: nit: s/shards/slices/
supernit: `There are ...`

core/keyserver/paginator.go
```diff
@@ -60,77 +58,36 @@ func (s SourceMap) ParseToken(token string) (*rtpb.ReadToken, error) {
...
 	}
 }
 
 // Next returns the next read token. Returns an empty struct when the read is finished.
 // lastRow is the (batchSize)th row from the last read, or nil if fewer than
 // batchSize + 1 rows were returned.
-func (s SourceMap) Next(rt *rtpb.ReadToken, lastRow *mutator.LogMessage) *rtpb.ReadToken {
+func (s SourceList) Next(rt *rtpb.ReadToken, lastRow *mutator.LogMessage) *rtpb.ReadToken {
 	if lastRow != nil {
 		// There are more items in this shard.
```
pavelkalinnikov: Is this "shard" shard?

Issue #1122, opened 2018-11-21
MutationFn: Support multiple leaf edits


gdbelvin:
The current KT mutation function only supports an edit to a single map leaf. 
Add support so that a single mutation can touch multiple leaves.

gdbelvin:
The need for this has been addressed by the ability to batch updates, since the sequencer will not split a batch.

PR #1127, opened 2018-11-26
 Simplify the Reduce function interface

gdbelvin:
Replaces the mutation interface with a reduce function type.

It is not generally true that a struct based mutation function
will work. This change clarifies that a bit and reduces code stutter.

This is part of making a multi-leaf mutation possible #1122

### DIFF
--- a/core/sequencer/server_test.go
+++ b/core/sequencer/server_test.go
@@ -191,7 +191,7 @@ func TestDuplicateMutations(t *testing.T) {
 	} {
 		t.Run(tc.desc, func(t *testing.T) {
 			directoryID := "test"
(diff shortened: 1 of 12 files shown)


core/sequencer/server_test.go
```diff
@@ -191,7 +191,7 @@ func TestDuplicateMutations(t *testing.T) {
 	} {
 		t.Run(tc.desc, func(t *testing.T) {
 			directoryID := "test"
-			newLeaves, err := s.applyMutations(directoryID, entry.New(), tc.msgs, tc.leaves)
+			newLeaves, err := s.applyMutations(directoryID, entry.MutateFn, tc.msgs, tc.leaves)
```
golangcibot: Using the variable on range scope `tc` in function literal (from `scopelint`)

Issue #1129, opened 2018-11-27
API: BatchQueueMutations


gdbelvin:
- [x] Add batch level to mutation storage
- [x] Add batch API
- [x] Call new batch API in hammer. 
- [x] Call new batch API in front end.

gdbelvin:
Fixed by #1142

PR #1132, opened 2018-11-27
Refactor ApplyMutations into its own package

gdbelvin:
By putting `ApplyMutations` in its own package we can make it available for use 
 * During Sequencing
 * In the KeyServer as a precondition of allowing a batch of writes.
 * In the monitor
 * In any standalone tools 

This PR is best reviewed commit by commit. 

This PR contains 
 - a `mapper` package which contains individual data manipulation functions.
 - a `runner` package which contains code for wiring the above functions together.  
       Future implementations of `runner` will be increasingly parallelized and distributed. 

This PR is blocking #1129


core/sequencer/server.go
```diff
@@ -29,6 +29,9 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
+	"github.com/google/keytransparency/core/sequencer/mapper"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

core/sequencer/runner/native.go
```diff
@@ -0,0 +1,98 @@
+// Copyright 2018 Google Inc. All Rights Reserved.
```
golangcibot: : import cycle not allowed (from `typecheck`)

core/sequencer/runner/native.go
```diff
@@ -0,0 +1,98 @@
...
+
+// Package runner contains methods for running a pipeline.
+package runner
+
+import (
+	"github.com/golang/glog"
+
+	"github.com/google/keytransparency/core/mutator"
+	"github.com/google/keytransparency/core/sequencer/mapper"
+	"github.com/google/keytransparency/core/sequencer/runner"
```
golangcibot: could not import github.com/google/keytransparency/core/sequencer/runner (import cycle: [github.com/google/keytransparency/core/sequencer/runner]) (from `typecheck`)

core/sequencer/runner/native.go
```diff
@@ -0,0 +1,98 @@
...
+
+	pb "github.com/google/keytransparency/core/api/v1/keytransparency_go_proto"
+	tpb "github.com/google/trillian"
+)
+
+// ApplyMutations takes the set of mutations and applies them to given leafs.
+// Returns a list of map leaves that should be updated.
+func ApplyMutations(mutatorFunc mutator.ReduceMutationFn,
+	msgs []*ktpb.EntryUpdate, leaves []*tpb.MapLeaf) ([]*tpb.MapLeaf, error) {
```
golangcibot: undeclared name: ktpb (from `typecheck`)

core/sequencer/server.go
```diff
@@ -243,7 +246,7 @@ func (s *Server) CreateRevision(ctx context.Context, in *spb.CreateRevisionReque
 	}
 
 	// Apply mutations to values.
-	newLeaves, err := s.applyMutations(directoryID, entry.MutateFn, mutations, leaves)
+	newLeaves, err := runner.ApplyMutations(entry.MutateFn, mutations, leaves)
```
golangcibot: cannot use mutations (variable of type []*github.com/google/keytransparency/core/api/v1/keytransparency_go_proto.EntryUpdate) as []*invalid type value in argument to runner.ApplyMutations (from `typecheck`)

gdbelvin:
PTAL.  @benlaurie might be taking a look at the auditor soon and it might be nice to remove some code duplication between the sequencer and the auditor.

core/sequencer/runner/native.go
```diff
@@ -0,0 +1,97 @@
...
+			func(l *tpb.MapLeaf) { ret = append(ret, l) })
+	}
+	glog.V(2).Infof("ApplyMutations applied %v mutations to %v leaves", len(msgs), len(leaves))
+	return ret, nil
+}
+
+// Joined is the result of a CoGroupByKey on []*MapLeaf and []*IndexUpdate.
+type Joined struct {
+	Index  []byte
+	Leaves []*tpb.MapLeaf
```
pavelkalinnikov: Should it not be possible to have multiple leaves with the same index?

core/sequencer/mapper/fns.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// ReduceFn decides which of multiple updates can be applied in this revision.
+// TODO(gbelvin): Move to mutator interface.
+func ReduceFn(mutatorFn mutator.ReduceMutationFn,
+	index []byte, leaves []*tpb.MapLeaf, msgs []*pb.EntryUpdate, emit func(*tpb.MapLeaf)) {
+	var oldValue *pb.SignedEntry // If no map leaf was found, oldValue will be nil.
+	if len(leaves) > 0 {
+		var err error
+		oldValue, err = entry.FromLeafValue(leaves[0].GetLeafValue())
```
pavelkalinnikov: Should existence of more than 1 leaf for this index be an error, or at least a warning?

core/sequencer/mapper/fns.go
```diff
@@ -0,0 +1,83 @@
...
+		if err != nil {
+			glog.Warningf("entry.FromLeafValue(): %v", err)
+			return
+		}
+	}
+
+	if got := len(msgs); got < 1 {
```
pavelkalinnikov: I'd say just `if len(msgs) == 0`.

core/sequencer/mapper/fns.go
```diff
@@ -0,0 +1,83 @@
...
+		return
+	}
+
+	// TODO(gbelvin): Create an associative function to choose the mutation to apply.
```
pavelkalinnikov: nit: "associative" usually relates to a binary operator. How about?
`Choose the mutation deterministically, regardless of the messages order.`

core/sequencer/mapper/fns.go
gdbelvin: yes, it should be an error

core/sequencer/mapper/fns.go
gdbelvin: Thanks

core/sequencer/runner/native.go
gdbelvin: Depends on which way we look at it. 
From the point of view of the map which we know should only produce one leaf per index, yes. 
From the perspective of a map reduce, the mapping step can emit multiple <K,V> pairs for any given K.

core/sequencer/runner/native.go
```diff
@@ -0,0 +1,97 @@
...
+
+// Join pairs up MapLeaves and IndexUpdates by index.
+func Join(leaves []*tpb.MapLeaf, msgs []*mapper.IndexUpdate) []*Joined {
+	joinMap := make(map[string]*Joined)
+	for _, l := range leaves {
+		row, ok := joinMap[string(l.Index)]
+		if !ok {
+			row = &Joined{}
+		}
+		row.Index = l.Index
```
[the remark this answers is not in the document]
gdbelvin: good idea

core/sequencer/runner/native.go
```diff
@@ -0,0 +1,97 @@
...
+		row.Leaves = append(row.Leaves, l)
+		joinMap[string(l.Index)] = row
+	}
+	for _, m := range msgs {
+		row, ok := joinMap[string(m.Index)]
+		if !ok {
+			row = &Joined{}
```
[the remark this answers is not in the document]
gdbelvin: Thanks

core/sequencer/runner/native.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	ret := make([]*Joined, 0, len(joinMap))
+	for _, r := range joinMap {
+		ret = append(ret, r)
+	}
+	return ret
+}
+
+// MapUpdateFn converts an update into an IndexedUpdate.
+type MapUpdateFn func(msg *pb.EntryUpdate) (*mapper.IndexUpdate, error)
```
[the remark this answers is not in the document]
gdbelvin: Good idea

gdbelvin:
Updated PTAL

core/sequencer/mapper/fns.go
```diff
@@ -0,0 +1,87 @@
...
+	index []byte, leaves []*tpb.MapLeaf, msgs []*pb.EntryUpdate, emit func(*tpb.MapLeaf)) {
+	if got := len(leaves); got != 1 {
+		glog.Errorf("expected 1 map leaf for index %x, got %v", index, got)
+		return
+	}
+	var oldValue *pb.SignedEntry // If no map leaf was found, oldValue will be nil.
+	var err error
+	if len(leaves) > 0 {
+		oldValue, err = entry.FromLeafValue(leaves[0].GetLeafValue())
+		if err != nil {
```
pavelkalinnikov: optional: Merge this line with the previous one?

core/sequencer/mapper/fns.go
```diff
@@ -0,0 +1,87 @@
...
+			glog.Warningf("entry.FromLeafValue(): %v", err)
```
pavelkalinnikov: I wonder if there is a better practice for reporting errors in Beam model, other than silently throwing them all to glog. I vaguely remember some error callbacks mentioned in the user guide.
TODO?

core/sequencer/mapper/fns.go
```diff
@@ -0,0 +1,87 @@
...
+			return
+		}
+	}
+
+	if len(msgs) == 0 {
+		return
```
pavelkalinnikov: Can `msgs` be empty in normal operation, or it's an error/warning? Presumably the pipeline fetches only the indexes that are being updated, i.e. those that have non-empty `msgs`.

core/sequencer/mapper/fns.go
gdbelvin: The above logic is actually incorrect.  0 or 1 leaves are permitted for every index.

core/sequencer/mapper/fns.go
gdbelvin: Errors in beam will cause the entire pipeline to fail. I'll add this.

core/sequencer/mapper/fns.go
gdbelvin: `len(msgs)` means that the join step was provided a map value, but not a log item. 
In general this is permitted, but in our case it would be a surprise, so yes, we should error.

PR #1131, opened 2018-11-27
Implement BatchGetUser

gdbelvin:
Implements `BatchGetUser`.

BatchGetUser is an important milestone on the path to a batch UpdateUser flow.

PR #1133, opened 2018-11-27
BatchQueueUser API

gdbelvin:
Support queueing updates to large sets of users at once. 

Batches of updates will not be split across revisions.


core/api/v1/keytransparency.proto
```diff
@@ -474,8 +484,15 @@ service KeyTransparency {
   // update appears after a timeout.
   rpc QueueEntryUpdate(UpdateEntryRequest) returns (google.protobuf.Empty) {
     option (google.api.http) = {
-      post: "/v1/directories/{directory_id}/users/{user_id}:queue"
+      post: "/v1/directories/{directory_id}/users/{entry_update.user_id}:queue"
       body: "entry_update"
     };
   }
+  // BatchQueueUser enqueues a list of user profiles.
+  rpc BatchQueueUser(BatchQueueUserRequest) returns (google.protobuf.Empty) {
```
jtoohill: The naming seems weird - you're queuing an update for a user, not queuing a user. What about `BatchQueueUserUpdate`, with the custom verb `:batchQueueUpdate`?

That'd fall slightly afoul of the Cloud pattern of having API endpoints named VerbResource, but `BatchQueueUpdateUser` also strikes me as odd.

core/keyserver/keyserver.go
```diff
@@ -89,115 +88,123 @@ func New(tlog tpb.TrillianLogClient,
 // this user and that it is the same one being provided to everyone else.
 // GetUser also supports querying past values by setting the revision field.
 func (s *Server) GetUser(ctx context.Context, in *pb.GetUserRequest) (*pb.GetUserResponse, error) {
```
jtoohill: GetUser changes seem out of place in this PR, did this get mixed up with #1131?

gdbelvin: Yes, it did.  Just rebased to remove it.

core/api/v1/keytransparency.proto
gdbelvin: That sounds good.  
I think the reason `BatchQueueUpdateUser` sounds odd is because `Queue` and `Update` are both verbs :-)

gdbelvin:
Removed spurious `GetUser` commit and renamed `BatchQueueUser` to `BatchQueueUserUpdate`

core/keyserver/keyserver.go
```diff
@@ -306,47 +312,19 @@ func (s *Server) QueueEntryUpdate(ctx context.Context, in *pb.UpdateEntryRequest
...
-	if err != nil {
-		glog.Errorf("GetUser failed: %v", err)
-		return nil, status.Errorf(codes.Internal, "Read failed")
+	userIDs := make([]string, 0, len(in.Updates))
+	for _, u := range in.Updates {
+		if err := validateEntryUpdate(u, vrfPriv); err != nil {
+			glog.Warningf("Invalid UpdateEntryRequest: %v", err)
+			return nil, status.Errorf(codes.InvalidArgument, "Invalid request")
+		}
+		userIDs = append(userIDs, u.UserId)
```
golangcibot: this result of append is never used, except maybe in other appends (from `megacheck`)

PR #1136, opened 2018-11-29
Cleanup Sequencer API

gdbelvin:
Rework Sequencer API
 - Consistency of `Revision` terminology
 - Create `DefineRevisions` API
 - Return metrics in the API responses

### DIFF
--- a/core/sequencer/sequencer_api.proto
+++ b/core/sequencer/sequencer_api.proto
@@ -44,19 +44,6 @@ message MapMetadata {
   repeated SourceSlice sources = 2;
 }
 
-// CreateRevisionRequest contains information needed to create a new revision.
-message CreateRevisionRequest {
-  // directory_id is the directory to apply the mutations to.
-  string directory_id = 1;
-  // deprecated messages. Server now reads messages via source watermarks.
-  reserved 2;
-  // revision is the expected revision of the new revision.
-  int64 revision = 3;
-  // deprecated map_metadata. Server now reads source watermarks from batch
-  // table.
-  reserved 4;
-}
-
 // RunBatchRequest triggers the sequencing of a batch of mutations for a
 // directory, with the batch size governed by the request parameters.
 message RunBatchRequest {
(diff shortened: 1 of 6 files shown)


core/sequencer/server.go
```diff
@@ -129,34 +129,11 @@ func NewServer(
...
-			return nil, err
-		}
-	}
-	return &empty.Empty{}, nil
-}
-
 // DefineRevisions examines the outstanding mutations and returns a list of
 // outstanding revisions that have not been applied.
-func (s *Server) DefineRevisions(ctx context.Context, dirID string, minBatch, maxBatch int32) ([]int64, error) {
+func (s *Server) DefineRevisions(ctx context.Context, in *spb.DefineRevisionsRequest) (*spb.DefineRevisionsResponse, error) {
```
golangcibot: line is 125 characters (from `lll`)

core/sequencer/server.go
```diff
@@ -139,27 +139,34 @@ func NewServer(
...
 			Revision:    rev,
 		}); err != nil {
-			return nil, err
+			glog.Errorf("ApplyRevision(dir: %v, rev: %v): %v", in.DirectoryId, rev, err)
+			break
 		}
 	}
+
+	publishReq := &spb.PublishRevisionsRequest{DirectoryId: in.DirectoryId}
+	_, err = s.loopback.PublishRevisions(ctx, publishReq)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

core/sequencer/sequencer_api.proto
```diff
@@ -71,19 +58,66 @@ message RunBatchRequest {
...
+
+// PublishRevisions
+message PublishRevisionsResponse {
+  // revisions published.
+  repeated int64 revisions = 1;
 }
 
 // The KeyTransparency Sequencer API.
 service KeyTransparencySequencer {
   // RunBatch reads outstanding mutations and calls CreateRevision.
```
jtoohill: "and calls DefineRevisions, ApplyRevision, and PublishRevisions in sequence."

core/sequencer/sequencer_api.proto
```diff
@@ -71,19 +58,66 @@ message RunBatchRequest {
...
-  // CreateRevision applies the contained mutations to the current map root.
+  // DefineRevision examines the outstanding items in the queue and optionally
+  // commits to one or more revisions.
+  rpc DefineRevisions(DefineRevisionsRequest) returns (DefineRevisionsResponse);
+  // ApplyRevision applies the contained mutations to the current map root.
   // If this method fails, it must be retried with the same arguments.
-  rpc CreateRevision(CreateRevisionRequest) returns (google.protobuf.Empty);
-  // PublishBatch copies the MapRoots of all known map revisions into the Log
+  rpc ApplyRevision(ApplyRevisionRequest) returns (ApplyRevisionResponse);
+  // PublishRevision copies the MapRoots of all known map revisions into the Log
```
jtoohill: PublishRevisions

core/sequencer/server.go
```diff
@@ -180,16 +191,16 @@ func (s *Server) DefineRevisions(ctx context.Context, dirID string, minBatch, ma
 
 	// Don't create new revisions if there are ones waiting to be applied.
 	if len(outstanding) > 0 {
-		return outstanding, nil
+		return &spb.DefineRevisionsResponse{OutstandingRevisions: outstanding}, nil
 	}
 
-	// Query metadata about outstanding mutations.
+	// Query metadatab about outstanding log items.
```
jtoohill: metadata

core/sequencer/sequencer_api.proto
```diff
@@ -71,19 +58,66 @@ message RunBatchRequest {
...
+  repeated int64 revisions = 1;
 }
 
 // The KeyTransparency Sequencer API.
 service KeyTransparencySequencer {
   // RunBatch reads outstanding mutations and calls CreateRevision.
   rpc RunBatch(RunBatchRequest) returns (google.protobuf.Empty);
-  // CreateRevision applies the contained mutations to the current map root.
+  // DefineRevision examines the outstanding items in the queue and optionally
+  // commits to one or more revisions.
```
jtoohill: > and optionally commits to one or more revisions

I'm not sure what committing to a revision means, can you clarify?

core/sequencer/server.go
```diff
@@ -139,27 +139,38 @@ func NewServer(
...
 
-	for _, rev := range outstandingRevs {
-		if _, err := s.loopback.CreateRevision(ctx, &spb.CreateRevisionRequest{
+	for _, rev := range defResp.OutstandingRevisions {
+		if _, err := s.loopback.ApplyRevision(ctx, &spb.ApplyRevisionRequest{
 			DirectoryId: in.DirectoryId,
 			Revision:    rev,
 		}); err != nil {
-			return nil, err
+			glog.Errorf("ApplyRevision(dir: %v, rev: %v): %v", in.DirectoryId, rev, err)
```
jtoohill: I notice you log and continue to publish the applied revisions, rather than retrying failed ApplyRevision calls. That seems correct to me, but it's somewhat at odds with the comment in the API doc:

> If this method fails, it must be retried with the same arguments.

Should we remove that line from the API?

core/sequencer/sequencer_api.proto
gdbelvin: Update requests go through the following process:

1. `QueueEntryUpdate` writes `EntryUpdate` to a *log of mutations* 
2. `DefineBatch` reads the highest timestamp in the log of mutations and decides that all the log items between `LowestWatermark` and `HighestWatermark` are going to be part of map revision `N`.  We write this to disk first so that we can retry it in case anything goes wrong. 
3. `ApplyRevision` reads all the log messages between low and high watermarks, decides which messages are valid, and writes all the information to the Trillian map at revision `N`. 
4. `PublishRevisions` takes all the signed map roots, including revision `N` and writes them to the Trillian log of signed map roots.  At this point, clients will be able to see map revision `N` and use the values inside of it. 

Each of these steps can be retried on its own if anything goes wrong.

core/sequencer/server.go
gdbelvin: The API line is correct. 

`RunBatch` gets called on a periodic timer.  The next call to `DefineRevisions` will contain the failed revision since it has not yet been applied, and we will try it again.

PR #1144, opened 2018-12-05
Change (low,high] to [low,high) intervals in HighWatarmarks

gdbelvin:
This change allows the us to use input logs with an item at `0`. 
It also greatly simplifies much of the iteration logic. 

Fixes #1144


core/keyserver/revisions_test.go
```diff
@@ -76,11 +76,10 @@ func (m *mutations) Send(ctx context.Context, dirID string, mutation ...*pb.Entr
 func (m *mutations) ReadLog(ctx context.Context, dirID string,
 	logID, low, high int64, batchSize int32) ([]*mutator.LogMessage, error) {
 	logShard := (*m)[logID]
-	low = low + 1 // Begin exclusive
 	if low > int64(len(logShard)) {
 		return nil, fmt.Errorf("invalid argument: low: %v, want < max watermark: %v", low, len(*m))
```
pavelkalinnikov: want <= max watermark?

pavelkalinnikov: len(*m) -> len(logShard)?

core/sequencer/server.go
```diff
@@ -244,8 +244,8 @@ func (s *Server) readMessages(ctx context.Context, directoryID string, meta *spb
 	batchSize int32) ([]*mutator.LogMessage, error) {
 	msgs := make([]*mutator.LogMessage, 0)
 	for _, source := range meta.Sources {
-		low := source.GetLowestWatermark()
-		high := source.GetHighestWatermark()
+		low := source.GetLowestInclusive()
```
pavelkalinnikov: Can `source` be nil? Maybe just `source.HighestWatermark`, like in other similar lines below.

impl/sql/mutationstorage/queue.go
```diff
@@ -162,24 +162,24 @@ func (m *Mutations) HighWatermark(ctx context.Context, directoryID string, logID
 		`SELECT COUNT(*), COALESCE(MAX(T1.Time), ?) FROM
```
pavelkalinnikov: How about:

- `SELECT COUNT(*), COALESCE(MAX(T1.Time)+1, ?)`
- `start-1` -> `start` in line 169
- `return count, high, nil` in line 173

This way it will return `high=low` if the batch is empty, whereas the current implementation will always increase `high` for no reason.

PR #1148, opened 2018-12-13
Submit SequencedLeaves with Batch API 

gdbelvin:
AddSequencedLeaf takes 1s to return. Rather than calling it
sequentially, we can submit all the leaves for publishing at once,
reducing the latency of PublishRevisions by multiple seconds.

Depends on google/trillian#1375


core/sequencer/trillian_client.go
```diff
@@ -91,7 +92,23 @@ func (t *Trillian) LogClient(ctx context.Context, dirID string) (trillianLog, er
 
 	// Create verifying log client.
 	trustedRoot := types.LogRootV1{} // TODO(gbelvin): Store and track trustedRoot.
-	return tclient.NewFromTree(t.tlog, logTree, trustedRoot)
+	client, err := tclient.NewFromTree(t.tlog, logTree, trustedRoot)
+	if err != nil {
+		return nil, err
+	}
+	return &LogClient{LogClient: client}, nil
```
golangcibot: cannot use &(LogClient literal) (value of type *LogClient) as trillianLog value in return statement: missing method AddSequencedLeaves (from `typecheck`)

PR #1149, opened 2018-12-21
Nonblocking PublishRevisions path

gdbelvin:
The last step of creating a new map revision involves appending the map root to the log of map revisions. After sending the map root to the log, however, it is unnecessary to wait for the log to sequence the entry and produce an STH. Typically producing an STH takes about 2 seconds and greatly exceeds the time taken by the other steps in `RunBatch. 

Tests which require strict sequential operations of `RunBatch` can keep the existing blocking behavior by setting a new `Block` argument.


core/sequencer/sequencer_api.proto
```diff
@@ -56,6 +56,8 @@ message RunBatchRequest {
   int32 min_batch = 2;
   // max_batch is the maximum number of items in a batch.
   int32 max_batch = 3;
+  // block until a Signed Log Root has been published which encompases all map roots.
```
pavelkalinnikov: Is it all map roots, or just one map root (that contains all mutations of the batch)?

gdbelvin:
It is all map roots. Hence the name Publish map revisions plural

PR #1157, opened 2019-01-08
Trim Directory info

gdbelvin:
Don't return unnecessary information in the KT directory create / list / delete methods. 
This helps clarify which fields are important and saves space.


gdbelvin:
I was unable to find any field mask support in Go. Not sure if I missed it

PR #1164, opened 2019-01-18
Fix CI

gdbelvin:
This PR is best reviewed commit by commit. 

There were two essential issues that had the Continuous Integration environment wedged. 
- The kubernetes configs for Trillian had stale paths for the binaries. 
-  The key transparency sequencer was opening the same port twice.  

Fixing the later involved refactoring the server setup to reuse a common `net.Listener`.  

Contrary to best practice, this PR also includes a few fixups and optimizations: 
- `strings.HasPrefix` vs. `strings.Contains` should be faster for matching non-GRPC connections.
- Split out a few functions to be purpose specific. 
- Use `DialContext`

### DIFF
--- a/cmd/keytransparency-sequencer/main.go
+++ b/cmd/keytransparency-sequencer/main.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"net"
 	"os"
-	"strings"
 	"time"
 
 	"github.com/golang/glog"
@@ -193,11 +192,6 @@ func main() {
 	httpServer := startHTTPServer(grpcServer, addr,
 		pb.RegisterKeyTransparencyAdminHandlerFromEndpoint,
 	)
-
-	cli, err := etcd.NewClient(strings.Split(*etcdServers, ","), 5*time.Second)
-	if err != nil || cli == nil {
-		glog.Exitf("Failed to create etcd client: %v", err)
-	}
 
 	// Periodically run batch.
 	electionFactory, closeFactory := getElectionFactory()
--- a/cmd/keytransparency-sequencer/main.go
+++ b/cmd/keytransparency-sequencer/main.go
@@ -114,13 +114,15 @@ func getElectionFactory() (election2.Factory, func()) {
 func main() {
 	flag.Parse()
 	ctx := context.Background()
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
 
 	// Connect to trillian log and map backends.
-	mconn, err := grpc.Dial(*mapURL, grpc.WithInsecure())
+	mconn, err := grpc.DialContext(ctx, *mapURL, grpc.WithInsecure())
 	if err != nil {
 		glog.Exitf("grpc.Dial(%v): %v", *mapURL, err)
 	}
-	lconn, err := grpc.Dial(*logURL, grpc.WithInsecure())
+	lconn, err := grpc.DialContext(ctx, *logURL, grpc.WithInsecure())
 	if err != nil {
 		glog.Exitf("Failed to connect to %v: %v", *logURL, err)
 	}
--- a/cmd/keytransparency-sequencer/main.go
+++ b/cmd/keytransparency-sequencer/main.go
@@ -151,9 +151,15 @@ func main() {
 	if err != nil {
 		glog.Exitf("error creating TCP listener: %v", err)
 	}
-	addr := lis.Addr().String()
+	glog.Infof("Listening on %v", lis.Addr().String())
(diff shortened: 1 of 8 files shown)


cmd/keytransparency-sequencer/server.go
```diff
@@ -25,12 +25,13 @@ import (
 	"google.golang.org/grpc/credentials"
 )
 
-func startHTTPServer(grpcServer *grpc.Server, addr string,
-	services ...serverutil.RegisterServiceFromEndpoint) *http.Server {
-	// Wire up gRPC and HTTP servers.
-	tcreds, err := credentials.NewClientTLSFromFile(*certFile, "")
-	if err != nil {
-		glog.Exitf("Failed opening cert file %v: %v", *certFile, err)
+func serveHTTPMetric(port string) {
```
pavelkalinnikov: Is it `addr` rather than `port`?

cmd/keytransparency-sequencer/server.go
```diff
@@ -25,12 +25,13 @@ import (
...
+	metricMux := http.NewServeMux()
+	metricMux.Handle("/metrics", promhttp.Handler())
+
+	glog.Infof("Hosting metrics on %v", port)
+	if err := http.ListenAndServe(port, metricMux); err != nil {
+		glog.Fatalf("ListenAndServeTLS(%v): %v", *metricsAddr, err)
```
pavelkalinnikov: `*metricsAddr` -> `addr`?

cmd/keytransparency-sequencer/server.go
pavelkalinnikov: `Metric` -> `Metrics`?

cmd/keytransparency-sequencer/server.go
```diff
@@ -25,12 +25,13 @@ import (
...
 	}
 	gwmux, err := serverutil.GrpcGatewayMux(addr, tcreds, services...)
```
pavelkalinnikov: When is this meant to run? `ListenAndServe` above returns only when the server is stopped.

cmd/keytransparency-sequencer/main.go
```diff
@@ -140,12 +140,7 @@ func main() {
 		glog.Exitf("Failed to create directory storage object: %v", err)
 	}
 
-	creds, err := credentials.NewServerTLSFromFile(*certFile, *keyFile)
```
pavelkalinnikov: Should `certFile` and `keyFile` be marked as deprecated?

cmd/keytransparency-sequencer/main.go
```diff
@@ -202,15 +209,9 @@ func main() {
...
-	defer cancel()
-	sequencer.PeriodicallyRun(cctx, time.Tick(*refresh), func(ctx context.Context) {
+	sequencer.PeriodicallyRun(ctx, time.Tick(*refresh), func(ctx context.Context) {
 		if err := signer.RunBatchForAllDirectories(ctx); err != nil {
 			glog.Errorf("PeriodicallyRun(RunBatchForAllDirectories): %v", err)
 		}
 	})
-
-	// Shutdown.
-	httpServer.Shutdown(cctx)
```
pavelkalinnikov: How is the server shut down now?

cmd/keytransparency-sequencer/server.go
gdbelvin: done

cmd/keytransparency-sequencer/server.go
gdbelvin: done

cmd/keytransparency-sequencer/server.go
gdbelvin: It's meant to run in it's own go routine like so:
```go
go serveHTTPMetric(addr)
```

cmd/keytransparency-sequencer/main.go
gdbelvin: We still use `certFile` and `keyFile` - they just get used as part of `http.ServeTLS` rather than the go server itself.

cmd/keytransparency-sequencer/main.go
gdbelvin: Reading the [documentation](https://golang.org/pkg/net/http/#Server.Shutdown) closely, `Shutdown` looks like it's intended to be called from a separate thread than the serving thread.  
If there was a way to trigger a shutdown, we would call it there, but at the moment this open source main function has SIGINT as the only shutdown signal.

gdbelvin:
Updated PTAL

Issue #1152, opened 2019-01-07
Store Trillian Tree object in the Directory table


gdbelvin:
Each operation currently touches the Trillian Admin API in order to retrieve the public keys and other static configuration information for the the Trillian Tree. This is both less secure, and induces many unnecessary RPCs.

By storing relevant parts of the the Trillian.Tree object in the directory table we can remove these RPCs and possibly also remove quite a few references to the Trillian admin API from the code.

Zyqsempai:
Sounds like a fun) I can take this one.

Zyqsempai:
It would be great if you could list the parts of Tree you want to move to Directory table.

gdbelvin:
I've just created #1165 which trims the Tree info down to the essential pieces that Key Transparency needs for verification. 

It might be tempting to split out all fields separately for storage in the database, but it may be easier and more future proof to just store the serialized proto of the `trillian.Tree`

Zyqsempai:
Sounds quite reasonable, also it will make `pb.Directory{}` more consistent with `directory.Directory{}` struct.

gdbelvin:
Indeed it would. That would be very nice.

PR #1167, opened 2019-01-23
Store trillian map,log trees in DB

Zyqsempai:
* Romeved from directory table:
MapId, LogId fields
* Added to directory table:
Map, Log fields with type BLOB
* Removed fecth request to TrillianTree

Fixes: https://github.com/google/keytransparency/issues/1152

### DIFF
--- a/impl/sql/directory/storage_test.go
+++ b/impl/sql/directory/storage_test.go
@@ -23,7 +23,7 @@ import (
 	"github.com/golang/protobuf/proto"
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/keytransparency/core/directory"
-	"github.com/google/trillian"
+	tpb "github.com/google/trillian"
 	"github.com/google/trillian/crypto/keyspb"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
@@ -58,10 +58,10 @@ func TestList(t *testing.T) {
 			directories: []*directory.Directory{
 				{
 					DirectoryID: "directory1",
-					Map: &trillian.Tree{
+					Map: &tpb.Tree{
 						TreeId: 1,
 					},
-					Log: &trillian.Tree{
+					Log: &tpb.Tree{
 						TreeId: 2,
 					},
 					VRF:         &keyspb.PublicKey{Der: []byte("pubkeybytes")},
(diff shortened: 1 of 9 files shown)


Zyqsempai:
@gdbelvin Hi, can you help a little bit with failing tests, i am confused with tests reports i get.
Also, how can I run test ENV for e-to-e tests?

impl/sql/directory/storage_test.go
```diff
@@ -23,11 +23,11 @@ import (
 	"github.com/golang/protobuf/proto"
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/keytransparency/core/directory"
+	"github.com/google/trillian"
```
gdbelvin: nit: because the `trillian` package is a proto, we prefer to import rename it to something ending in "pb" 

```suggestion
	tpb "github.com/google/trillian"
```

gdbelvin:
Your changes so far look great. 
To run end to end tests: `go test ./impl/integration`.  This creates a test `Env` and passes it to the integration tests in `core/integration`. 

The errors CI is currently generating suggest some kind of data corruption. Perhaps the log and map trees are getting swapped? 
```
tlog.GetLatestSignedLogRoot(9171319461164051086): rpc error: code = NotFound desc = tree 9171319461164051086 not found

NewLogVerifierFromTree(): TreeType: MAP, want LOG or PREORDERED_LOG
```

Zyqsempai:
@gdbelvin Not sure if that only my problem, but `docker-compose up` doesn't work, so i cannot build up all infra for tests.

gdbelvin:
Please see #1169 for the `docker-compose` fix. 
Please ensure that the continuous integration tests and the following commands pass:
```sh
go test ./...
golangci-lint run
```

Zyqsempai:
Unfortunately still have problems with `docker-compose` even after your fix.
```Building log-server
Step 1/8 : FROM golang:1.10 as build
ERROR: Service 'log-server' failed to build: Get https://registry-1.docker.io/v2/: Service Unavailable```
Seems like i doesn't have access to the registry.

Zyqsempai:
So, the tests finaly fixed, but i found important issue.
The problem is that we cannot update diretories in DB, but we need it as far as we now store trees there, and for example on Delete action we should somehow update `Deleted` status in those trees, but we can't.
So looks like we should extend storage interface with Update method, or make Write method upsert type.
I can do it in separate PR.

gdbelvin:
As for updates, I think its debatable.   The `deleted` column would be the only relevant field that could change, but I think a good argument could be made to remove that field from the filter list.  At the KT layer, we should care about deleted / undeleted directories, not trees.

Zyqsempai:
@gdbelvin Agree, if directory marked as delete I assume we can consider trees deleted as well.
I can remove those fields in separate PR.

PR #1172, opened 2019-01-25
Remove deleted status from map and log trees

Zyqsempai:
Removed unused field `deleted` from Trillian tree struct we store in our DB.
Removed part of the tests related to this field.


Zyqsempai:
@gdbelvin Not sure if this PR is what you meant;)

gdbelvin:
Yep. That looks about right. Best to be returning misleading fields :-)

Zyqsempai:
Hmm, again strange test report.

gdbelvin:
The build is currently broken. #1173 Fixes it

PR #1173, opened 2019-01-25
Create an encrypted tink reader

gdbelvin:
Tink has a new `tink.KeysetReader` interface in google/tink#4e4d11e 

This PR refactors the keyset reader and writer implementations into their own package in a format that implements the new interface.

### DIFF
--- a/core/crypto/tinkreader/encwritier.go
+++ b/core/crypto/tinkreader/encwritier.go
@@ -1,0 +1,67 @@
+package tinkreader
+
+import (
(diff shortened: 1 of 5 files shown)


core/crypto/tinkreader/encwritier.go
```diff
@@ -0,0 +1,67 @@
+package tinkreader
```
jtoohill: Should the package just be called "tink" since it also defines a writer?

PR #1180, opened 2019-02-04
AddDirectoryForSequencing

gdbelvin:
Support controlling the directories a binary sequences with a flag, rather than the entire set by default.

### DIFF
--- a/cmd/keytransparency-sequencer/main.go
+++ b/cmd/keytransparency-sequencer/main.go
@@ -213,7 +213,10 @@ func runSequencer(ctx context.Context, conn, mconn *grpc.ClientConn,
 	)
 
 	sequencer.PeriodicallyRun(ctx, time.Tick(*refresh), func(ctx context.Context) {
-		if err := signer.RunBatchForAllDirectories(ctx); err != nil {
+		if err := signer.AddAllDirectoriesForSequencing(ctx); err != nil {
+			glog.Errorf("PeriodicallyRun(AddAllDirectoriesForSequencing): %v", err)
+		}
+		if err := signer.RunBatchForAllMasterships(ctx); err != nil {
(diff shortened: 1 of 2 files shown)


cmd/keytransparency-sequencer/main.go
```diff
@@ -213,7 +213,10 @@ func runSequencer(ctx context.Context, conn, mconn *grpc.ClientConn,
 	)
 
 	sequencer.PeriodicallyRun(ctx, time.Tick(*refresh), func(ctx context.Context) {
-		if err := signer.RunBatchForAllDirectories(ctx); err != nil {
+		if err := signer.AddAllDirectoriesForSequencing(ctx); err != nil {
+			glog.Errorf("PeriodicallyRun(AddAllDirectoriesForSequencing): %v", err)
+		}
+		if err := signer.RunBatchForAllMasterships(ctx); err != nil {
 			glog.Errorf("PeriodicallyRun(RunBatchForAllDirectories): %v", err)
```
pavelkalinnikov: message needs updating

core/sequencer/sequencer.go
```diff
@@ -77,18 +77,29 @@ func (s *Sequencer) TrackMasterships(ctx context.Context) {
...
 	for _, d := range directories {
-		knownDirectories.Set(1, d.DirectoryID)
-		s.tracker.AddResource(d.DirectoryID)
+		s.AddDirectoryForSequencing(d.DirectoryID)
 	}
+	return nil
+}
+
+// AddDirectoryForSequencing adds dirIDs to the set of resources this sequencer attempts to obtain mastership for.
+func (s *Sequencer) AddDirectoryForSequencing(dirIDs ...string) {
```
pavelkalinnikov: Since the type is called `Sequencer`, maybe `ForSequencing` suffix is a bit redundant. How about renaming this to `AddDirectoryIDs`, and the above method to `AddAllDirectories`?

Issue #1182, opened 2019-02-04
Mysql docker container not reachable from outside.


Zyqsempai:
Current mysql docker container port, 3306 is not exposed outside.
At least it cause errors on `go generate ./core/testdata` step.

PR #1183, opened 2019-02-04
Expose mysql docker container port

Zyqsempai:
Exposed my-sql DB port 3306 on same port outside.
At least it fails on `go generate ./core/testdata` step.

Fixes: https://github.com/google/keytransparency/issues/1182


Zyqsempai:
Yep, coz for me it's the most appropriate and easier way to build up environment.

Issue #1158, opened 2019-01-08
[README] "Running the server"


DazWilkin:
The current instructions are misordered and duplicative:

```
go get -u github.com/google/keytransparency/...
go get -u github.com/google/trillian/...
./scripts/prepare_server.sh -f
cd $GOPATH/src/github.com/google/keytransparency
docker-compose up -d
docker-compose logs --tail=0 --follow
```
May be better:
```
go get -u github.com/google/keytransparency/...
go get -u github.com/google/trillian/...

cd $GOPATH/src/github.com/google/keytransparency
./scripts/prepare_server.sh -f

docker-compose up -d
docker-compose logs --tail=0 --follow
```
Also, rather than run detached and then tail the logs, it is simpler to run 'attached':
```
docker-compose up
```

gdbelvin:
This is a really great suggestion. Would you be willing to put together a PR?

DazWilkin:
Sure. I'm now on vacation but will submit something when I return.

PR #1187, opened 2019-02-05
Proposed changes to README

DazWilkin:
Some proposals to address #1158.

1. Consistent usage of `--[flag]`
1. Be more explicit about the key set password: `${PASSWORD}` rather than `password`
1. Added clarity to the use of an implicit URL for the Key Transparency server (`35.202.56.9:443`)
1. L119 and L120 incorrectly refer to the published server but ought -- I think? -- reference the locally running service. Revised the URLs.
1. Duplicated these as LL84-86 to use the public Key Transparency server
1. Propose (haven't done) adding commands to create and test a specific email address rather than `user@domain.com` and `foo@bar.com`

Unsures:
1. L40 reads `Generate a private key` and this complements L50 `Publish the public key` but these aren't (I think and not in the example given) a public-private key pair?
1. The example posts data with value `test` (base64 encoded: `dGVzdA==`) not a public key. Would it be more clear to state that arbitrary data may be posted? [Or am I missing something (else)?]
1. Add predefined (perhaps `foo@bar.com`) email address to the public server so that users can lookup a known email? Or, would this need to be gdbelvin@google.com? Is that a problem?

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,15 +40,22 @@ development.
 #### Generate a private key
 
   ```sh
-  keytransparency-client authorized-keys create-keyset -p password
-  keytransparency-client authorized-keys list-keyset -p password
+  PASSWORD=[[YOUR-KEYSET-PASSWORD]]
+  keytransparency-client authorized-keys create-keyset --password=${PASSWORD}
+  keytransparency-client authorized-keys list-keyset --password=${PASSWORD}
   ```
+NB A default for the Key Transparency server URL is being used here. The default value is "35.202.56.9:443". The flag `--kt-url` may be used to specify the URL of Key Transparency server explicitly.
+
 
 #### Publish the public key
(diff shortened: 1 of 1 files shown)


gdbelvin:
L40:  Now that I look at it, this is confusing. Key Transparency has two sets of private keys: one for authorizing updates (the `authorized-keys`), and a second one which is the "data" being committed to.  In the example, we have the user generating real authorized keys, but then submitting `test` in base64 as their public key.  We should probably change this to be a more representative scenario. 

Adding some predefined keys to the server sounds like a great idea.  What keys to use exactly would depend on what format we choose for the representative use case.  We could (for example) mirror all the public keys in the PGP key directory if we wanted...

DazWilkin:
Very helpful explanation, thank you.

The PGP keys would be a good|useful example.

More simply, you could create an entry for `foo@bar.com` with some dummy data (`Dummy Public Key #1`, `Dummy Public Key #2` ....) similar to what's shown now.

PR #1191, opened 2019-02-07
Remove secrets from Docker Images

gdbelvin:
Private keys should not be part of the docker images. 
This was an old hack that predated docker's `secrets` feature.  

TODO / Help wanted: add a kubernetes configmap that does the same thing.


gdbelvin:
@DazWilkin I also verified that in this configuration, `docker-compose up` works successfully.  Can you verify with respect to #1190 ?

DazWilkin:
Checking it now.

DazWilkin:
`sequencer` service needs the secret references too.

DazWilkin:
Otherwise, it appears to work. I continue to have issues with `context deadline exceeded`

gdbelvin:
Just added the keys to the sequencer.  See if that doesn't solve the context deadline exceeded issue...
If the sequencer isn't starting because it doesn't have access to those keys, it would prevent sequencing progress from happening which would cause a context exceeded error

DazWilkin:
I'd added they keys myself to get it to work but it doesn't address my issue :-(

gdbelvin:
The other place to look for these context deadline errors is the trillian log sequencer. If either the KT sequencer or the log sequencer are having issues clients will see deadline exceeded.

DazWilkin:
checking

DazWilkin:
Everything else appears (!?) to be OK.

The only other recurring log entry that I see across the services is in `server` and it's:
`interceptor.go:38] auth interceptor: no hander for /google.keytransparency.v1.KeyTransparency/GetRevision`

NB Typo in "hander" sb "handler"

But, otherwise, everything appears OK. Should I publish my logs?

DazWilkin:
I have been unclear as to whether I should use `0.0.0.0:8080` (sequencer) or `0.0.0.0:443` (server).

Trying again using `:443` for both `authorized-keys create-keyset` and then...

```
keytransparency-client post dazwilkin@google.com \
--client-secret=./client_secret.190205.json \
--insecure \
--data='dGVzdA==' \
--password=${PASSWORD} \
--kt-url=0.0.0.0:443 \
--verbose \
--logtostderr
Go to the following link in your browser then type the authorization code: 
https://accounts.google.com/o/oauth2/auth?access_type=online&client_id=[[REDACTED]]
[[REDACTED]]
2019/02/07 10:59:56 ✓ Signed Map Head signature verified.
2019/02/07 10:59:56 ✓ Log inclusion proof verified.
I0207 10:59:56.987019  158923 client.go:145] Trusted root updated to TreeSize 1
2019/02/07 10:59:56 ✓ Log root updated.
2019/02/07 10:59:56 ✓ Commitment verified.
2019/02/07 10:59:57 ✓ VRF verified.
2019/02/07 10:59:57 ✓ map inclusion proof verified.
2019/02/07 10:59:57 Got current entry...
2019/02/07 10:59:57 Sending Update request...
2019/02/07 10:59:58 ✓ Signed Map Head signature verified.
2019/02/07 10:59:58 ✓ Log inclusion proof verified.
Error: update failed: context deadline exceeded
```

gdbelvin:
That timeout error is happening *really* quickly.  I think the client is picking up an odd default timeout value of `0s` for some reason.

gdbelvin:
The other thing that is happening is that we're starting the timeout counter *before* the user goes through the OAuth flow

DazWilkin:
So, you're telling me that my 'open link', allow, copy-paste then enter is too slow? ;-)

gdbelvin:
It's embarrassing :-) Fix: #1195

RJPercival:
This isn't an issue introduced by this PR, but shouldn't the key generation scripts create and chmod the files *before* writing the keys to them?

scripts/gen_monitor_keys.sh
```diff
@@ -15,8 +15,8 @@
 # limitations under the License.
 
 # Create output directory.
-mkdir -p "${GOPATH}/src/github.com/google/keytransparency/genfiles"
-cd "${GOPATH}/src/github.com/google/keytransparency/genfiles"
+mkdir -p "$(go env GOPATH)/src/github.com/google/keytransparency/genfiles"
+cd "$(go env GOPATH)/src/github.com/google/keytransparency/genfiles"
```
RJPercival: super-nit: Won't work as expected if $GOPATH contains more than one path

scripts/gen_server_keys.sh
```diff
@@ -23,7 +23,7 @@ while getopts d:a:s: option; do
 	d) COMMONNAME=${OPTARG};;
 	a) ADDRESS=${OPTARG};;
 	s) SAN_DNS=${OPTARG};;
-	*) echo "usage: ./generate.sh -d <domain> -a <ip_address> -s <san_extension_DNS>"; exit 1;;
+	*) echo "usage: ./gen_server_keys.sh -d <domain> -a <ip_address> -s <san_extension_DNS>"; exit 1;;
```
RJPercival: nit: Can use `$0` or `$(basename $0)` instead of the filename here.

PR #1194, opened 2019-02-07
Merge cmd/gen-test-vectors and integration/client_tests.go

Zyqsempai:
Merged `cmd/gen-test-vectors` and `integration/client_tests.go`
Changed generate command `go test ../../impl/integration --generate`
Added  bool `gnerate` flag to `impl/integration/main_test.go`

I think we need also change the test command in TravisCI.


Zyqsempai:
@gdbelvin Can't say that this is the best implementation. Coz all the tests are pretty messed up and also i wasn't able to test it because of docker mysql issue.

impl/integration/main_test.go
```diff
@@ -16,22 +16,32 @@ package integration
...
 
 	for _, test := range integration.AllTests {
 		t.Run(test.Name, func(t *testing.T) {
 			ctx, cancel := context.WithCancel(context.Background())
 			defer cancel()
-			env, err := NewEnv(ctx)
+			env, err := integration.NewEnv(ctx)
+
+			if *generate {
+				integration.GenerateTestVectors(ctx, env)
```
golangcibot: Error return value of `integration.GenerateTestVectors` is not checked (from `errcheck`)

core/integration/client_tests.go
```diff
@@ -492,3 +501,172 @@ func (m uint64Slice) Less(i, j int) bool { return m[i] < m[j] }
 func cp(tag int) []byte {
 	return []byte(fmt.Sprintf("bar%v", tag))
 }
+
+// GenerateTestVectors verifies set/get semantics.
+// TODO(gbelvin): Migrate into core/integration/client_tests to avoid code rot.
```
gdbelvin: Thanks for moving this!  You can now delete this TODO

core/integration/env.go
```diff
@@ -31,7 +31,6 @@ import (
```
gdbelvin: The separation of `impl/integration` from `core/integration` isn't documented, so let me explain here:
The layout of the code is designed to make using Key Transparency in an enterprise environment easy. Everything that might be implementation specific is in the `impl` directory.  In order for the `core` directory to operate independently of implementation details, it must not have any references to `impl`.  This is why the unit tests that rely only on the interfaces can remain in `core`, while the actual test that supplies the dependencies lives in `impl`. 

Please keep `impl/integration/env.go` in `impl`.

core/integration/alltests.go
```diff
@@ -29,7 +29,7 @@ import (
 )
 
 // Env holds a complete testing environment for end-to-end tests.
-type Env struct {
+type EnvClient struct {
```
gdbelvin: If we keep `impl/integration` I think this change can be reverted.

core/integration/env.go
Zyqsempai: I moved it there because of the cycle dependecies, so it cannot be solved without moving something, or env.go or all the tests to impl/integration.

impl/integration/main_test.go
```diff
@@ -16,25 +16,37 @@ package integration
...
 	for _, test := range integration.AllTests {
 		t.Run(test.Name, func(t *testing.T) {
 			ctx, cancel := context.WithCancel(context.Background())
 			defer cancel()
-			env, err := NewEnv(ctx)
+			env, err := integration.NewEnv(ctx)
 			if err != nil {
 				t.Fatalf("Could not create Env: %v", err)
 			}
+			if *generate {
```
gdbelvin: In this configuration, `GenerateTestVectors` is it's own unit test and as such probably belongs outside the for loop, else we will generate the test vectors `len(AllTests)` times.

core/integration/client_tests.go
```diff
@@ -17,16 +17,25 @@ package integration
...
+	"github.com/google/keytransparency/core/testdata"
+	"github.com/google/keytransparency/impl/authentication"
+	"github.com/google/trillian/types"
+
 	"github.com/google/keytransparency/core/client"
 	"github.com/google/keytransparency/core/sequencer"
 	"github.com/google/keytransparency/core/testutil"
 
+	"github.com/golang/protobuf/jsonpb"
+	"github.com/google/martian/log"
```
gdbelvin: Should this import be `"log"` or `"glog"`?

impl/integration/main_test.go
Zyqsempai: Yeah, but anyway we run `NewEnv(ctx)` for each loop cycle, maybe we should move it outside too?

core/integration/env.go
gdbelvin: What was the cycle in particular?

core/integration/client_tests.go
Zyqsempai: I suggest glog;)

core/integration/client_tests.go
```diff
@@ -492,3 +501,172 @@ func (m uint64Slice) Less(i, j int) bool { return m[i] < m[j] }
 func cp(tag int) []byte {
 	return []byte(fmt.Sprintf("bar%v", tag))
 }
+
+// GenerateTestVectors verifies set/get semantics.
+// TODO(gbelvin): Migrate into core/integration/client_tests to avoid code rot.
+func GenerateTestVectors(ctx context.Context, env *Env) error {
```
gdbelvin: This function is largely a duplicate of `TestEmptyGetAndUpdate`. If we could merge the two that would be ideal - or we can save that for a follow on PR.

core/integration/client_tests.go
```diff
@@ -492,3 +501,172 @@ func (m uint64Slice) Less(i, j int) bool { return m[i] < m[j] }
...
+		ctx            context.Context
+		userID         string
+		signers        []tink.Signer
+		authorizedKeys *tinkpb.Keyset
+	}{
+		{
+			desc:           "empty_alice",
+			wantProfile:    nil,
+			setProfile:     []byte("alice-key1"),
+			ctx:            authentication.WithOutgoingFakeAuth(ctx, "alice"),
```
gdbelvin: `TestEmptyGetAndUpdate` uses the following to avoid an import reference to `impl`.
```suggestion
			opts:           env.CallOpts("alice"),
```

core/integration/env.go
Zyqsempai: `cmd/gen-test-vectors` has call to `impl/integration`
and `impl/integration` has call to `integration/client_tests.go`
if we merge `cmd/gen-test-vectors` into `integration/client_tests.go` we get cycle.

gdbelvin: I think the link between `cmd/gen-test-vectors` to `impl/integration` can be split:
1. `main()` in `cmd/gen-test-vectors` is now `impl/integration/main_test.go`
1. `GenerateTestVectors` used to depend on `impl/integration/env.go`, but this can be swapped with `core/integration/env.go` 

There are a few other references to `impl` in `GenerateTestVectors` which I've annotated with fixes for.

Zyqsempai: Ok, i'l try to do it in this way.

core/integration/client_tests.go
gdbelvin: This was a pre existing issue. Fixed in #1197

Zyqsempai:
@gdbelvin Looks like done, but i have doubts, since i wasn't able to test it and i have strong feeling that i am confused a little bit;))

impl/integration/main_test.go
```diff
@@ -48,3 +61,14 @@ func TestIntegration(t *testing.T) {
 		})
 	}
 }
+
+func runGenerateTestVectors() error {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	env, err := NewEnv(ctx)
+	if err != nil {
+		return fmt.Errorf("Could not create Env: %v", err)
```
golangcibot: error strings should not be capitalized or end with punctuation or a newline (from `golint`)

PR #1198, opened 2019-02-08
Support testing with docker

### LINKED ISSUE #1184: Integration tests are not working with docker-compose mysql
After fixing this one https://github.com/google/keytransparency/issues/1182 I am getting thi error.
```
 Could not create Env: env: failed to open database: error running statement "CREATE DATABASE trl_1549349836763157901": Error 1045: Access denied for user 'root'@'172.21.0.1' (using password: NO)
```
It's an error from trillian testdb.go and it's supper confusing for me, the first `sql.Open` request is made outside on `127.0.0.1:3306` but in next step `db.ExecContext()` makes it on `172.21.0.1`

So the question is for what purpose do we have docker-compose mysql, if tests are not interacting with it?
Even though, it should work, port is exposed, we have `MYSQL_RANDOM_ROOT_PASSWORD: "yes"`.

gdbelvin:
Rather than setting up a full mysql instance, which can be a hassle, support testing against the local database, run by docker with `docker-compose up -d db`. 

The tests require the mysql database to have an empty root password, which is not suitable for production, so this PR splits out the docker-compose file into a separate `dev` and `prod` configurations: 
- dev: `docker-compose up`
- prod: `docker-compose up -f docker-compose.yml docker-compose.prod.yml`

This feature depends on google/trillian#1425
Fixes #1184

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,35 +87,20 @@ NB A default for the Key Transparency server URL is being used here. The default
 
 ## Running the server
 
-### Install
 1. [OpenSSL](https://www.openssl.org/community/binaries.html)
 1. [Docker](https://docs.docker.com/engine/installation/)
    - Docker Engine 1.13.0+ `docker version -f '{{.Server.APIVersion}}'`
    - Docker Compose 1.11.0+ `docker-compose --version`
-1. `go get -u github.com/google/keytransparency/...`
-1. `go get -u github.com/google/trillian/...`
-1. `./scripts/prepare_server.sh -f`
 
-### Run
-1. Run Key Transparency
-
-  ```sh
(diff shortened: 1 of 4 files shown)


gdbelvin:
@Zyqsempai - see if this change, along with google/trillian#1425 makes testing with docker possible.

DazWilkin:
The revised docker-compose command doesn't work for me and I'm unclear what's intended:

```docker-compose up -f docker-compose.yml docker-compose.prod.yml```

**Comment by gdbelvin** [comment]
The command was in error.  It should be:
```
docker-compose -f docker-compose.yml docker-compose.prod.yml up
```

PR #1201, opened 2019-02-08
Add docker-compose integration test

### LINKED ISSUE #1209: Integration test for docker-compose configs
As suggested on PR https://github.com/google/keytransparency/pull/1169#pullrequestreview-199608491, adding an integration test that runs `docker-compose up` and interacts with KT would give some confidence that the Docker images and Docker Compose config files are working.

gdbelvin:
Make sure we don't run across issues in the `cmd` directory due to not actually running things! 

Fixes #1209


gdbelvin:
Note to self: we need to either pick docker as the db or the local mysql db engine. 
TODO: @Zyqsempai  Move the docker db port export flag to `docker-compose.override.yml` 
```
ERROR: for db  Cannot start service db: driver failed programming external connectivity on endpoint keytransparency_db_1 (983ae78cbde9491d7391a6e330646d5c6dfa6e81960fee0087144ad2534334c3): Error starting userland proxy: listen tcp 0.0.0.0:3306: bind: address already in use
```

Zyqsempai:
@gdbelvin Ok

RJPercival:
Test appears to be failing on Travis.

docker-compose.yml
```diff
@@ -30,10 +30,7 @@ services:
       - 9090:9090
 
   db:
-    image: us.gcr.io/key-transparency/db
-    build:
-      context: ../trillian
-      dockerfile: examples/deployment/docker/db_server/Dockerfile
+    image: gcr.io/trillian-opensource-ci/db_server
```
RJPercival: This doesn't explicitly specify ":latest", like the other image fields below do. Either add it here or remove it there for consistency.

scripts/integration.sh
```diff
@@ -0,0 +1,25 @@
...
+
+if [ ! -f genfiles/server.key ]; then
+	./scripts/prepare_server.sh -f
+fi
+
+docker-compose -f docker-compose.yml -f docker-compose.prod.yml up -d
+trap "docker-compose down" INT EXIT
+
+set +e
+wget -t 60 --spider --retry-connrefused --waitretry=1  0.0.0.0:443 || sleep 2
```
RJPercival: nit: 0.0.0.0 isn't typically used as an address to connect to. For a server, it means listen on all interfaces but it doesn't have obvious meaning for a client (Wikipedia says "In the Internet Protocol Version 4, the address 0.0.0.0 is a non-routable meta-address used to designate an invalid, unknown or non-applicable target"). It works for wget because it special-cases 0.0.0.0 to be equivalent to 127.0.0.1, but not all clients do this, so it's a bit confusing to encounter. The clearest thing to do is use 127.0.0.1.

RJPercival: `-t 60` means make 60 retries, but I suspect you wanted to retry for 60 seconds. 

```
  -t,  --tries=NUMBER              set number of retries to NUMBER (0 unlimits)
  -T,  --timeout=SECONDS           set all timeout values to SECONDS
```

RJPercival: Why sleep for 2 seconds when the first `wget` fails? It'll already have been waiting for quite some time at this point.

scripts/integration.sh
```diff
@@ -0,0 +1,25 @@
...
+wget -t 60 --spider --retry-on-http-error=404  1 0.0.0.0:443/v1/directories/default
```
RJPercival: Any idea what the "1" just before the URL does?

RJPercival: nit: Could prefix with "https://" rather than specifying port 443.
```suggestion
wget -t 60 --spider --retry-on-http-error=404  1 https://0.0.0.0/v1/directories/default
```

scripts/integration.sh
```diff
@@ -0,0 +1,25 @@
...
+set -e
+
+PASSWORD="foobar"
+go run ./cmd/keytransparency-client/main.go authorized-keys create-keyset --password=${PASSWORD}
+go run ./cmd/keytransparency-client/main.go post foo@bar.com \
```
RJPercival: nit: Don't need "/main.go"
```suggestion
go run ./cmd/keytransparency-client post foo@bar.com \
```

gdbelvin: I used to need it. I wonder when that changed...

scripts/integration.sh
gdbelvin: nope -- removed

scripts/integration.sh
```diff
@@ -10,13 +10,13 @@ docker-compose -f docker-compose.yml -f docker-compose.prod.yml up -d
 trap "docker-compose down" INT EXIT
 
 set +e
-wget -t 60 --spider --retry-connrefused --waitretry=1  0.0.0.0:443 || sleep 2
-wget -t 60 --spider --retry-on-http-error=404  1 0.0.0.0:443/v1/directories/default
+wget -T 60 --spider --retry-connrefused localhost:443
+wget -T 60 --spider --retry-on-http-error=404 --no-check-certificate https://localhost/v1/directories/default
```
RJPercival: Are you sure you *don't* want to fail the script here if you can't get a 200 response? It seems it'd make it easier to tell what went wrong. Is it just because you've found it to be a bit unreliable?

scripts/integration.sh
```diff
@@ -10,13 +10,13 @@ docker-compose -f docker-compose.yml -f docker-compose.prod.yml up -d
 trap "docker-compose down" INT EXIT
 
 set +e
-wget -t 60 --spider --retry-connrefused --waitretry=1  0.0.0.0:443 || sleep 2
-wget -t 60 --spider --retry-on-http-error=404  1 0.0.0.0:443/v1/directories/default
+wget -T 60 --spider --retry-connrefused localhost:443
```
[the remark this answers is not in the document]
RJPercival: Also, if you merged them, you could get rid of the `set` commands either side and just append `|| echo "Healthcheck failed - continuing anyway"`, or something like that.

scripts/integration.sh
gdbelvin: Fixed

PR #1208, opened 2019-02-11
Set write watermark in keyserver

gdbelvin:
Add an interface to return metadata about writes to the mutation queue so it can be exposed in the keyserver as a metric.

### DIFF
--- a/impl/sql/mutationstorage/queue.go
+++ b/impl/sql/mutationstorage/queue.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/golang/glog"
 	"github.com/golang/protobuf/proto"
+	"github.com/google/keytransparency/core/keyserver"
 	"github.com/google/keytransparency/core/mutator"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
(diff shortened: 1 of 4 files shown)


impl/sql/mutationstorage/queue.go
```diff
@@ -46,27 +47,32 @@ func (m *Mutations) AddLogs(ctx context.Context, directoryID string, logIDs ...i
...
 		if err != nil {
-			return err
+			return nil, err
 		}
 		updateData = append(updateData, data)
 	}
 	// TODO(gbelvin): Implement retry with backoff for retryable errors if
 	// we get timestamp contention.
-	return m.send(ctx, time.Now(), directoryID, logID, updateData...)
+	ts := time.Now()
```
pavelkalinnikov: Looks dangerous to me, as servers have time skew. Shouldn't the returned watermark be based on what the storage thinks the "time" is? Consider returning a "strong" watermark from the `send` call below.

gdbelvin: The write implementation first verifies that the timestamp provided is greater than any other timestamp currently in the database.  A server that has a slow clock will simply fail more often when racing another writer that has a clock further ahead.

gdbelvin: I provide the watermark here instead of returning it so that I can have meaningful unit tests.

pavelkalinnikov: Why does returning the same watermark that [this](https://github.com/google/keytransparency/blob/edd2c90f078bec0d40021e5c076e36c9899fb2e1/impl/sql/mutationstorage/queue.go#L138) line generates in `m.send` not work?

pavelkalinnikov: (or just the timestamp)

gdbelvin: That line doesn't query the current time, it just convertes the passed in `ts time.Time` into an integer `ts.UnixNano()`

pavelkalinnikov: Ah, indeed.

PR #1210, opened 2019-02-12
Follow tink updates

gdbelvin:
google/tink#13c0094

### DIFF
--- a/cmd/keytransparency-client/cmd/hammer.go
+++ b/cmd/keytransparency-client/cmd/hammer.go
@@ -21,7 +21,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/google/tink/go/tink"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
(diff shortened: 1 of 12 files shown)


cmd/keytransparency-client/cmd/hammer.go
```diff
@@ -68,13 +68,13 @@ var hammerCmd = &cobra.Command{
...
 		}
-		handle, err := tink.NewKeysetHandleFromReader(
+		handle, err := keyset.Read(
 			&tinkio.ProtoKeysetFile{File: keysetFile},
 			masterKey)
 		if err != nil {
 			log.Fatal(err)
 		}
-		keyset = handle
+		ks = handle
```
thaidn: This is a bit confusing. Can you directly use handle, without reassigning to ks?

cmd/keytransparency-client/cmd/keys.go
```diff
@@ -36,7 +34,7 @@ var (
 	masterPassword string
 )
 
-var keyset *tink.KeysetHandle
+var ks *keyset.Handle
```
thaidn: should be handle *keyset.Handle? or keysetHandle *keyset.Handle

cmd/keytransparency-client/cmd/hammer.go
gdbelvin: yeah, I can move this to the `RunE` function

PR #1212, opened 2019-02-13
Add logid to read metric

gdbelvin:
* Add logid to the sequencer read metric.

### DIFF
--- a/core/sequencer/server.go
+++ b/core/sequencer/server.go
@@ -62,7 +62,7 @@ func createMetrics(mf monitoring.MetricFactory) {
 	logEntryCount = mf.NewCounter(
 		"log_entry_count",
 		"Total number of log entries read since process start. Duplicates are not removed.",
-		directoryIDLabel)
+		directoryIDLabel, logIDLabel)
 	mapLeafCount = mf.NewCounter(
 		"map_leaf_count",
 		"Total number of map leaves written since process start. Duplicates are not removed.",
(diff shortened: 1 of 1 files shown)


core/keyserver/keyserver.go
```diff
@@ -52,7 +53,7 @@ func createMetrics(mf monitoring.MetricFactory) {
 	watermarkWritten = mf.NewGauge(
 		"watermark_written",
 		"High watermark of each input log that has been written",
-		directoryIDLabel, logIDLabel)
+		directoryIDLabel, logIDLabel, stateIDLabel)
```
golangcibot: undeclared name: stateIDLabel (from `typecheck`)

libinjG:
LGTM

core/keyserver/keyserver.go
```diff
@@ -41,6 +41,7 @@ import (
 const (
 	directoryIDLabel = "directoryid"
 	logIDLabel       = "logid"
+	stageIDLabel     = "stage"
```
pavelkalinnikov: Why is this label needed? Do you expect `watermarkWritten` to have at least one `stage` other than "written"? Same for other metrics - so far each of them has one stage with the same name as the metric.

gdbelvin: The long and the short is that I'm having trouble building a graph based in these three metrics: watermark_written, watermark_defined, watermark_applied. 

I had originally thought I could just graph them as three lines without difficulty. 
In practice, these three metrics get combined into one time series (one line), so I need to mark them separately so there's something to break them apart. That's one strategy - perhaps there is another way. @AlCutter ideas?

gdbelvin: We figured out a way to not need this

gdbelvin:
Removed stage label PTAL

PR #1220, opened 2019-02-21
Move reduceFn to mutator definition

gdbelvin:
Define a proper reducer function as an interface of the mutator. Fixes a TODO.
 
- Moves the reducer implementation from `core/runner` to `core/mutator/entry`. 
- Make the reducer commutative and associative. Fixes another TODO. 
- Emit errors in the reducer so they can be reported out as a metric in `core/sequencer`.

### DIFF
--- a/core/client/client.go
+++ b/core/client/client.go
@@ -127,7 +127,7 @@ func New(ktClient pb.KeyTransparencyClient,
 		Verifier:    ktVerifier,
 		cli:         ktClient,
 		DirectoryID: directoryID,
-		mutate:      entry.MutateFn,
+		mutate:      entry.ReduceFn,
 		RetryDelay:  retryDelay,
 	}
 }
--- a/core/client/client.go
+++ b/core/client/client.go
@@ -98,7 +98,7 @@ type Client struct {
 	Verifier
 	cli         pb.KeyTransparencyClient
 	DirectoryID string
-	mutate      mutator.ReduceMutationFn
+	reduce      mutator.ReduceMutationFn
 	RetryDelay  time.Duration
 	trusted     types.LogRootV1
 	trustedLock sync.Mutex
@@ -127,7 +127,7 @@ func New(ktClient pb.KeyTransparencyClient,
 		Verifier:    ktVerifier,
 		cli:         ktClient,
 		DirectoryID: directoryID,
-		mutate:      entry.ReduceFn,
+		reduce:      entry.ReduceFn,
 		RetryDelay:  retryDelay,
 	}
 }
--- a/core/mutator/entry/mutator.go
+++ b/core/mutator/entry/mutator.go
@@ -86,7 +86,7 @@ func ReduceFn(index []byte, msgs []*pb.EntryUpdate, leaves []*tpb.MapLeaf, emit 
 	newValue, err := MutateFn(oldValue, msg.Mutation)
 	if err != nil {
 		glog.Warningf("Mutate(): %v", err)
-		return nil // A bad mutation should not make the whole batch to fail.
+		return nil // A bad mutation should not cause the whole batch to fail.
 	}
 	leafValue, err := ToLeafValue(newValue)
 	if err != nil {
--- a/core/mutator/entry/mutator.go
+++ b/core/mutator/entry/mutator.go
@@ -19,6 +19,7 @@ import (
 	"crypto/sha256"
 	"errors"
 	"fmt"
+	"sort"
 
 	"github.com/golang/glog"
 	"github.com/golang/protobuf/proto"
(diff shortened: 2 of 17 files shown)


core/mutator/mutator.go
```diff
@@ -44,10 +45,12 @@ var (
 	ErrUnauthorized = errors.New("mutation: unauthorized")
 )
 
-// ReduceMutationFn takes the existing mapleaf and a new mutation and returns the new value for that map leaf.
-// ReduceMutationFn verifies that this is a valid mutation for this item.
-// ReduceMutationFn must be idempotent.
-type ReduceMutationFn func(existingMapLeafValue, mutation *pb.SignedEntry) (*pb.SignedEntry, error)
+// ReduceMutationFn takes all the mutations for an index an an auxillary input
```
golangcibot: `auxillary` is a misspelling of `auxiliary` (from `misspell`)

core/mutator/mutator.go
```diff
@@ -44,10 +45,12 @@ var (
 	ErrUnauthorized = errors.New("mutation: unauthorized")
 )
 
-// ReduceMutationFn takes the existing mapleaf and a new mutation and returns the new value for that map leaf.
-// ReduceMutationFn verifies that this is a valid mutation for this item.
-// ReduceMutationFn must be idempotent.
-type ReduceMutationFn func(existingMapLeafValue, mutation *pb.SignedEntry) (*pb.SignedEntry, error)
+// ReduceMutationFn takes all the mutations for an index an an auxiliary input
+// of existing mapleaf(s) returns a new map leaf.  ReduceMutationFn must be
+// idempotent, communative, and associative. i.e. must produce the same output
```
mhutchinson:
```suggestion
// idempotent, commutative, and associative. i.e. must produce the same output
```

core/client/client.go
```diff
@@ -127,7 +127,7 @@ func New(ktClient pb.KeyTransparencyClient,
 		Verifier:    ktVerifier,
 		cli:         ktClient,
 		DirectoryID: directoryID,
-		mutate:      entry.MutateFn,
+		mutate:      entry.ReduceFn,
```
mhutchinson: nit: the "mutate" property of the struct would be clearer if named "reducer", or similar.

core/mutator/entry/mutator.go
```diff
@@ -43,6 +44,46 @@ func MapLogItemFn(m *mutator.LogMessage, emit func(index []byte, mutation *pb.En
...
+		return fmt.Errorf("no msgs for index %x", index)
+	}
+
+	// TODO(gbelvin): Choose the mutation deterministically, regardless of the messages order.
+	// (optional): Select the mutation based on it's correctness.
+	msg := msgs[0]
+	newValue, err := MutateFn(oldValue, msg.Mutation)
+	if err != nil {
+		glog.Warningf("Mutate(): %v", err)
+		return nil // A bad mutation should not make the whole batch to fail.
```
mhutchinson:
```suggestion
		return nil // A bad mutation should not cause the whole batch to fail.
```

core/mutator/mutator.go
```diff
@@ -44,10 +45,12 @@ var (
 	ErrUnauthorized = errors.New("mutation: unauthorized")
 )
 
-// ReduceMutationFn takes the existing mapleaf and a new mutation and returns the new value for that map leaf.
-// ReduceMutationFn verifies that this is a valid mutation for this item.
-// ReduceMutationFn must be idempotent.
-type ReduceMutationFn func(existingMapLeafValue, mutation *pb.SignedEntry) (*pb.SignedEntry, error)
+// ReduceMutationFn takes all the mutations for an index an an auxiliary input
```
mhutchinson:
```suggestion
// ReduceMutationFn takes all the mutations for an index and an auxiliary input
```

core/mutator/entry/mutator.go
```diff
@@ -43,6 +44,46 @@ func MapLogItemFn(m *mutator.LogMessage, emit func(index []byte, mutation *pb.En
...
+		if err != nil {
+			return fmt.Errorf("entry.FromLeafValue(): %v", err)
+		}
+	}
+
+	if len(msgs) == 0 {
+		return fmt.Errorf("no msgs for index %x", index)
+	}
+
+	// TODO(gbelvin): Choose the mutation deterministically, regardless of the messages order.
```
mhutchinson: This TODO seems higher priority now that you've explicitly stated on the interface that it needs to be associative.

gdbelvin: Fixed

core/mutator/entry/mutator.go
```diff
@@ -61,6 +63,67 @@ func IsValidEntry(signedEntry *pb.SignedEntry) error {
...
+			continue
+		}
+		newEntries = append(newEntries, &pb.EntryUpdate{
+			Mutation:  newValue,
+			Committed: msg.Committed,
+		})
+	}
+	if len(newEntries) == 0 {
+		emitErr(fmt.Errorf("entry: no valid mutations for index %x", index))
+		return // No valid mutations for one index not cause the whole batch to fail.
```
mhutchinson:
```suggestion
		return // No valid mutations for one index should not cause the whole batch to fail.
```

core/mutator/entry/mutator.go
```diff
@@ -61,6 +63,67 @@ func IsValidEntry(signedEntry *pb.SignedEntry) error {
...
+	}
+	// Choose the mutation deterministically, regardless of the messages order.
+	sort.Slice(newEntries, func(i, j int) bool {
```
mhutchinson: Minor nit: this is slightly wasteful to sort the whole slice and then pick the smallest element. This op can be performed in O(n). We don't expect large n though, so it's fine.

core/mutator/mutator.go
```diff
@@ -50,10 +51,13 @@ var (
 // VerifyMutationFn verifies that a mutation is internally consistent.
 type VerifyMutationFn func(mutation *pb.SignedEntry) error
 
-// ReduceMutationFn takes the existing mapleaf and a new mutation and returns the new value for that map leaf.
-// ReduceMutationFn verifies that this is a valid mutation for this item.
-// ReduceMutationFn must be idempotent.
-type ReduceMutationFn func(existingMapLeafValue, mutation *pb.SignedEntry) (*pb.SignedEntry, error)
+// ReduceMutationFn takes all the mutations for an index and an auxiliary input
+// of existing mapleaf(s) returns a new map leaf.  ReduceMutationFn must be
+// idempotent, commutative, and associative. i.e. must produce the same output
```
mhutchinson: It's no longer clear to me this morning how "associative" can be sensibly interpreted for a function that returns a different type than it consumes.

gdbelvin: I'll save this for the next PR :-)

Issue #1177, opened 2019-02-01
Split Verification Function


gdbelvin:
Verifying the legitimacy of a mutation can be broken into two parts:
1. Verifying that it is well formed without knowing what the current map value is.  This verification should be checked by the `keyserver`. 

2. Verifying that any references to the `current` value in the map are correct. Both the first check and this check should be run in the `sequencer`.

Zyqsempai:
I think i will take this one.

gdbelvin:
Thanks!

Note there is a code link to this bug in `core/keyserver/keyserver.go:572`

Zyqsempai:
Hi, some clarification is needed: You say that first check should be checked in the keyserver but then you say that both of them should be ran in sequencer. Where is the truth;)?

gdbelvin:
The keyserver's check is for spam protection. 
The sequencer's check is the final check. It should do both, but if we trust the keyserver, it could get away with only doing the checks involving the `current` value. 

I feel like I just repeated myself -- happy to jump on a call to discuss if need.

Zyqsempai:
So, should we trust keyserver or not?;)

gdbelvin:
We trust the keyserver because the sequencer and the keyserver are run by
the same entity.
But it would still be nice to have an "extra sure" option to do both in the
sequencer. Would you mind adding the double check behind a flag?

Zyqsempai:
Nope, doesn't see any problem in doing that. 
Started working on this, and have few questions:
1) Should we keep validation functions in keyserver? even if now they will be consumed by sequencer?
2) We have that check https://github.com/google/keytransparency/blob/master/core/keyserver/validate.go#L66 but we couldn't get `vrf.PrivateKey` in sequencer in current configuration of trillian_client, should we change it, or skip that check?

gdbelvin:
Good questions.

1. Let's keep the VRF verification in the keyserver. The Sequencer doesn't have the `vrf.PrivateKey`, and in order to keep the sequencer fairly generic, I think we should keep it that way.  We could change that with some refactoring, but let's save that for a separate PR. 

2. The verification function that needs to be split is here:
https://github.com/google/keytransparency/blob/master/core/mutator/entry/mutator.go#L48

Specifically, the items that can be broken into Step A and B
A
- Size check (L50)
- Format checks (L55-L60)
- At least one signature with a key from the new authorized_key set should exist. (Mixed in with `verifyKeys`) 

B
- Verification of previous entry hash (L66) 
[...]

PR #1221, opened 2019-02-22
Made logical split for mutation validation function

Zyqsempai:
Fixes https://github.com/google/keytransparency/issues/1177

Since `MutateFn` is used in many places as a argumant, I decided not split it in two functions and not multiply arguments, I made logical split by pathing flag `fullCheck`.


gdbelvin:
#1222 Includes this commit

PR #1222, opened 2019-02-22
Split Mutation Verification Function

### LINKED ISSUE #1177: Split Verification Function

gdbelvin:
@Zyqsempai I've built on your PR a bit. (See #10968b4) I didn't want to presume to push to your branch, so I've opened a new one. Does this look reasonable? 

This PR splits the mutation function into two components:
1. `IsValidEntry` verifies the internal correctness of a mutation on it's own. 
2. `MutateFn` verifies pairs of entries and uses `IsValidEntry` as a helper. 

`IsValidEntry` also imposes the *new* requirement that every mutation be signed by one of the authorized keys in that same mutation. 

The Keyserver frontend will only check internal correctness on incoming mutations.  It doesn't make sense to do a full check at the keyserver because the ordering of mutations at that point has not been determined. 

The Sequencer will both check internal consistency and pairwise correctness in order to compute the next value for a user. 

Fixes #1177

### DIFF
--- a/core/mutator/entry/mutation_test.go
+++ b/core/mutator/entry/mutation_test.go
@@ -114,7 +114,7 @@ func TestCreateAndVerify(t *testing.T) {
 			if err != nil {
 				t.Fatalf("FromLeafValue(%v): %v", tc.old, err)
 			}
-			newSignedEntry, err := MutateFn(oldSignedEntry, update.GetMutation())
+			newSignedEntry, err := MutateFn(oldSignedEntry, update.GetMutation(), true)
 			if err != nil {
 				t.Fatalf("Mutate(%v): %v", update.GetMutation(), err)
 			}
--- a/core/mutator/entry/mutation_test.go
+++ b/core/mutator/entry/mutation_test.go
@@ -114,7 +114,7 @@ func TestCreateAndVerify(t *testing.T) {
 			if err != nil {
 				t.Fatalf("FromLeafValue(%v): %v", tc.old, err)
 			}
-			newSignedEntry, err := MutateFn(oldSignedEntry, update.GetMutation(), true)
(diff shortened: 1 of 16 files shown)


core/mutator/entry/mutator_test.go
```diff
@@ -177,7 +177,7 @@ func TestCheckMutation(t *testing.T) {
 			continue
 		}
 
-		if _, got := MutateFn(tc.old, m); got != tc.err {
+		if _, got := MutateFn(tc.old, m, true); got != tc.err {
```
golangcibot: too many arguments (from `typecheck`)

core/mutator/entry/mutation_test.go
```diff
@@ -114,7 +114,7 @@ func TestCreateAndVerify(t *testing.T) {
 			if err != nil {
 				t.Fatalf("FromLeafValue(%v): %v", tc.old, err)
 			}
-			newSignedEntry, err := MutateFn(oldSignedEntry, update.GetMutation())
+			newSignedEntry, err := MutateFn(oldSignedEntry, update.GetMutation(), true)
```
Zyqsempai: Don't forget to remove unused param here.

Zyqsempai:
@gdbelvin I like your version more then mine;) So, no worries.

gdbelvin:
Marking `cla: yes` with @Zyqsempai 's permission to merge with their commit.

Zyqsempai:
@gdbelvin What should i do to make googlebot stop complaining about my commit?;)

PR #1224, opened 2019-02-26
Make the reduceFn associative

gdbelvin:
Follow up on #1220

### DIFF
--- a/core/mutator/entry/entry.go
+++ b/core/mutator/entry/entry.go
@@ -16,6 +16,7 @@
 package entry
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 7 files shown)


core/sequencer/runner/native.go
```diff
@@ -51,19 +65,19 @@ func ApplyMutations(reduceFn mutator.ReduceMutationFn,
 // Joined is the result of a CoGroupByKey on []*MapLeaf and []*IndexedUpdate.
```
mhutchinson: Comment is stale now.

core/sequencer/runner/native.go
```diff
@@ -51,19 +65,19 @@ func ApplyMutations(reduceFn mutator.ReduceMutationFn,
...
 type Joined struct {
 	Index  []byte
-	Leaves []*tpb.MapLeaf
+	Leaves []*pb.EntryUpdate
 	Msgs   []*pb.EntryUpdate
 }
 
 // Join pairs up MapLeaves and IndexedUpdates by index.
```
mhutchinson: Stale.

core/mutator/entry/entry.go
```diff
@@ -16,12 +16,43 @@
...
+	"fmt"
+
 	"github.com/golang/glog"
 	"github.com/golang/protobuf/proto"
 
 	pb "github.com/google/keytransparency/core/api/v1/keytransparency_go_proto"
+	tpb "github.com/google/trillian"
 )
 
+func FromMapLeaf(leaf *tpb.MapLeaf) (*pb.EntryUpdate, error) {
```
mhutchinson: [nit, not blocking] some of the naming here still feels a little uncomfortable. It feels a little strange to have a method to turn a map leaf into an "update". An Update feels like something that would wrap a logical Value type until it is applied to the map. Once it is applied to the Map I would expect to be talking about Leaves/Nodes and the Value type.

gdbelvin: The types are admittedly non-ideal. Trying to work toward something better. 

## Current State
`LogItem` is currently an `EntryUpdate`. 
`LogItems` get mapped to `IndexedUpdates` - Perhaps there's a better name / type for this. 
`IndexedUpdates` get grouped by index 
`EntryUpdate` + `[]EntryUpdate` gets combined `EntryUpdate`.
The reducer adds the index back, to get `IndexedUpdate` 
In the final step `IndexedUpdate` gets converted to `MapLeaf`.

## Future State?
`LogItem` is an `EntryUpdate`.  (Also an IndexedValue?) 
`LogItems` get mapped to `???` / `IndexedValue`
`IndexedValue`s get grouped by index 
`Value` + `[]Value` gets combined to a new `Value`.
*Question: should we be reducing `[]Value` or `[]IndexedValue` ?*
The reducer adds the index back, to get `IndexedValue` 
In the final step `IndexedValue` gets converted to `MapLeaf`.

mhutchinson: I'm happy reducing an IndexedValue. I can imagine some situations where the key might be relevant.

Your proposed changes sound much clearer to me. Thanks for the clarification :-)

gdbelvin: Moved to IndexedValue

core/sequencer/runner/native_test.go
```diff
@@ -18,27 +18,25 @@ import (
...
 		{
 			desc:   "onerow",
-			leaves: []*tpb.MapLeaf{{Index: []byte("A")}},
-			msgs:   []*mapper.IndexedUpdate{{Index: []byte("A"), Update: &pb.EntryUpdate{}}},
+			leaves: []*entry.IndexedValue{{Index: []byte("A"), Value: &pb.EntryUpdate{UserId: "bob"}}},
+			msgs:   []*entry.IndexedValue{{Index: []byte("A"), Value: &pb.EntryUpdate{}}},
 			want: []*Joined{{
 				Index:  []byte("A"),
-				Leaves: []*tpb.MapLeaf{{Index: []byte("A")}},
+				Leaves: []*pb.EntryUpdate{{UserId: "bob"}},
```
golangcibot: unknown field `Leaves` in struct literal (from `typecheck`)

PR #1226, opened 2019-03-01
Reduce batch_size and timeouts

gdbelvin:
This test was flakey

### DIFF
--- a/core/integration/client_tests.go
+++ b/core/integration/client_tests.go
@@ -135,7 +135,8 @@ func TestBatchUpdate(ctx context.Context, env *Env, t *testing.T) []testdata.Res
 	}{
 		{desc: "zero", userIDs: nil},
 		{desc: "one", userIDs: []string{"test"}},
-		{desc: "100", userIDs: genUserIDs(100)},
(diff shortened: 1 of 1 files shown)


core/integration/client_tests.go
```diff
@@ -148,14 +149,11 @@ func TestBatchUpdate(ctx context.Context, env *Env, t *testing.T) []testdata.Res
...
-			cctx, cancel := context.WithTimeout(ctx, env.Timeout)
-			defer cancel()
-
-			mutations, err := env.Client.BatchCreateMutation(cctx, users)
+			mutations, err := env.Client.BatchCreateMutation(ctx, users)
 			if err != nil {
 				t.Fatalf("BatchCreateMutation(): %v", err)
 			}
-			if err := env.Client.BatchQueueUserUpdate(cctx, mutations, signers1); err != nil {
+			if err := env.Client.BatchQueueUserUpdate(ctx, mutations, signers1); err != nil {
```
Martin2112: The timeout has gone, do you not need it?

gdbelvin: The timeout is not needed because tests as a whole have timeouts.

PR #1228, opened 2019-03-04
Fix: Allow multiple updates to the same user in the same batch

### LINKED ISSUE #1225: Don't request the same map leaf more than once.

gdbelvin:
If two updates to the same user occurred during the same batch, the sequencer would request the same index in the map multiple times, causing subsequent verifications to fail.

Adds a failing test and then a fix.
Fixes #1225

### DIFF
--- a/core/sequencer/server.go
+++ b/core/sequencer/server.go
@@ -299,7 +299,7 @@ func (s *Server) ApplyRevision(ctx context.Context, in *spb.ApplyRevisionRequest
 	for _, iv := range indexedValues {
 		groupByIndex[string(iv.Index)] = true
 	}
-	indexes := make([][]byte, 0, len(indexedValues))
+	indexes := make([][]byte, 0, len(groupByIndex))
 	for i := range groupByIndex {
 		indexes = append(indexes, []byte(i))
 	}
--- a/core/sequencer/server.go
+++ b/core/sequencer/server.go
@@ -295,9 +295,13 @@ func (s *Server) ApplyRevision(ctx context.Context, in *spb.ApplyRevisionRequest
 	}
 
 	// Collect Indexes.
-	indexes := make([][]byte, 0, len(indexedValues))
+	groupByIndex := make(map[string]bool)
 	for _, iv := range indexedValues {
-		indexes = append(indexes, iv.Index)
+		groupByIndex[string(iv.Index)] = true
+	}
+	indexes := make([][]byte, 0, len(indexedValues))
+	for i := range groupByIndex {
+		indexes = append(indexes, []byte(i))
 	}
 
 	// Read Map.
--- a/core/sequencer/server_test.go
+++ b/core/sequencer/server_test.go
@@ -26,12 +26,16 @@ import (
 	"github.com/google/tink/go/tink"
 	"github.com/google/trillian/monitoring"
 	"github.com/google/trillian/types"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 
 	"github.com/google/keytransparency/core/mutator"
 	"github.com/google/keytransparency/core/mutator/entry"
 
 	spb "github.com/google/keytransparency/core/sequencer/sequencer_go_proto"
 	tpb "github.com/google/trillian"
+	tclient "github.com/google/trillian/client"
 )
 
 const directoryID = "directoryID"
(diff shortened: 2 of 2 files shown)


core/sequencer/server.go
```diff
@@ -295,9 +295,13 @@ func (s *Server) ApplyRevision(ctx context.Context, in *spb.ApplyRevisionRequest
 	}
 
 	// Collect Indexes.
-	indexes := make([][]byte, 0, len(indexedValues))
+	groupByIndex := make(map[string]bool)
 	for _, iv := range indexedValues {
-		indexes = append(indexes, iv.Index)
+		groupByIndex[string(iv.Index)] = true
+	}
+	indexes := make([][]byte, 0, len(indexedValues))
```
pavelkalinnikov: How about `len(groupByIndex)`?

core/sequencer/server_test.go
```diff
@@ -229,3 +252,68 @@ func TestHighWatermarks(t *testing.T) {
...
+			t.Fatalf("SerializeAndSign(): %v", err)
+		}
+		log0 = append(log0, mutator.LogMessage{
+			ID:        int64(i),
+			Mutation:  update.Mutation,
+			ExtraData: update.Committed},
+		)
+	}
+
+	s := Server{
```
pavelkalinnikov: It seems like the only reason to create `Server` this way is that a non-default `trillian` field can't be initialized in the constructor. How about calling `NewServer`, and updating the `trillian` field post-factum? Also, this would allow removing line 258.
Alternatively, maybe add some shorter non-exported `newServerImpl` method to `Server`? The `NewServer` method would call that one.

core/sequencer/server_test.go
```diff
@@ -229,3 +252,68 @@ func TestHighWatermarks(t *testing.T) {
...
+				MapClient:     MapClient{&tclient.MapClient{Conn: &fakeMapConn{}}},
+				latestMapRoot: &types.MapRootV1{Revision: uint64(mapRev)},
+			},
+		},
+	}
+
+	if _, err := s.ApplyRevision(ctx, &spb.ApplyRevisionRequest{
+		DirectoryId: directoryID,
+		Revision:    1,
+	}); status.Code(err) != codes.Unimplemented {
```
mhutchinson: Maybe this is a normal way to test in Go, but this feels a little grim to me. If I'm reading this correctly you stub out GetLeavesByRevision such that it returns Unimplemented and then rely on the code failing in specifically that way to measure success. It would some more wholesome to have a fake for the Map. As written, this test could succeed if a future change was introduced that simply returned an Unimplemented error on ApplyRevision, which isn't really successfully handling multiple updates in the same batch.

pavelkalinnikov: How did this fail previously?

gdbelvin: Indeed this is a bit of a hack. 
The recommended way to test things in go is to use the real server objects. Mocks are even discouraged. Code that is going to be imported, however, faces a bit of a challenge doing this because the open source dependencies are different. I'm trying avoid referencing anything in `impl` from `core`.  

This leaves us with a few choices:
- Return a more specific error we can test for directly. 
- Move to test to `core/integration` where we can inject real dependencies
- Mocks don't really work when every bit of info is cryptographically verified, unless we also supply a mock verifier that doesn't care.

gdbelvin: The previous failure was from the trillian map client complaining that there were multiple keys returned for a single index.

core/sequencer/server_test.go
```diff
@@ -88,16 +97,30 @@ func (m *fakeMap) GetAndVerifyLatestMapRoot(_ context.Context) (*tpb.SignedMapRo
...
+	}
+	return meta, nil
+}
+
+type fakeMapConn struct {
+	tpb.TrillianMapClient
+}
+
+func (m *fakeMapConn) GetLeavesByRevision(_ context.Context, _ *tpb.GetMapLeavesByRevisionRequest, _ ...grpc.CallOption) (*tpb.GetMapLeavesResponse, error) {
+	return nil, status.Errorf(codes.Unimplemented, "unimplemented")
```
pavelkalinnikov: As per discussion, would it be possible to test that this call doesn't get duplicates indices? This would be a better indication of what this PR aims for.

core/sequencer/server_test.go
gdbelvin: Added better error checking

core/sequencer/server_test.go
gdbelvin: Yes, but that's the code being fixed... so unfortunately it won't be in test vector format.

gdbelvin:
PTAL

mhutchinson:
The new test seems better. It still has a flaw that any other code path that raised codes.Unimplemented would cause the test to pass, but I can live with those odds.

core/sequencer/server_test.go
pavelkalinnikov: Not sure I understand. Line 316 checks that `Unimplemented` is returned. I propose to still return `Unimplemented`, but record the request in `fakeMapConn`, and then check that there are no duplicates in this request (gmock-like). As far as I understand, checking this would distinguish the old behavior from the new one.

gdbelvin:
LGTM?

pavelkalinnikov:
To resolve Martin's concern, you could return a pre-defined unexported error, and check equality instead of the error code:

`var errNoDuplicates = status.Errorf(codes.Unimplemented, "Success! No Duplicates")`

PR #1232, opened 2019-03-05
Read LogMessages in a MapFn

gdbelvin:
Brings `ApplyRevision` closer to looking more and more like a proper pipeline with the intention of having better metrics for each pipeline stage. 

1. Map MetaData -> []SourceSlice
2. Map SourceSlice -> []LogItem

Questions:
- If we error while reading log items we should probably hard fail. 
- Using `emit(*)` constructs looses ordering information. At the moment we preserve it because of an implementation detail in the runner, but we should probably come up with a better way of preserving order.


core/sequencer/runner/native.go
```diff
@@ -58,8 +60,41 @@ func Join(leaves []*entry.IndexedValue, msgs []*entry.IndexedValue) []*Joined {
 	return ret
 }
 
+// MapMetaFn emits a source slice for every map slice.
+type MapMetaFn func(meta *spb.MapMetadata, emit func(*spb.MapMetadata_SourceSlice))
+
+// DoMapMetaFn runs MapMetaFn on meta and collects the outputs.
+func DoMapMetaFn(fn MapMetaFn, meta *spb.MapMetadata) []*spb.MapMetadata_SourceSlice {
+	outs := make([]*spb.MapMetadata_SourceSlice, 0, 1)
```
pavelkalinnikov: `1` -> `len(meta.GetSources())`?

core/sequencer/server_test.go
```diff
@@ -198,11 +200,13 @@ func TestReadMessages(t *testing.T) {
 			{LogId: 1, LowestInclusive: 1, HighestExclusive: 11},
 		}}},
 	} {
-		msgs, err := s.readMessages(ctx, directoryID, tc.meta, tc.batchSize)
+
```
pavelkalinnikov: unnecessary blank like

PR #1233, opened 2019-03-05
Check mastership

gdbelvin:


core/sequencer/server.go
```diff
@@ -327,6 +327,10 @@ func (s *Server) ApplyRevision(ctx context.Context, in *spb.ApplyRevisionRequest
 		return nil, err
 	}
 
+	// Verify that we are still master.
+	if err := ctx.Err(); err != nil {
```
pavelkalinnikov: Would `SetLeavesAtRevision` not return if `ctx` is done?

gdbelvin: I hope so, I can't seem to actually find the code that would do that

pavelkalinnikov: I guess it should be in grpc code that underpins the call?

gdbelvin:
Redundant PR

PR #1234, opened 2019-03-05
Complete ApplyRevisions MR pipeline

gdbelvin:
Fold the nebulously named `ApplyMutations` into separate MR functions. 
This makes `ApplyRevision` a contiguous series of MR steps that we can optimize and add metrics to going forward.

### DIFF
--- a/core/sequencer/runner/native.go
+++ b/core/sequencer/runner/native.go
@@ -88,10 +88,10 @@ func DoMapMapLeafFn(fn MapMapLeafFn, leaves []*tpb.MapLeaf) ([]*entry.IndexedVal
 }
 
 // ReduceMutationFn takes all the mutations for an index and an auxiliary input
-// of existing mapleaf(s) returns a new map leaf.  ReduceMutationFn must be
-// idempotent, commutative, and associative. i.e. must produce the same output
-// regardless of input order or grouping, and it must be safe to run multiple
-// times.
+// of existing mapleaf(s) and emits a new value for the index.
+// ReduceMutationFn must be  idempotent, commutative, and associative.  i.e.
+// must produce the same output  regardless of input order or grouping,
+// and it must be safe to run multiple times.
 type ReduceMutationFn func(msgs []*pb.EntryUpdate, leaves []*pb.EntryUpdate,
 	emit func(*pb.EntryUpdate), emitErr func(error))
 
@@ -111,6 +111,7 @@ func DoReduceFn(reduceFn ReduceMutationFn, joined []*Joined, emitErr func(error)
 }
 
 // DoMarshalIndexedValues executes Marshal on each IndexedValue
+// If marshal fails, it will emit an error and continue with a subset of ivs.
 func DoMarshalIndexedValues(ivs []*entry.IndexedValue, emitErr func(error)) []*tpb.MapLeaf {
 	ret := make([]*tpb.MapLeaf, 0, len(ivs))
 	for _, iv := range ivs {
--- a/core/sequencer/runner/native.go
+++ b/core/sequencer/runner/native.go
@@ -18,44 +18,12 @@ package runner
 import (
 	"fmt"
 
-	"github.com/golang/glog"
-
 	"github.com/google/keytransparency/core/mutator"
 	"github.com/google/keytransparency/core/mutator/entry"
-	"github.com/google/keytransparency/core/sequencer/mapper"
 
 	pb "github.com/google/keytransparency/core/api/v1/keytransparency_go_proto"
 	tpb "github.com/google/trillian"
 )
-
-// ApplyMutations takes the set of mutations and applies them to given leaves.
-// Returns a list of map leaves that should be updated.
-func ApplyMutations(reduceFn mutator.ReduceMutationFn,
-	indexedUpdates []*entry.IndexedValue, leaves []*tpb.MapLeaf, emitErr func(error)) ([]*tpb.MapLeaf, error) {
-
-	indexedLeaves, err := DoMapMapLeafFn(mapper.MapMapLeafFn, leaves)
-	if err != nil {
-		return nil, err
-	}
-
-	joined := Join(indexedLeaves, indexedUpdates)
-
-	ret := make([]*tpb.MapLeaf, 0, len(joined))
-	for _, j := range joined {
-		reduceFn(j.Values1, j.Values2,
-			func(e *pb.EntryUpdate) {
-				mapLeaf, err := (&entry.IndexedValue{Index: j.Index, Value: e}).Marshal()
-				if err != nil {
-					emitErr(err)
-				}
-				ret = append(ret, mapLeaf)
-			},
-			func(err error) { emitErr(fmt.Errorf("reduceFn on index %x: %v", j.Index, err)) },
-		)
-	}
-	glog.V(2).Infof("ApplyMutations applied %v mutations to %v leaves", len(indexedUpdates), len(leaves))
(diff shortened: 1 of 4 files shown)


core/sequencer/runner/native.go
```diff
@@ -118,3 +86,39 @@ func DoMapMapLeafFn(fn MapMapLeafFn, leaves []*tpb.MapLeaf) ([]*entry.IndexedVal
 	}
 	return outs, nil
 }
+
+// ReduceMutationFn takes all the mutations for an index and an auxiliary input
+// of existing mapleaf(s) returns a new map leaf.  ReduceMutationFn must be
+// idempotent, commutative, and associative. i.e. must produce the same output
+// regardless of input order or grouping, and it must be safe to run multiple
+// times.
+type ReduceMutationFn func(msgs []*pb.EntryUpdate, leaves []*pb.EntryUpdate,
```
mhutchinson: This type is now defined twice. I'm used to duplication being a smell. Is there a reason this isn't?

core/sequencer/runner/native.go
```diff
@@ -118,3 +86,39 @@ func DoMapMapLeafFn(fn MapMapLeafFn, leaves []*tpb.MapLeaf) ([]*entry.IndexedVal
...
+			func(e *pb.EntryUpdate) {
+				ret = append(ret, &entry.IndexedValue{Index: j.Index, Value: e})
+			},
+			func(err error) { emitErr(fmt.Errorf("reduceFn on index %x: %v", j.Index, err)) },
+		)
+	}
+	return ret
+}
+
+// DoMarshalIndexedValues executes Marshal on each IndexedValue
```
mhutchinson: It looks like some of the elements in the result could be nil. Is this worth documenting, or generally understood behaviour for Go functions?

core/sequencer/server.go
```diff
@@ -313,14 +315,24 @@ func (s *Server) ApplyRevision(ctx context.Context, in *spb.ApplyRevisionRequest
...
 	if err != nil {
 		return nil, err
 	}
 
+	// GroupByIndex.
+	joined := runner.Join(indexedLeaves, indexedValues)
+
+	// Apply mutations to values.
+	newIndexedLeaves := runner.DoReduceFn(entry.ReduceFn, joined,
+		func(err error) { glog.Warning(err); mutationFailures.Inc(err.Error()) },
```
mhutchinson: This can be emitErrFn. But it strikes me we may want different counters for the different stages of the pipeline?

core/sequencer/runner/native.go
gdbelvin: Go encourages interfaces to be declared near the interface consumers, not the producers:
https://blog.chewxy.com/2018/03/18/golang-interfaces/

This keeps packages separate and not tied together. 

I could be persuaded that the client needs a reference to the `mapper`, but that feels like an odd dependency.

core/sequencer/runner/native.go
gdbelvin: Generally nil values in go are not expected.  Typically they should be accompanied by an error. 
Maybe I'm missing something. What's the path to producing a nil value in this function?

core/sequencer/server.go
gdbelvin: Thanks for that catch. 

Yes, different errors for different stages of the pipeline - up next! (with metrics too)

core/sequencer/runner/native.go
mhutchinson: On line 117 (mapLeaf == nil xor err == nil) from my reading of Marshal. You handle err by emitting it, but then carry on to poke the nil mapLeaf into the slice. This means you could get a nil value in there, which might surprise anyone iterating over it.

Not sure what the right approach is here now we have the emitErr mechanism. You could either fully fail, or at least document that the result contain a subset of the expected values?

core/sequencer/runner/native.go
mhutchinson: Nice. I like the ideology. I guess the smell here then is that they both have the same documentation string, and any updates to it may not be propagated to the other place. I will need to spend more time living away from Java to understand how this situation should be handled. For now I trust your wisdom :-)

core/sequencer/runner/native.go
```diff
@@ -118,3 +86,39 @@ func DoMapMapLeafFn(fn MapMapLeafFn, leaves []*tpb.MapLeaf) ([]*entry.IndexedVal
...
+func DoMarshalIndexedValues(ivs []*entry.IndexedValue, emitErr func(error)) []*tpb.MapLeaf {
+	ret := make([]*tpb.MapLeaf, 0, len(ivs))
+	for _, iv := range ivs {
+		mapLeaf, err := iv.Marshal()
+		if err != nil {
+			emitErr(err)
```
gdbelvin: TODO: add a continue statement here

core/sequencer/server.go
```diff
@@ -248,6 +249,7 @@ func (s *Server) DefineRevisions(ctx context.Context,
 }
 
 // readMessages returns the full set of EntryUpdates defined by sources.
+
```
pavelkalinnikov: I'd remove the blank line.

core/sequencer/server.go
```diff
@@ -313,14 +315,22 @@ func (s *Server) ApplyRevision(ctx context.Context, in *spb.ApplyRevisionRequest
 		return nil, err
 	}
 
-	// Apply mutations to values.
-	newLeaves, err := runner.ApplyMutations(entry.ReduceFn, indexedValues, leaves,
-		func(err error) { glog.Warning(err); mutationFailures.Inc(err.Error()) },
-	)
+	// Map Map Leaves
```
pavelkalinnikov: This comment doesn't convey new information, compared to the function name below. Please remove it, or make it more specific, e.g. `// Convert Trillian map leaves into indexed KT updates.`

core/sequencer/server.go
```diff
@@ -313,14 +315,22 @@ func (s *Server) ApplyRevision(ctx context.Context, in *spb.ApplyRevisionRequest
...
 	}
 
+	// GroupByIndex.
+	joined := runner.Join(indexedLeaves, indexedValues)
+
+	// Apply mutations to values.
+	newIndexedLeaves := runner.DoReduceFn(entry.ReduceFn, joined, emitErrFn)
+	glog.V(2).Infof("DoReduceFn reduced %v values on %v indexes", len(indexedValues), len(joined))
+
+	// Map IndexedValues
```
pavelkalinnikov: `// Marshal the updated indexed values back to Trillian map leaves.`?

core/sequencer/runner/native.go
```diff
@@ -118,3 +86,39 @@ func DoMapMapLeafFn(fn MapMapLeafFn, leaves []*tpb.MapLeaf) ([]*entry.IndexedVal
 	}
 	return outs, nil
 }
+
+// ReduceMutationFn takes all the mutations for an index and an auxiliary input
+// of existing mapleaf(s) returns a new map leaf.  ReduceMutationFn must be
```
pavelkalinnikov: Please fix the sentence around "returns".

core/sequencer/runner/native.go
gdbelvin: Added a `continue` and documented the results might be a subset.

core/sequencer/server.go
gdbelvin: Yep, not sure how that made it in there.

gdbelvin:
Updated. PTAL

PR #1237, opened 2019-03-07
Add metric for unapplied log entries

gdbelvin:
Questions:
- Doing a table scan of all the remaining items in the queue will get more and more expensive over time if a sequencer is stuck. How much of a risk is this? Maybe it's not too bad? Are there better ways of counting the number of items in the queue? 
- Setting a gauge of the number of remaining items in the queue might be difficult to aggregate over multiple servers - especially considering that only the master for a particular domain will be updating this metric.


pphaneuf:
Not sure if this place in the code is the best one to have a comprehensive count of the queue? Something on a recurring timer instead?

One way that I've found to make this sort of backlog metric more palatable is to cap the number of entries that will be counted in one go, so that you can tell if there's just a reasonably short backlog, or if there's *lots*.

If you're just trying to see your utilisation or how overloaded it is, another way might be to have a metric to indicate if the maximum batch size was attained in each run. This can also be combined with the previous, to have a separate marker to indicate that the gauge is hitting the red line.

gdbelvin:
Updated PTAL

Martin2112:
This may perform badly on cloudspanner etc. if it's getting row counts. The unsequenced count we did for the log didn't work well on mysql either. Pierre's idea sounds good.

PR #1241, opened 2019-03-13
Add metrics for each stage of the sequencing pipeline

gdbelvin:
This PR adds metrics at each stage of the sequencing pipeline to help identify issues when things get stuck in the middle of the pipeline.

### DIFF
--- a/core/sequencer/runner/native_test.go
+++ b/core/sequencer/runner/native_test.go
@@ -23,14 +23,13 @@ import (
 	pb "github.com/google/keytransparency/core/api/v1/keytransparency_go_proto"
 )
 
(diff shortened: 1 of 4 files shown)


core/sequencer/runner/native_test.go
```diff
@@ -23,6 +23,8 @@ import (
 	pb "github.com/google/keytransparency/core/api/v1/keytransparency_go_proto"
 )
 
+func fakeMetricFn(label string) {}
```
mhutchinson: Consider making this fake backed by a map. It would be a nice testing touch to assert that the metrics you will use for monitoring are actually being incremented as you expect.

PR #1244, opened 2019-03-20
Reduce mutationErr cardinality

gdbelvin:
Because error messsages can include arbitrary strings, using them as the index in a field counter resulted in a metric
that quickly became unusable due to its extremely large vector size.  

This PR reduces the cardinality by using error codes as the field to increment in ErrFn.
Doing so require creating a custom error code space and status object. 
Fortunately, this object will auto convert into a gRPC status object when referenced by gRPC.


core/sequencer/codes/code.go
```diff
@@ -0,0 +1,47 @@
...
+// limitations under the License.
+
+package codes
+
+import "google.golang.org/grpc/codes"
+
+// Code is an enum that represents a kind of error.
+type Code int
+
+const (
```
therealdrake: You don't need these to actually be error codes. Using an enum or fixed string would work just as well. This seems like it will be setting us up for needing to maintain an error code for every possible error which I don't believe is desirable.

core/sequencer/server.go
```diff
@@ -34,6 +34,7 @@ import (
 	"github.com/google/keytransparency/core/sequencer/runner"
 
 	spb "github.com/google/keytransparency/core/sequencer/sequencer_go_proto"
+	ktstatus "github.com/google/keytransparency/core/sequencer/status"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

gdbelvin:
@therealdrake Thanks for the advice on this.  I'm closing this PR in favor of a more canonical error code approach: #1248

PR #1248, opened 2019-03-25
Add Canonical Err Codes

gdbelvin:
Add canonical error codes to the errors that may occur during sequencing.

The main purpose of this is to expose a metric that can export the numbers of sequencing errors that are:
- InvalidArgument - various kinds of data corruption or user error.
- Internal - conditions that we expect to never happen.

### DIFF
--- a/core/mutator/entry/mutator.go
+++ b/core/mutator/entry/mutator.go
@@ -90,7 +90,7 @@ func ReduceFn(leaves []*pb.EntryUpdate, msgs []*pb.EntryUpdate,
 		newValue, err := MutateFn(oldValue, msg.GetMutation())
 		if err != nil {
 			s := status.Convert(err)
-			emitErr(status.Errorf(s.Code(), "entry: ReduceFn(msg %d/%d): %v", i, len(msgs)-1, s.Message()))
+			emitErr(status.Errorf(s.Code(), "entry: ReduceFn(msg %d/%d): %v", i+1, len(msgs), s.Message()))
 			continue
 		}
 		newEntries = append(newEntries, &pb.EntryUpdate{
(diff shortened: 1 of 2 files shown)


core/sequencer/runner/native.go
```diff
@@ -166,7 +174,7 @@ func DoMarshalIndexedValues(ivs []*entry.IndexedValue, emitErr func(error), incF
 		incFn("MarshalIndexedValue")
 		mapLeaf, err := iv.Marshal()
 		if err != nil {
-			emitErr(err)
+			emitErr(status.Errorf(codes.Internal, "MarshalIndexedValue(): %v", err))
```
golangcibot: undeclared name: `codes` (from `typecheck`)

core/mutator/entry/mutator.go
```diff
@@ -88,7 +89,8 @@ func ReduceFn(leaves []*pb.EntryUpdate, msgs []*pb.EntryUpdate,
 	for i, msg := range msgs {
 		newValue, err := MutateFn(oldValue, msg.GetMutation())
 		if err != nil {
-			emitErr(fmt.Errorf("entry: ReduceFn(msg %d/%d): %v", i, len(msgs)-1, err))
+			s := status.Convert(err)
+			emitErr(status.Errorf(s.Code(), "entry: ReduceFn(msg %d/%d): %v", i, len(msgs)-1, s.Message()))
```
pavelkalinnikov: Should it be `len(msgs)` without `-1`?

core/sequencer/runner/native.go
pavelkalinnikov: Please fix.

core/mutator/entry/mutator.go
gdbelvin: i iterates from 0 to n-1, so this either needs to be n-1 or i+1.  Perhaps +1 is better.

gdbelvin:
PTAL

core/mutator/entry/mutator.go
pavelkalinnikov: Hm, don't know. Still confusing a bit, but I think any syntax of this kind if ambiguous. Leaving this up to you.

PR #1250, opened 2019-03-27
{List,Create,Change}InputLogs Admin APIs

### LINKED ISSUE #1063: Create tool for adding / removing queue shards at runtime
The key server supports writing to multiple shards for horizontal scaling of incoming requests. 

This is currently set to a static number during initial domain creation.

We should create an API so an administrator can turn shards up/down as they please.

gdbelvin:
Support admin control of input log horizontal scaling.

Fixes #1063


core/adminserver/admin_server.go
```diff
@@ -83,8 +83,14 @@ var (
 type LogsAdmin interface {
 	// AddLogs creates and adds new logs for writing to a directory.
 	AddLogs(ctx context.Context, directoryID string, logIDs ...int64) error
+	// SetWritable enables or disables new writes from going to logID.
+	SetWritable(ctx context.Context, directoryID string, logID int64, enabled bool) error
+	// ListLogs returns a list of logs, optionally filtered by the writable bit.
+	ListLogs(ctx context.Context, directoryID string, writable bool) ([]int64, error)
 }
 
+var _ pb.KeyTransparencyAdminServer = &Server{}
```
therealdrake: what is this line doing?

core/adminserver/admin_server.go
```diff
@@ -213,11 +219,13 @@ func (s *Server) CreateDirectory(ctx context.Context, in *pb.CreateDirectoryRequ
 	// Generate VRF key.
 	wrapped, err := privKeyOrGen(ctx, in.GetVrfPrivateKey(), s.keygen)
 	if err != nil {
```
therealdrake: could this just be:
if code := status.Code(err); code != codes.OK {
  return nil, status.Errorf(code, "adminserver: keygen(): %v", err)
}

core/adminserver/admin_server.go
```diff
@@ -370,6 +389,39 @@ func (s *Server) UndeleteDirectory(ctx context.Context, in *pb.UndeleteDirectory
...
 }
 
+// ListInputLogs returns a list of input logs for a directory.
+func (s *Server) ListInputLogs(ctx context.Context, in *pb.ListInputLogsRequest) (*pb.ListInputLogsResponse, error) {
+	logIDs, err := s.logsAdmin.ListLogs(ctx, in.GetDirectoryId(), in.GetFilterWritable())
+	if err != nil {
+		s := status.Convert(err)
+		return nil, status.Errorf(s.Code(), "adminserver: ListLogs(): %v", s.Message())
+	}
+	inputLogs := make([]*pb.InputLog, 0, len(logIDs))
```
therealdrake: I don't know how much better this is than:
inputLogs := []*pb.InputLog{}
which I find more readable

core/adminserver/admin_server.go
```diff
@@ -370,6 +389,39 @@ func (s *Server) UndeleteDirectory(ctx context.Context, in *pb.UndeleteDirectory
...
+	for _, logID := range logIDs {
+		inputLogs = append(inputLogs, &pb.InputLog{LogId: logID, Writable: true})
+	}
+
+	return &pb.ListInputLogsResponse{Logs: inputLogs}, nil
+}
+
+// CreateInputLog returns a the created log.
```
therealdrake: "a" or "the"

impl/sql/mutationstorage/queue_test.go
```diff
@@ -40,6 +40,52 @@ func newForTest(ctx context.Context, t testing.TB, logIDs ...int64) *Mutations {
...
+			}
+
+			for logID, enabled := range tc.set {
+				if err := m.SetWritable(ctx, directoryID, logID, enabled); err != nil {
+					t.Errorf("SetWritable(): %v", err)
+				}
+			}
+
+			logs := make(map[int64]bool)
+			// Run enough times to ensure that random sampling will get us all logs.
```
therealdrake: could you instead set all logs not-writable except for one and ensure that is the one that gets written to?

core/adminserver/admin_server.go
gdbelvin: This line ensures that `&Server{}` implements the `pb.KeyTransparencyAdminServer` interface.

gdbelvin: I'll add a comment to that effect.

core/adminserver/admin_server.go
gdbelvin: hmm, that would be better...

core/adminserver/admin_server.go
gdbelvin: The only difference is that the way it's written pre-allocates memory so we can avoid multiple runtime array resizing.

impl/sql/mutationstorage/queue_test.go
gdbelvin: Sure

core/adminserver/admin_server.go
therealdrake: ah that's clever ok

gdbelvin:
Updated. PTAL

core/adminserver/admin_server.go
```diff
@@ -404,19 +390,17 @@ func (s *Server) ListInputLogs(ctx context.Context, in *pb.ListInputLogsRequest)
...
 		return nil, status.Errorf(s.Code(), "adminserver: AddLogs(%+v): %v", in.GetLogId(), s.Message())
 	}
 	return &pb.InputLog{LogId: in.GetLogId(), Writable: true}, nil
 }
 
 // UpdateInputLog updates the write bit for an input log.
 func (s *Server) UpdateInputLog(ctx context.Context, in *pb.InputLog) (*pb.InputLog, error) {
-	if err := s.logsAdmin.SetWritable(ctx, in.GetDirectoryId(), in.GetLogId(), in.GetWritable()); err != nil {
-		s := status.Convert(err)
+	if s := status.Convert(s.logsAdmin.SetWritable(ctx, in.GetDirectoryId(), in.GetLogId(), in.GetWritable())); s.Code() != codes.OK {
```
therealdrake: I feel like the call to SetWritable should go on the line above

core/adminserver/admin_server.go
```diff
@@ -404,19 +390,17 @@ func (s *Server) ListInputLogs(ctx context.Context, in *pb.ListInputLogsRequest)
 	return &pb.ListInputLogsResponse{Logs: inputLogs}, nil
 }
 
-// CreateInputLog returns a the created log.
+// CreateInputLog returns the created log.
 func (s *Server) CreateInputLog(ctx context.Context, in *pb.InputLog) (*pb.InputLog, error) {
-	if err := s.logsAdmin.AddLogs(ctx, in.GetDirectoryId(), in.GetLogId()); err != nil {
-		s := status.Convert(err)
+	if s := status.Convert(s.logsAdmin.AddLogs(ctx, in.GetDirectoryId(), in.GetLogId())); s.Code() != codes.OK {
```
therealdrake: I feel like the call to AddLogs should go on the line above

PR #1259, opened 2019-04-17
Pass through grpc error codes

gdbelvin:
Many codes.Canceled errors are being converted into codes.Internal.
As a result, monitoring is getting overly excited.

Solution: keep the original error codes.

### DIFF
--- a/core/keyserver/keyserver.go
+++ b/core/keyserver/keyserver.go
@@ -189,13 +189,13 @@ func (s *Server) batchGetUserByRevision(ctx context.Context, sth *tpb.SignedLogR
 		Index:    indexes,
 		Revision: mapRevision,
 	})
-	if err != nil {
+	if st := status.Convert(err); st.Code() != codes.OK {
 		glog.Errorf("GetLeavesByRevision(%v, rev: %v): %v", d.Map.TreeId, mapRevision, err)
-		return nil, status.Errorf(codes.Internal, "Failed fetching map leaf")
+		return nil, status.Errorf(st.Code(), "Failed fetching map leaf")
 	}
 	if got, want := len(getResp.MapLeafInclusion), len(userIDs); got != want {
 		glog.Errorf("GetLeavesByRevision() len: %v, want %v", got, want)
-		return nil, status.Errorf(codes.Internal, "Failed fetching map leaf")
+		return nil, status.Errorf(codes.Internal, "Wrong number of map leaves returned")
 	}
 	leaves := make(map[string]*pb.MapLeaf)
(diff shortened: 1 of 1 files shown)


core/keyserver/keyserver.go
```diff
@@ -189,13 +189,13 @@ func (s *Server) batchGetUserByRevision(ctx context.Context, sth *tpb.SignedLogR
 		Index:    indexes,
 		Revision: mapRevision,
 	})
-	if err != nil {
+	if st := status.Convert(err); st.Code() != codes.OK {
```
jtoohill: Is this preferred over checking err != nil? status upholds that the code is not OK iff err is non-nil.

core/keyserver/keyserver.go
```diff
@@ -189,13 +189,13 @@ func (s *Server) batchGetUserByRevision(ctx context.Context, sth *tpb.SignedLogR
...
 		glog.Errorf("GetLeavesByRevision(%v, rev: %v): %v", d.Map.TreeId, mapRevision, err)
-		return nil, status.Errorf(codes.Internal, "Failed fetching map leaf")
+		return nil, status.Errorf(st.Code(), "Failed fetching map leaf")
```
jtoohill: Depending on whether you check err != nil above, you could write a helper which takes an error and returns that error annotated with a string, passing along the existing status code if it was set.

core/keyserver/keyserver.go
gdbelvin: It works regardless of whether err is a Status error or nil. 
Creating it in this way saves us a line of code and scopes `st` to the error block too. 
Not sure what the 100% canonical pattern for doing this is.

core/keyserver/keyserver.go
gdbelvin: Indeed.  But that feels a little like it should really be part of `status`.
A true solution for this kind of thing will be coming in Go 2.0

core/keyserver/keyserver.go
```diff
@@ -189,13 +189,13 @@ func (s *Server) batchGetUserByRevision(ctx context.Context, sth *tpb.SignedLogR
...
 	}
 	if got, want := len(getResp.MapLeafInclusion), len(userIDs); got != want {
 		glog.Errorf("GetLeavesByRevision() len: %v, want %v", got, want)
-		return nil, status.Errorf(codes.Internal, "Failed fetching map leaf")
```
therealdrake: does this case indicate some error in the request? or a server error?

core/keyserver/keyserver.go
therealdrake: I think Jon is maybe suggesting passing the error up the stack?

PR #1262, opened 2019-04-19
Limit the amount of catch-up work done in RunBatch

therealdrake:
To keep from timing out in the case that the sequencer falls behind in its job. Limit how much catchup we attempt in any given RunBatch call.

### DIFF
--- a/core/sequencer/server.go
+++ b/core/sequencer/server.go
@@ -46,18 +46,20 @@ const (
 )
 
 var (
-	initMetrics       sync.Once
-	knownDirectories  monitoring.Gauge
-	logEntryCount     monitoring.Counter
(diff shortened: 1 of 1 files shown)


core/sequencer/server.go
```diff
@@ -231,16 +243,24 @@ func (s *Server) RunBatch(ctx context.Context, in *spb.RunBatchRequest) (*empty.
 		return nil, err
 	}
 
+	unappliedRevisions.Set(len(defResp.OutstandingRevisions))
```
golangcibot: cannot use len(defResp.OutstandingRevisions) (value of type int) as float64 value in argument to unappliedRevisions.Set (from `typecheck`)

PR #1263, opened 2019-04-19
 Separate Log Revision Writes from Map Writes

therealdrake:
Moved PublishRevision out of RunBatch and created its own function to call periodically. Publishing to the log and writing to the map are independent actions.

### DIFF
--- a/cmd/keytransparency-sequencer/main.go
+++ b/cmd/keytransparency-sequencer/main.go
@@ -220,6 +220,9 @@ func runSequencer(ctx context.Context, conn *grpc.ClientConn,
 		}
 	})
 
+	if err := signer.AddAllDirectories(ctx); err != nil {
+		glog.Errorf("runSequencer(AddAllDirectories): %v", err)
+	}
 	go sequencer.PeriodicallyRun(ctx, time.Tick(*refresh), func(ctx context.Context) {
 		if err := signer.AddAllDirectories(ctx); err != nil {
 			glog.Errorf("PeriodicallyRun(AddAllDirectories): %v", err)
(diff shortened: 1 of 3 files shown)


cmd/keytransparency-sequencer/main.go
```diff
@@ -228,4 +228,13 @@ func runSequencer(ctx context.Context, conn *grpc.ClientConn,
 			glog.Errorf("PeriodicallyRun(RunBatchForAllMasterships): %v", err)
 		}
 	})
+
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

core/sequencer/sequencer.go
```diff
@@ -120,3 +120,24 @@ func (s *Sequencer) RunBatchForAllMasterships(ctx context.Context, batchSize int
...
+	glog.Infof("PublishLogForAllMasterships")
+	cctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	masterships, err := s.tracker.Masterships(cctx)
+	if err != nil {
+		return err
+	}
+
+	var lastErr error
+	for dirID, whileMaster := range masterships {
```
golangcibot: ineffectual assignment to `whileMaster` (from `ineffassign`)

cmd/keytransparency-sequencer/main.go
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

cmd/keytransparency-sequencer/main.go
```diff
@@ -228,4 +228,13 @@ func runSequencer(ctx context.Context, conn *grpc.ClientConn,
...
+	sequencer.PeriodicallyRun(ctx, time.Tick(*refresh), func(ctx context.Context) {
```
jtoohill: Need to use a goroutine above, since sequencer.PeriodicallyRun runs in a loop until context cancellation.

cmd/keytransparency-sequencer/main.go
```diff
@@ -228,4 +228,13 @@ func runSequencer(ctx context.Context, conn *grpc.ClientConn,
...
+		if err := signer.AddAllDirectories(ctx); err != nil {
```
jtoohill: I think you have a data race here, since both periodic jobs are trying to add directories.

PR #1268, opened 2019-04-24
Fix broken test in monitor_tests.go

therealdrake:
Need to call PublishRevisions explicitly now that it has been separated from RunBatch.


core/integration/monitor_tests.go
```diff
@@ -108,13 +108,20 @@ func TestMonitor(ctx context.Context, env *Env, t *testing.T) []testdata.Respons
 				t.Errorf("QueueMutation(): %v", err)
 			}
 		}
-		if _, err := env.Sequencer.RunBatch(ctx, &spb.RunBatchRequest{
+		batchReq :=  &spb.RunBatchRequest{
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

PR #1270, opened 2019-04-24
Add new metric to count revisions written to the queue

therealdrake:
Add new metric to count revisions written to the sequencer queue.

### DIFF
--- a/core/keyserver/keyserver.go
+++ b/core/keyserver/keyserver.go
@@ -47,12 +47,17 @@ const (
 var (
 	initMetrics      sync.Once
 	watermarkWritten monitoring.Gauge
+	sequencerQueueWritten monitoring.Counter
 )
 
 func createMetrics(mf monitoring.MetricFactory) {
 	watermarkWritten = mf.NewGauge(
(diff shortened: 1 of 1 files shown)


core/keyserver/keyserver.go
```diff
@@ -591,6 +596,7 @@ func (s *Server) BatchQueueUserUpdate(ctx context.Context, in *pb.BatchQueueUser
 	}
 	if wm != nil {
 		watermarkWritten.Set(float64(wm.Watermark), directory.DirectoryID, fmt.Sprintf("%v", wm.LogID))
+		sequencerQueueWritten.Add(float64(len(in.Updates)), directory.DirectoryID, fmt.Sprintf("%v", wm.LogID))
```
mhutchinson: Optional: it may be useful to include another counter to capture the number of updates this method receives regardless of whether they are successfully written. The ratio of these counters would tell you how large an impact any errors are causing. Suggest keyserver_queue_received.

therealdrake: I'll do that in a followup. Thanks!

PR #1278, opened 2019-05-15
Parallize functions in the BatchCreateUser flow

gdbelvin:
This will help us create more fake traffic for load testing.

Profiling these functions has revealed a significant amount of time being spent in running the VRF functions.

### DIFF
--- a/core/client/batch_client.go
+++ b/core/client/batch_client.go
@@ -108,7 +108,7 @@ func (c *Client) BatchCreateMutation(ctx context.Context, users []*User) ([]*ent
 		defer close(rChan)
 		var wg sync.WaitGroup
 		defer wg.Wait() // Wait before closing rChan
-		for w := 1; w < runtime.NumCPU(); w++ {
+		for w := 0; w < runtime.NumCPU(); w++ {
 			wg.Add(1)
 			go func(uChan <-chan *User, rChan chan<- result) {
 				defer wg.Done()
--- a/core/client/batch_client.go
+++ b/core/client/batch_client.go
@@ -109,13 +109,13 @@ func (c *Client) BatchCreateMutation(ctx context.Context, users []*User) ([]*ent
 		var wg sync.WaitGroup
 		for w := 1; w < (runtime.NumCPU() - 1); w++ {
 			wg.Add(1)
-			go func(id int, uChan <-chan *User, rChan chan<- result) {
+			go func(uChan <-chan *User, rChan chan<- result) {
 				defer wg.Done()
 				for u := range uChan {
 					m, err := c.createMutation(u, leavesByUserID[u.UserID])
 					rChan <- result{m: m, err: err}
 				}
-			}(w, uChan, rChan)
+			}(uChan, rChan)
 		}
 		wg.Wait()
 	}()
--- a/core/client/batch_client.go
+++ b/core/client/batch_client.go
@@ -107,6 +107,7 @@ func (c *Client) BatchCreateMutation(ctx context.Context, users []*User) ([]*ent
 	go func() {
 		defer close(rChan)
 		var wg sync.WaitGroup
+		defer wg.Wait() // Wait before closing rChan
 		for w := 1; w < runtime.NumCPU(); w++ {
 			wg.Add(1)
 			go func(uChan <-chan *User, rChan chan<- result) {
@@ -117,7 +118,6 @@ func (c *Client) BatchCreateMutation(ctx context.Context, users []*User) ([]*ent
 				}
 			}(uChan, rChan)
 		}
-		wg.Wait()
 	}()
 	// Collect
 	mutations := make([]*entry.Mutation, 0, len(users))
--- a/core/client/batch_client.go
+++ b/core/client/batch_client.go
@@ -16,6 +16,8 @@ package client
 
 import (
 	"context"
+	"runtime"
+	"sync"
 
 	"fmt"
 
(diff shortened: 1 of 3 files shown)


core/client/batch_client.go
```diff
@@ -86,34 +88,71 @@ func (c *Client) BatchCreateMutation(ctx context.Context, users []*User) ([]*ent
...
-		if err != nil {
-			return nil, err
+	}()
+	// Workerpool
+	go func() {
+		defer close(rChan)
+		var wg sync.WaitGroup
+		for w := 1; w < (runtime.NumCPU() - 1); w++ {
+			wg.Add(1)
+			go func(id int, uChan <-chan *User, rChan chan<- result) {
```
golangcibot: `(*Client).BatchCreateMutation$2$1` - `id` is unused (from `unparam`)

core/keyserver/keyserver.go
```diff
@@ -580,15 +609,37 @@ func (s *Server) BatchQueueUserUpdate(ctx context.Context, in *pb.BatchQueueUser
 	// - Correct profile commitment.
 	// - Correct key formats.
 	_, tdone := monitoring.StartSpan(ctx, "BatchQueueUserUpdate.verify")
-	for _, u := range in.Updates {
-		if err := s.verifyMutation(u.Mutation); err != nil {
-			glog.Warningf("Invalid UpdateEntryRequest: %v", err)
-			return nil, status.Errorf(codes.InvalidArgument, "Invalid mutation")
+	updates := make(chan *pb.EntryUpdate)
+	results := make(chan error)
```
AlCutter: /results/errors/?

core/client/batch_client.go
```diff
@@ -86,34 +88,71 @@ func (c *Client) BatchCreateMutation(ctx context.Context, users []*User) ([]*ent
...
 		}
-		index, err := c.Index(leaf.GetVrfProof(), c.DirectoryID, u.UserID)
-		if err != nil {
-			return nil, err
+	}()
+	// Workerpool
+	go func() {
+		defer close(rChan)
+		var wg sync.WaitGroup
+		for w := 1; w < (runtime.NumCPU() - 1); w++ {
```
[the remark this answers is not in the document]
gdbelvin: I think this is why Travis is failing...

core/client/batch_client.go
```diff
@@ -107,7 +107,7 @@ func (c *Client) BatchCreateMutation(ctx context.Context, users []*User) ([]*ent
 	go func() {
 		defer close(rChan)
 		var wg sync.WaitGroup
-		for w := 1; w < (runtime.NumCPU() - 1); w++ {
+		for w := 1; w < runtime.NumCPU(); w++ {
```
AlCutter: This has exactly the same problem if `NumCPU() == 1`, doesn't it?

core/client/batch_client.go
```diff
@@ -86,34 +88,71 @@ func (c *Client) BatchCreateMutation(ctx context.Context, users []*User) ([]*ent
...
+					rChan <- result{m: m, err: err}
+				}
+			}(uChan, rChan)
 		}
-		mutation := entry.NewMutation(index, c.DirectoryID, u.UserID)
-
-		leafValue := leaf.MapInclusion.GetLeaf().GetLeafValue()
-		if err := mutation.SetPrevious(leafValue, true); err != nil {
-			return nil, err
+		wg.Wait()
```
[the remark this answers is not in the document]
gdbelvin: How about putting the two deferreds close together?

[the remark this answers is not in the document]
AlCutter: :+1: take care with ordering of `defer`s, though :)

core/client/batch_client.go
```diff
@@ -86,34 +88,71 @@ func (c *Client) BatchCreateMutation(ctx context.Context, users []*User) ([]*ent
...
 
-	for _, u := range users {
-		leaf, ok := leavesByUserID[u.UserID]
-		if !ok {
-			return nil, fmt.Errorf("no leaf found for %v", u.UserID)
+	type result struct {
+		m   *entry.Mutation
+		err error
+	}
+	uChan := make(chan *User)
```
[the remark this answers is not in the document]
gdbelvin: I don't see any particular reason a buffer would help.  There should be N readers executing at any time, and we're not trying to avoid blocking of the allocation function which is in it's own thread.

core/client/batch_client.go
gdbelvin: It does. The previous logic was effectively `NumCPU - 2`.  It's now `NumCPU -1`.
Would you recommend straight `NumCPU` or a fancy function to determine the parallelism?

core/client/batch_client.go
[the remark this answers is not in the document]
AlCutter: You'll generally get better performance because you increase the potential for concurrency:

code: https://play.golang.org/p/b-yXwNX2VDx
```
BenchmarkNoBuffer-6           	 2000000	       809 ns/op
BenchmarkWithBufferOf10-6     	 3000000	       518 ns/op
BenchmarkWithBufferof100-6    	 5000000	       286 ns/op
BenchmarkWithBufferof1000-6   	 5000000	       297 ns/op
```

Might not be super important in this context, but it's worth being aware of, and I think there should be a reason _not_ to buffer, rather than the other way around.

core/client/batch_client.go
AlCutter: `w := 0` should fix the general problem, and then maybe you could do `w < max(1, runtime.NumCPU()-2)` if you wanted to be fancy -- although consider how intensive the `allocate` vs `work` functions are in the different use cases here - if the cost of doing the work is vastly greater than the `allocate` (e.g. signature/mutation verification), then it's probably better to just have a worker goroutine per core and let the scheduler steal slivers of time for the `allocate` bit.

gdbelvin: `allocate` work should be super minimal, happy to allow the scheduler to handle it. I was more worried about starving the entire node of resources -- the node might be busy doing other things too.

gdbelvin: OK, we'll use all the cpus :-)

PR #1282, opened 2019-05-17
Parallize VerifyBatchGetUserIndex

gdbelvin:
This produces a 7x speedup for batches of 1000 on a 12 core computer. 
Before: 15s
After: 2s

### DIFF
--- a/core/client/batch_get_and_verify.go
+++ b/core/client/batch_get_and_verify.go
@@ -16,8 +16,11 @@ package client
 
 import (
 	"context"
+	"runtime"
+	"sync"
 
 	pb "github.com/google/keytransparency/core/api/v1/keytransparency_go_proto"
+	"github.com/google/trillian/monitoring"
 )
 
 // BatchVerifyGetUserIndex fetches and verifies the indexes for a list of users.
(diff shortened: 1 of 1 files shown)


core/client/batch_get_and_verify.go
```diff
@@ -30,13 +33,57 @@ func (c *Client) BatchVerifyGetUserIndex(ctx context.Context, userIDs []string)
 		return nil, err
 	}
 
+	_, spanEnd := monitoring.StartSpan(ctx, "VerifyBatchGetUserIndex")
```
pavelkalinnikov: Did you intend this to match method name? OTOH, it doesn't span the whole method, so maybe names should be even more different.

core/client/batch_get_and_verify.go
```diff
@@ -30,13 +33,57 @@ func (c *Client) BatchVerifyGetUserIndex(ctx context.Context, userIDs []string)
...
+	defer spanEnd()
+
+	type proof struct {
+		userID string
+		proof  []byte
+	}
+	proofs := make(chan proof)
+	done := make(chan struct{})
+	defer close(done)
+	go func() {
```
pavelkalinnikov: How about some commenting next to each of the goroutines?

core/client/batch_get_and_verify.go
```diff
@@ -30,13 +33,57 @@ func (c *Client) BatchVerifyGetUserIndex(ctx context.Context, userIDs []string)
...
+		defer close(proofs)
+		for UID, p := range resp.GetProofs() {
+			select {
+			case proofs <- proof{UID, p}:
```
pavelkalinnikov: `proof{userID: UID, proof: p}`

core/client/batch_get_and_verify.go
```diff
@@ -30,13 +33,57 @@ func (c *Client) BatchVerifyGetUserIndex(ctx context.Context, userIDs []string)
...
+			}
+		}
+	}()
+	type result struct {
+		userID string
+		index  []byte
+		err    error
+	}
+	results := make(chan result)
+	go func() {
```
pavelkalinnikov: optional: Could make a little less nesting here as follows:

```golang
var wg sync.WaitGroup
go func() {
	wg.Wait()
	close(results)
}()
for w := 0; w < runtime.NumCPU(); w++ {
	wg.Add(1)
	go func() {
		defer wg.Done()
		...
	}()
}
```

core/client/batch_get_and_verify.go
gdbelvin: Why not both? `BatchVerifyGetUserIndex.Verify`

core/client/batch_get_and_verify.go
gdbelvin: Thanks!

PR #1283, opened 2019-05-17
Parallize BatchVerifyGetUser

gdbelvin:
`go test ./impl/integration/ -run TestIntegration/TestBatchUpdate/100`
github.com/google/keytransparency/impl/integration	2.927s Before
github.com/google/keytransparency/impl/integration	2.341s  After

### DIFF
--- a/core/client/batch_get_and_verify.go
+++ b/core/client/batch_get_and_verify.go
@@ -60,12 +60,10 @@ func (c *Client) BatchVerifiedGetUser(ctx context.Context, userIDs []string) (ma
 	}
 	c.updateTrusted(slr)
 
-	leavesByUserID := make(map[string]*pb.MapLeaf)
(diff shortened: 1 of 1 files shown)


pavelkalinnikov:
Do you measure impact of these parallelizations? I wonder if there are any hidden pitfalls in making everything parallel. A `Benchmark` test (for individual pipeline components and/or the whole thing)  would at least give some idea.

gdbelvin:
Everything probably doesn't need to be parallel. However, the parallel verifies do make a difference because of the CPU heavy nature of VRF verification.

gdbelvin:
I added a benchmark for the VRF function #1284.

I also timed the following test
go test ./impl/integration/ -run TestIntegration/TestBatchUpdate/100

gdbelvin:
Deprecated by #1328

PR #1285, opened 2019-05-24
Experiment with Go Modules

gdbelvin:


.travis.yml
```diff
@@ -30,11 +34,11 @@ before_install:
 install:
   - gcloud -q components install kubectl
   - curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(go env GOPATH)/bin v1.16.0
-  - go get -t ./...
+  - if [ "$GO111MODULE" = "auto" ]; then go get -t -v ./... ; fi
 
 script:
   - go build ./...
-  - golangci-lint run
+  - golangci-lint run --deadline=5m
```
RJPercival: Is the timeout necessary both here and in .golangci.yml?

go.mod
```diff
@@ -0,0 +1,33 @@
...
+go 1.12
+
+require (
+	github.com/benlaurie/objecthash v0.0.0-20180202135721-d1e3d6079fc1
+	github.com/go-sql-driver/mysql v1.4.1 // indirect
+	github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b
+	github.com/golang/mock v1.3.1
+	github.com/golang/protobuf v1.3.1
+	github.com/google/btree v1.0.0 // indirect
+	github.com/google/certificate-transparency-go v1.0.21 // indirect
```
RJPercival: This is [a fairly old release of certificate-transparency-go](https://github.com/google/certificate-transparency-go/releases/tag/v1.0.21) (2018-08-20) compared to the version of Trillian being used (from 2019-05-24). You're certain this won't cause any problems?

gdbelvin: It is old... but it is also the latest tag.  Should we create a new release of ct-go?

gdbelvin: It's not causing any issues the I am aware of

.travis.yml
gdbelvin: Testing showed that the command didn't pick up the one in golangci.yml

RJPercival: Strange. What effect does the setting in golangci.yml have? If none, we should probably remove it.

go.mod
RJPercival: It's most likely fine, but this combination of versions is unlikely to have been tested beyond the unit/integration tests (compared to picking commits/releases that are closer together temporally, which will be more similar to versions that went through CI and into production together).

PR #1286, opened 2019-05-24
Use new trillian.GetLeaf API

gdbelvin:
Avoid using batch APIs where the single shot API would work.
This will help clarify server performance by keeping batch and single user traffic separate.

### DIFF
--- a/core/keyserver/keyserver.go
+++ b/core/keyserver/keyserver.go
@@ -193,16 +193,33 @@ func (s *Server) batchGetUserByRevision(ctx context.Context, sth *tpb.SignedLogR
 	for index := range usersByIndex {
 		indexes = append(indexes, []byte(index))
 	}
-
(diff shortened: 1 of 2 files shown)


core/keyserver/keyserver.go
```diff
@@ -193,16 +193,33 @@ func (s *Server) batchGetUserByRevision(ctx context.Context, sth *tpb.SignedLogR
...
+		}
+	} else {
+		getResp, err = s.tmap.GetLeavesByRevision(ctx, &tpb.GetMapLeavesByRevisionRequest{
+			MapId:    d.Map.TreeId,
+			Index:    indexes,
+			Revision: mapRevision,
+		})
+		if st := status.Convert(err); st.Code() != codes.OK {
+			glog.Errorf("GetLeavesByRevision(%v, rev: %v): %v", d.Map.TreeId, mapRevision, err)
+			return nil, status.Errorf(st.Code(), "Failed fetching map leaf")
```
pavelkalinnikov: map leaves

PR #1287, opened 2019-05-24
Refactor ForAllMasterships

gdbelvin:
Factor out ForAllMasterships so it can be used for other functions

### DIFF
--- a/core/sequencer/sequencer.go
+++ b/core/sequencer/sequencer.go
@@ -95,50 +95,51 @@ func (s *Sequencer) AddDirectory(dirIDs ...string) {
 	}
 }
 
-// RunBatchForAllMasterships runs RunBatch on all directires this sequencer is currently master for.
-func (s *Sequencer) RunBatchForAllMasterships(ctx context.Context, batchSize int32) error {
-	glog.Infof("RunBatchForAllMasterships")
+// ForAllMasterships runs f once for all directories this server is master for.
+func (s *Sequencer) ForAllMasterships(ctx context.Context, f func(ctx context.Context, dirID string) error) error {
 	cctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 	masterships, err := s.tracker.Masterships(cctx)
 	if err != nil {
 		return err
 	}
-
 	var lastErr error
-	for dirID, whileMaster := range masterships {
+	for dirID, mastershipCtx := range masterships {
+		if err := f(mastershipCtx, dirID); err != nil {
+			lastErr = err
+		}
+	}
+	return lastErr
+}
+
+// RunBatchForAllMasterships runs RunBatch on all directires this sequencer is currently master for.
+func (s *Sequencer) RunBatchForAllMasterships(ctx context.Context, batchSize int32) error {
+	glog.Infof("RunBatchForAllMasterships")
+	return s.ForAllMasterships(ctx, func(ctx context.Context, dirID string) error {
 		req := &spb.RunBatchRequest{
 			DirectoryId: dirID,
 			MinBatch:    1,
 			MaxBatch:    batchSize,
 		}
-		if _, err := s.sequencerClient.RunBatch(whileMaster, req); err != nil {
-			lastErr = err
+		_, err := s.sequencerClient.RunBatch(ctx, req)
+		if err != nil {
 			glog.Errorf("RunBatch for %v failed: %v", dirID, err)
+			return err
 		}
-	}
-
-	return lastErr
+		return nil
+	})
 }
 
 // PublishLogForAllMasterships run PublishRevisions on all directories this sequencer is currently master for.
 func (s *Sequencer) PublishLogForAllMasterships(ctx context.Context) error {
 	glog.Infof("PublishLogForAllMasterships")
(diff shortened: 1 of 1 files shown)


RJPercival:
nit: It looks like the function you actually refactored is called `RunBatchForAllMasterships`, not `ForAllMasterships` (the latter is a new function you've created).

core/sequencer/sequencer.go
```diff
@@ -95,50 +95,51 @@ func (s *Sequencer) AddDirectory(dirIDs ...string) {
...
 	masterships, err := s.tracker.Masterships(cctx)
 	if err != nil {
 		return err
 	}
-
 	var lastErr error
-	for dirID, whileMaster := range masterships {
+	for dirID, mastershipCtx := range masterships {
+		if err := f(mastershipCtx, dirID); err != nil {
+			lastErr = err
```
RJPercival: Future improvement: It'd be nice if all errors except the last weren't discarded. Consider aggregating them. As is, the last error is likely to end up getting logged twice, potentially implying it occurred twice (since you already log all errors that occur in `f` on line 126).

core/sequencer/sequencer.go
```diff
@@ -95,50 +95,51 @@ func (s *Sequencer) AddDirectory(dirIDs ...string) {
...
+		}
+	}
+	return lastErr
+}
+
+// RunBatchForAllMasterships runs RunBatch on all directires this sequencer is currently master for.
```
RJPercival:
```suggestion
// RunBatchForAllMasterships runs RunBatch on all directories this sequencer is currently master for.
```

RJPercival: nit: Since `RunBatch` is a method on `KeyTransparencySequencerClient`, rather than `Sequencer`, I'm not sure it's obvious which method you're referring to in this comment. Perhaps it'd be helpful to readers if you either:
- stated that it runs `KeyTransparencySequencerClient.RunBatch`
- described what the result of running `RunBatch` will be, rather than just saying it'll run it, e.g.
```suggestion
// RunBatchForAllMasterships integrates a batch of mutations into each directory for which this sequencer is master.
```
Having said this, your current comment is consistent with the comments on other method in this file.

core/sequencer/sequencer.go
```diff
@@ -95,50 +95,51 @@ func (s *Sequencer) AddDirectory(dirIDs ...string) {
...
+			return err
 		}
-	}
-
-	return lastErr
+		return nil
+	})
 }
 
 // PublishLogForAllMasterships run PublishRevisions on all directories this sequencer is currently master for.
```
RJPercival:
```suggestion
// PublishLogForAllMasterships runs PublishRevisions on all directories this sequencer is currently master for.
```

core/sequencer/sequencer.go
```diff
@@ -95,50 +95,51 @@ func (s *Sequencer) AddDirectory(dirIDs ...string) {
...
-	}
-
-	var lastErr error
-	for dirID, whileMaster := range masterships {
+	return s.ForAllMasterships(ctx, func(ctx context.Context, dirID string) error {
 		publishReq := &spb.PublishRevisionsRequest{DirectoryId: dirID}
-		if _, err = s.sequencerClient.PublishRevisions(whileMaster, publishReq); err != nil {
-			lastErr = err
+		_, err := s.sequencerClient.PublishRevisions(ctx, publishReq)
+		if err != nil {
```
RJPercival: nit: Why is this no longer on one line?

RJPercival:
It'd be nice to have tests - you've got absolutely no test coverage of this change.

core/sequencer/sequencer.go
gdbelvin: done

core/sequencer/sequencer.go
gdbelvin: That was a half-way reverted optimization.

core/sequencer/sequencer.go
```diff
@@ -95,50 +97,59 @@ func (s *Sequencer) AddDirectory(dirIDs ...string) {
...
+		}
+		return errors.New(strings.Join(msgs, ", "))
+	}
+	return nil
+}
 
-	var lastErr error
-	for dirID, whileMaster := range masterships {
+// RunBatchForAllMasterships runs KeyTransparencySequencerClient.RunBatch on all
+// directires this sequencer is currently master for.
```
RJPercival:
```suggestion
// directories this sequencer is currently master for.
```

core/sequencer/sequencer.go
```diff
@@ -95,50 +97,59 @@ func (s *Sequencer) AddDirectory(dirIDs ...string) {
...
 		}
-	}
-
-	return lastErr
+		return nil
+	})
 }
 
-// PublishLogForAllMasterships run PublishRevisions on all directories this sequencer is currently master for.
+// PublishLogForAllMasterships run KeyTransparencySequencer.PublishRevisions on
```
RJPercival:
```suggestion
// PublishLogForAllMasterships runs KeyTransparencySequencer.PublishRevisions on
```

PR #1289, opened 2019-05-29
Move to travis-ci.com

gdbelvin:
Use the travis-ci GitHub Apps integration, remove the old services based integration

https://blog.travis-ci.com/2018-05-02-open-source-projects-on-travis-ci-com-with-github-apps

TravisCI.com uses a smaller permission set:
- https://docs.travis-ci.com/user/github-oauth-scopes/
- Screenshot 2019-05-29 at 15 50 42


Martin2112:
As discussed offline not sure we can move yet.

gdbelvin:
Yes, I needed to change the required check to TravisCI. 

I've confirmed the permissions in all the ways I know how.  Could you attach a screenshot of what you're seeing?

Martin2112:
Yes I don't think you can sign into travis-ci.com without giving it repo write access. Please double check that and if so you need to revoke it.

RJPercival:
If that's a screenshot of the permissions that the GitHub App is requesting, then this seems fine. It appears that logging into travis-ci.com requires a larger set of permissions for some reason but, so long as you don't need to do that, it should be fine.

gdbelvin:
It does indeed look like the OAuth permissions to view travis-ci.com are sepate 

are a separate thing from the GitHub App 

Here's the screenshot from logging in, and the expanded set of permissions is the "Full control of private repos" line. 

Screenshot 2019-05-30 at 14 20 50

According to both of the following links, GitHub OAuth apps do not support read-only access for source code access.  This is likely the source of the issue. 
- https://developer.github.com/apps/differences-between-apps/#requesting-permission-levels-for-resources
- https://help.github.com/en/articles/authorizing-oauth-apps#repositories

Note that OAuth permissions only apply to individual user accounts, not organizational repos.

gdbelvin:
The GitHub App integration is what allows the pull requests to be processed and tested.  This has been installed by the group admin.

The OAuth permission is what allows individual users to login to travis-ci.com

PR #1292, opened 2019-05-31
Use GetMapLeavesNoProof

gdbelvin:
It's not necessary to fully verify the results from the Trillian Map inside the sequencer because 
this effort is duplicated by clients.

Depends on google/trillian#1647

### DIFF
--- a/core/sequencer/server_test.go
+++ b/core/sequencer/server_test.go
@@ -126,7 +126,7 @@ type fakeMapConn struct {
 
 var errSuccess = status.Errorf(codes.Unimplemented, "Success! No Duplicates. Shortcut return")
 
-func (m *fakeMapConn) GetLeavesByRevision(_ context.Context, in *tpb.GetMapLeavesByRevisionRequest, _ ...grpc.CallOption) (*tpb.GetMapLeavesResponse, error) {
+func (m *fakeMapConn) GetLeavesByRevisionNoProof(_ context.Context, in *tpb.GetMapLeavesByRevisionRequest, _ ...grpc.CallOption) (*tpb.MapLeaves, error) {
 	set := make(map[string]bool)
 	for _, i := range in.Index {
 		if set[string(i)] {
--- a/core/sequencer/server_test.go
+++ b/core/sequencer/server_test.go
@@ -18,27 +18,18 @@ import (
 	"context"
 	"fmt"
 	"sort"
-	"strings"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
-	"github.com/google/tink/go/keyset"
-	"github.com/google/tink/go/signature"
-	"github.com/google/tink/go/tink"
 	"github.com/google/trillian/monitoring"
 	"github.com/google/trillian/types"
-	"google.golang.org/grpc"
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
 
 	"github.com/google/keytransparency/core/mutator"
-	"github.com/google/keytransparency/core/mutator/entry"
 	"github.com/google/keytransparency/core/sequencer/mapper"
 	"github.com/google/keytransparency/core/sequencer/runner"
 
 	spb "github.com/google/keytransparency/core/sequencer/sequencer_go_proto"
 	tpb "github.com/google/trillian"
-	tclient "github.com/google/trillian/client"
 )
 
 const directoryID = "directoryID"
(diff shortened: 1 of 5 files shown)


core/sequencer/trillian_client.go
```diff
@@ -141,3 +141,34 @@ func (c *MapClient) GetAndVerifyMapRootByRevision(ctx context.Context,
...
 	return rawMapRoot, mapRoot, nil
 }
+
+// GetMapLeavesByRevisionNoProof verifies and returns the requested map leaves at a specific revision.
+// indexes may not contain duplicates.
+func (c *MapClient) GetMapLeavesByRevisionNoProof(ctx context.Context, revision int64, indexes [][]byte) ([]*tpb.MapLeaf, error) {
+	if err := hasDuplicates(indexes); err != nil {
+		return nil, err
+	}
+	getResp, err := c.Conn.GetLeavesByRevisionNoProof(ctx, &tpb.GetMapLeavesByRevisionRequest{
```
golangcibot: c.Conn.GetLeavesByRevisionNoProof undefined (type trillian.TrillianMapClient has no field or method GetLeavesByRevisionNoProof) (from `typecheck`)

core/sequencer/server_test.go
```diff
@@ -126,7 +126,7 @@ type fakeMapConn struct {
 
 var errSuccess = status.Errorf(codes.Unimplemented, "Success! No Duplicates. Shortcut return")
 
-func (m *fakeMapConn) GetLeavesByRevision(_ context.Context, in *tpb.GetMapLeavesByRevisionRequest, _ ...grpc.CallOption) (*tpb.GetMapLeavesResponse, error) {
+func (m *fakeMapConn) GetLeavesByRevisionNoProof(_ context.Context, in *tpb.GetMapLeavesByRevisionRequest, _ ...grpc.CallOption) (*tpb.MapLeaves, error) {
```
golangcibot: MapLeaves not declared by package trillian (from `typecheck`)

gdbelvin:
@pavelkalinnikov PTAL

core/sequencer/trillian_client.go
```diff
@@ -39,7 +39,7 @@ type trillianMap interface {
 	GetAndVerifyLatestMapRoot(ctx context.Context) (*tpb.SignedMapRoot, *types.MapRootV1, error)
 	SetLeavesAtRevision(ctx context.Context, rev int64, leaves []*tpb.MapLeaf, meta []byte) (*types.MapRootV1, error)
 	GetAndVerifyMapRootByRevision(ctx context.Context, rev int64) (*tpb.SignedMapRoot, *types.MapRootV1, error)
-	GetAndVerifyMapLeavesByRevision(ctx context.Context, rev int64, indexes [][]byte) ([]*tpb.MapLeaf, error)
+	GetMapLeavesByRevisionNoProof(ctx context.Context, rev int64, indexes [][]byte) ([]*tpb.MapLeaf, error)
```
pavelkalinnikov: Maybe just `GetMapLeavesByRevision`?

core/sequencer/trillian_client.go
```diff
@@ -141,3 +141,34 @@ func (c *MapClient) GetAndVerifyMapRootByRevision(ctx context.Context,
 	}
 	return rawMapRoot, mapRoot, nil
 }
+
+// GetMapLeavesByRevisionNoProof verifies and returns the requested map leaves at a specific revision.
```
pavelkalinnikov: Is "verifies" true?

core/sequencer/trillian_client.go
```diff
@@ -141,3 +141,34 @@ func (c *MapClient) GetAndVerifyMapRootByRevision(ctx context.Context,
...
+		return nil, err
+	}
+	getResp, err := c.Conn.GetLeavesByRevisionNoProof(ctx, &tpb.GetMapLeavesByRevisionRequest{
+		MapId:    c.MapID,
+		Index:    indexes,
+		Revision: revision,
+	})
+	if err != nil {
+		s := status.Convert(err)
+		return nil, status.Errorf(s.Code(), "map.GetLeaves(): %v", s.Message())
```
pavelkalinnikov: `GetLeavesByRevision[NoProof]`?

Also, maybe don't prefix errors returned from this method with "map.GetLeaves...". The client already knows which method they are calling, so they could add this prefix on their own, and once:

```golang
if err := c.GetMapLeavesByRevisionNoProof(...) {
	return fmt.Errorf("GetMapLeavesByRevisionNoProof: %v", err)
}
```

core/sequencer/trillian_client.go
```diff
@@ -141,3 +141,34 @@ func (c *MapClient) GetAndVerifyMapRootByRevision(ctx context.Context,
...
+	}
+	return getResp.Leaves, nil
+}
+
+// hasDuplicates returns an error if there are duplicates in indexes.
```
pavelkalinnikov: Why is this an error? Checking dups increases cost of every request, but in fact may not be needed cause clients don't send dups. Even if they did (and it was their fault), would it be bad to return 2 copies of the same data?

go.mod
```diff
@@ -4,26 +4,41 @@ go 1.12
 
 require (
 	github.com/benlaurie/objecthash v0.0.0-20180202135721-d1e3d6079fc1
-	github.com/go-sql-driver/mysql v1.4.1 // indirect
+	github.com/coreos/bbolt v1.3.2 // indirect
```
pavelkalinnikov: What caused all these changes? Should they be in a separate PR?

core/sequencer/trillian_client.go
pavelkalinnikov: More generally, methods that verify proofs could be "GetAndVerifyBlah", while non-proof ones be just "GetBlah"?

gdbelvin: This mimicks the underlying API call.
In general, yes, the API names need a broader review

core/sequencer/trillian_client.go
gdbelvin: Removed. Sorry about that.

core/sequencer/trillian_client.go
gdbelvin: Returning two copies of the same data is not horrible, but it does indicate that something has gone wrong.  This can be removed

go.mod
gdbelvin: see #1294

gdbelvin:
PTAL

PR #1303, opened 2019-06-28
Configurable maximum queue depth for metric reporting.

gdbelvin:
Computing the unapplied metric involves a full table scan.
Allow callers to configure how long to let the query run
before returning.


core/sequencer/server.go
```diff
@@ -186,16 +186,15 @@ func NewServer(
 }
 
 func (s *Server) UpdateMetrics(ctx context.Context, in *spb.UpdateMetricsRequest) (*spb.UpdateMetricsResponse, error) {
-	if err := s.unappliedMetric(ctx, in.DirectoryId); err != nil {
+	if err := s.unappliedMetric(ctx, in.DirectoryId, in.MaxUnappliedCount); err != nil {
```
pavelkalinnikov: Should `MaxUnappliedCount` be populated somewhere? Otherwise `UpdateMetrics` will read 0.

gdbelvin: Thanks for catching that. This should be set in the sequencer server

PR #1307, opened 2019-07-02
Make check-set mutation semantics optional

gdbelvin:
The hash chain in user entries is not currently being used.  `Entry.Previous`
was intended to help make account history verification efficient, but it did
not meet that goal.

In the meantime, `Entry.Previous` enforces quite a tight sequence of get,
modify, update.  Any delay in Key Transparency sequencing causes these updates
to fail, and continuously retrying exacerbates the issue.

### DIFF
--- a/core/mutator/entry/mutator_test.go
+++ b/core/mutator/entry/mutator_test.go
@@ -53,13 +53,11 @@ func TestCheckMutation(t *testing.T) {
 	// make the two entries (entryData1 and entryData2) different, otherwise
 	// it is not possible to test all cases.
 	key := []byte{0}
-	nilHash := sha256.Sum256(nil)
 
 	entryData1 := &tpb.Entry{
 		Index:            key,
 		Commitment:       []byte{1},
 		AuthorizedKeyset: keysetBytes(testPubKey1),
-		Previous:         nilHash[:],
 	}
 	signedEntryData1 := &tpb.SignedEntry{
 		Entry: mustMarshal(t, entryData1),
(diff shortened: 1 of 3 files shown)


core/testdata/directory_batch_get_user.json
```diff
@@ -1,7 +1,7 @@
 {
 	"directoryId": "integration",
 	"log": {
-		"treeId": "5360167840861150065",
+		"treeId": "7381771154556655958",
```
mhutchinson: It's not immediately clear to me why these tree ID updates are necessary, and why they are coupled to this CL?

gdbelvin:
After discussion in the "Rubber Duck Club™", we realized that preserving "Check Set" semantics is still important for clients that want to do "Fetch Modify Push".   

I'll modify this CL to make "Check Set" optional rather than removed.

core/testdata/directory_batch_get_user.json
gdbelvin: `treeId`, public keys, and various other bits in these test vectors are generated by reading from `/dev/random`.  This is less than ideal, but fixing it means patching out the randomness generators in Trillian which is a bit of a dangerous proposition.

gdbelvin: The reason these files are being *updated* is that the test vectors have changed due to the removal of "Previous"

gdbelvin:
Blocked on #1310

core/testdata/directory_batch_get_user.json
gdbelvin: Since making Previous optional rather than removed, the test vectors don't need updating.

gdbelvin:
This PR should greatly reduce the production error rate.  PTAL

core/mutator/entry/mutator_test.go
```diff
@@ -140,7 +135,7 @@ func TestCheckMutation(t *testing.T) {
 				entry: &tpb.Entry{
 					Index:            key,
 					Commitment:       []byte{2},
-					Previous:         nil,
+					Previous:         []byte("inavlidhash"),
```
mhutchinson: Not sure if the invalid spelling of invalid is part of the joke? :-)

gdbelvin: nope -- just a typo :-)

PR #1310, opened 2019-07-06
Track map revision inside mutation

gdbelvin:
After submitting a mutation, the client typically waits for a new
revision to appear before requesting for new data. 

This PR fixes a logic bug that assumed that the log root == map revision. 
Log roots and map revisions are separate concepts and should not be treated as such. 

Breaking this spurious dependency also supports refactoring root tracking #1311


core/client/batch_get_and_verify.go
```diff
@@ -95,7 +96,8 @@ func (c *Client) BatchVerifyGetUserIndex(ctx context.Context, userIDs []string)
 }
 
 // BatchVerifiedGetUser returns verified leaf values by userID.
-func (c *Client) BatchVerifiedGetUser(ctx context.Context, userIDs []string) (map[string]*pb.MapLeaf, error) {
+func (c *Client) BatchVerifiedGetUser(ctx context.Context, userIDs []string) (
```
mhutchinson: The doc string should be updated to explain the relationship between the returned values. 

It's not immediately clear to me what the contract of this method is. Is it supposed to return the latest value from the server, or is any "recent" (or even historical) revision okay? Passing in an explicit revision for the map would avoid this, but may go against some other design principles here?

core/mutator/entry/mutation.go
```diff
@@ -56,11 +57,12 @@ func NewMutation(index []byte, directoryID, userID string) *Mutation {
 
 // SetPrevious sets the previous hash.
 // If copyPrevious is true, AuthorizedKeys and Commitment are also copied.
-func (m *Mutation) SetPrevious(oldValue []byte, copyPrevious bool) error {
+func (m *Mutation) SetPrevious(rev uint64, oldValue []byte, copyPrevious bool) error {
```
mhutchinson: The target of the revision here is potentially ambiguous. My context-free instinct would be that this revision should be set to the revision at which the value was set to oldValue. My understanding from the rest of the code in this CL is that rev is actually intended to be set to the current live/latest revision for the map. This could be a much higher revision than when the value was last modified.

On further reading of waitOnceForUserUpdate I think either revision would be fine. Leaving this comment as I think a little clarity on expectation here would be beneficial.

core/client/batch_get_and_verify.go
gdbelvin: The API contract is indeed unclear and unverified.  
I'll add a TODO to fix this since the change affects multiple files. 
The intent is that this would fetch values against the *latest* map root, but as it is, the client does not enforce any constraint on the freshness of the MapRoot.

core/mutator/entry/mutation.go
gdbelvin: The intent here is to give the client a clue about how long to wait before attempting to verify that a mutation was applied correctly.   This behavior works best when `rev` is set to the revision at which `oldValue` was fetched since the client will wait until the server's reported current revision > `rev`.

core/client/client.go
```diff
@@ -329,12 +329,12 @@ func (c *Client) waitOnceForUserUpdate(ctx context.Context, m *entry.Mutation) (
 		return nil, fmt.Errorf("nil mutation")
 	}
 	// Wait for STH to change.
-	if err := c.WaitForSTHUpdate(ctx, int64(c.trusted.TreeSize)+1); err != nil {
+	if err := c.WaitForSTHUpdate(ctx, m.PrevRevision()+1); err != nil {
```
mhutchinson: Consider creating a method Mutation.ExpectedRevision() or MinApplicableRevision(), which returns PrevRevision()+1.

I'm not completely sold on the elegance of this, but "PrevRevision + 1" does map to some logical concept which is a property of the mutation.

gdbelvin: Thanks -- that's a much better API.

PR #1314, opened 2019-07-10
Pass along err message

gdbelvin:
This is useful when backends are overloaded.


core/keyserver/keyserver.go
```diff
@@ -272,7 +272,7 @@ func (s *Server) batchGetUserByRevision(ctx context.Context, sth *tpb.SignedLogR
 		})
 	if st := status.Convert(err); st.Code() != codes.OK {
 		glog.Errorf("tlog.GetInclusionProof(%v): %v", d.Log.TreeId, err)
-		return nil, status.Errorf(st.Code(), "Cannot fetch log inclusion proof")
+		return nil, status.Errorf(st.Code(), "Cannot fetch log inclusion proof: %v", st.Message())
```
Martin2112: Error message should start with lower case.

PR #1315, opened 2019-07-11
New test vector transcript format

gdbelvin:
This PR makes the test vector file format self contained. 

Previously public verification keys and the contents they were verifying were contained in separate files.

Issue #1311, opened 2019-07-08
Multi-threaded Client


gdbelvin:
The client currently has a full mutex around each RPC, preventing it from being used in multi-threaded use cases.

gdbelvin:
Proposed new design:

```
Client
    Verifier
        RootTracker
```

The client verifier will have a `RootTracker` which can 
1. Skip consistency proof checking
2. Enforce a mutex
3. Track multiple roots

gdbelvin:
#1310 Fixes a logic bug.  Rather than depending on the last verified log root (which may now be subject to change between calls), mutations should record the revision they were created at on their own.

gdbelvin:
Closed by #1317

PR #1316, opened 2019-07-11
Encapsulate Client Verifier State in test vectors

gdbelvin:
In order to remove the client mutex, we need to fully extract and isolate client verifier state.
Explicitly putting the client verifier state in the test vectors is the first step towards doing that. 

Blocked on #1315
Motivated by #1311


core/testdata/transcript_go_proto/transcript.proto
```diff
@@ -34,10 +34,9 @@ message Transcript {
 // Action contains one request and one response.
 message Action {
    string desc = 1;
-   // trust_new_log controls the client verifier state.
-   // If true, the client records the resp.LatestLogRoot and uses it
-   // for the next rpc's request.last_verified_tree_size.
-   bool trust_new_log = 2;
+   reserved 2;  // was trust_new_log
+   // last_verified_log_root encodes the client verifier state.
```
pavelkalinnikov: optional: In Go generated code the field name is formatted as `LastVerifiedLogRoot`, which is not consistent with formatting in the comment. My strategy is to not start comments with field names in proto files.

core/api/v1/keytransparency.proto
```diff
@@ -338,6 +338,15 @@ message MapRoot {
   repeated bytes log_inclusion = 2;
 }
 
+// LogRootRequest contains the information needed to requetst and verify LogRoot.
```
pavelkalinnikov: typo: to request

core/api/v1/keytransparency.proto
```diff
@@ -338,6 +338,15 @@ message MapRoot {
...
+message LogRootRequest {
+  // root_hash is the log root hash.
+  bytes root_hash = 1;
```
pavelkalinnikov: Why does the log root request contain the log root hash? Isn't it supposed to be returned as a result of the request?

core/api/v1/keytransparency.proto
```diff
@@ -338,6 +338,15 @@ message MapRoot {
   repeated bytes log_inclusion = 2;
 }
 
+// LogRootRequest contains the information needed to requetst and verify LogRoot.
+message LogRootRequest {
+  // root_hash is the log root hash.
```
pavelkalinnikov: optional: I'd avoid using the field name in the comments, because the Go generated code uses a different name. Passim.

core/testdata/transcript_go_proto/transcript.proto
```diff
@@ -34,10 +34,9 @@ message Transcript {
 // Action contains one request and one response.
 message Action {
    string desc = 1;
-   // trust_new_log controls the client verifier state.
-   // If true, the client records the resp.LatestLogRoot and uses it
-   // for the next rpc's request.last_verified_tree_size.
-   bool trust_new_log = 2;
+   reserved 2;  // was trust_new_log
```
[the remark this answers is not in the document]
gdbelvin: The `reserved` keyword just helps avoid tag number reuse, which is the real cause of problems. 
If tag numbers get reused the same binary encoding could end up decoding to different messages depending on what version of the message is being used.

core/testdata/transcript_go_proto/transcript.proto
gdbelvin: Alas, this is a common problem. I used to use the Go capitalization so the comments would come out correctly, but the mixing of different capitalization schemes in the proto file seemed messy and language specific. 

Unsure if not starting comments with field names fixes the problem, since doing so still violates go style in the generated code.

core/api/v1/keytransparency.proto
gdbelvin: The `LogRootRequest` is intended to encode the client's verifier state, as is being done in this PR; but it has a second purpose, which is to make request, response pairs statelessly verifiable.  At the moment, an observer of the system, or a server interceptor can't tell if a response is valid because the first hash of the consistency proof is not included in either the request or the response.

core/api/v1/keytransparency.proto
gdbelvin: See above

gdbelvin:
Updated. PTAL

impl/integration/main_test.go
```diff
@@ -42,24 +42,21 @@ func TestIntegration(t *testing.T) {
...
-				actions := test.Fn(ctx, env.Env, t)
-				if *generate {
-					if err := testdata.WriteTranscript(test.Name, &tpb.Transcript{
-						Description: test.Name,
-						Directory:   env.Env.Directory,
-						Actions:     actions,
-					}); err != nil {
-						t.Fatalf("WriteTranscript() failed: %v", err)
-					}
+			cctx, cancel := context.WithCancel(ctx)
```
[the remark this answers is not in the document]
gdbelvin: It is extraneous -- I can move to a separate PR.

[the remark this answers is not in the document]
gdbelvin: #1319

PR #1317, opened 2019-07-11
Extract log tracking into separate package 

gdbelvin:
This PR isolates root tracking inside a new `tracker` package. 
By putting all root tracking behind an interface, we can implement alternate, multi-threaded root trackers.  #1311

### DIFF
--- a/core/client/tracker/tracker.go
+++ b/core/client/tracker/tracker.go
@@ -32,8 +32,14 @@ type LogTracker struct {
 	v       *tclient.LogVerifier
 }
 
+// New creates a log tracker from no trusted root.
 func New(lv *tclient.LogVerifier) *LogTracker {
 	return &LogTracker{v: lv}
+}
+
+// NewFromSaved creates a log tracker from a previously saved trusted root.
+func NewFromSaved(lv *tclient.LogVerifier, lr types.LogRootV1) *LogTracker {
+	return &LogTracker{v: lv, trusted: lr}
 }
 
 // LastVerifiedTreeSize retrieves the tree size of the latest log root
--- a/core/client/tracker/tracker.go
+++ b/core/client/tracker/tracker.go
@@ -16,6 +16,8 @@
 package tracker
 
 import (
+	"sync"
+
 	"github.com/golang/glog"
 	"github.com/golang/protobuf/proto"
 	"github.com/google/trillian/types"
@@ -30,7 +32,7 @@ import (
 // for the latest SignedLogRoot should be updated.
 type UpdateTrustedPredicate func(cntRoot, newRoot types.LogRootV1) bool
 
-// LogRootVerifier verifies trillian Log Root.
+// LogRootVerifier verifies a Trillian Log Root.
 type LogRootVerifier interface {
 	VerifyRoot(trusted *types.LogRootV1, newRoot *tpb.SignedLogRoot, consistency [][]byte) (*types.LogRootV1, error)
 }
@@ -40,6 +42,7 @@ type LogTracker struct {
 	trusted       types.LogRootV1
 	v             LogRootVerifier
 	updateTrusted UpdateTrustedPredicate
+	mu            sync.RWMutex
 }
 
 // New creates a log tracker from no trusted root.
@@ -52,9 +55,10 @@ func NewFromSaved(lv LogRootVerifier, lr types.LogRootV1) *LogTracker {
 	return &LogTracker{v: lv, trusted: lr, updateTrusted: isNewer}
 }
 
-// LastVerifiedLogRoot retrieves the tree size of the latest log root
-// and it blocks further requests until VerifyRoot is called.
+// LastVerifiedLogRoot retrieves the tree size of the latest log root.
 func (l *LogTracker) LastVerifiedLogRoot() *pb.LogRootRequest {
+	l.mu.RLock()
+	defer l.mu.RUnlock()
 	return l.logRootRequest()
 }
 
@@ -66,9 +70,11 @@ func (l *LogTracker) logRootRequest() *pb.LogRootRequest {
 }
 
 // VerifyLogRoot verifies root and updates the trusted root if it is newer.
-// VerifyLogRoot unblocks the next call to LastVerifiedTreeSize.
-// req must come from LastVerifiedLogRoot()
+// req must be equal to the most recent value from LastVerifiedLogRoot().
+// If two clients race to VerifyLogRoot at the same time, if one of them updates the root, the other will fail.
 func (l *LogTracker) VerifyLogRoot(req *pb.LogRootRequest, root *pb.LogRoot) (*types.LogRootV1, error) {
+	l.mu.Lock()
+	defer l.mu.Unlock()
 	if want := l.logRootRequest(); !proto.Equal(req, want) {
 		glog.Warningf("logtracker: unexpected logRootRequest: %v, want %v", req, want)
 		return nil, status.Errorf(codes.InvalidArgument, "out of order VerifyLogRoot(%v, _), want %v", req, want)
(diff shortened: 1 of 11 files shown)


cmd/keytransparency-monitor/main.go
```diff
@@ -85,10 +84,10 @@ func main() {
 		glog.Exitf("Failed to initialize monitor: %v", err)
 	}
 
-	// TODO(gbelvin): persist trusted roots
-	trusted := types.LogRootV1{}
 	go func() {
-		if err := mon.ProcessLoop(ctx, *directoryID, trusted); err != nil {
+		// TODO(gbelvin): persist last verified revisioni
+		var lastVerifiedRevision int64 = 0
```
golangcibot: should drop = 0 from declaration of var lastVerifiedRevision; it is the zero value (from `golint`)

core/client/tracker/tracker.go
```diff
@@ -0,0 +1,113 @@
...
+	tpb "github.com/google/trillian"
+)
+
+// UpdateTrustedPredicate return a bool indicating whether the local reference
+// for the latest SignedLogRoot should be updated.
+type UpdateTrustedPredicate func(cntRoot, newRoot types.LogRootV1) bool
+
+// LogRootVerifier verifies a Trillian Log Root.
+type LogRootVerifier interface {
+	VerifyRoot(trusted *types.LogRootV1, newRoot *tpb.SignedLogRoot, consistency [][]byte) (*types.LogRootV1, error)
```
pavelkalinnikov: Please add a comment.
Also, how about renaming `consistency` to `proof`, and clarifying in a comment that it's a consistency proof between the two roots?

core/client/tracker/tracker.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// LogTracker keeps a continuous series of consistent log roots.
```
pavelkalinnikov: It doesn't keep the whole series, does it? Maybe "keeps track of", "tracks" or "advances through"?

core/client/tracker/tracker.go
```diff
@@ -0,0 +1,113 @@
...
+func (l *LogTracker) logRootRequest() *pb.LogRootRequest {
+	return &pb.LogRootRequest{
+		TreeSize: int64(l.trusted.TreeSize),
+		RootHash: l.trusted.RootHash,
+	}
+}
+
+// VerifyLogRoot verifies root and updates the trusted root if it is newer.
+// req must be equal to the most recent value from LastVerifiedLogRoot().
+// If two clients race to VerifyLogRoot at the same time, if one of them updates the root, the other will fail.
```
pavelkalinnikov: If the verification fails, it is critical, but if it's just a race then it looks recoverable for the clients? Do you handle this race thing in clients?

core/client/tracker/tracker.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+	if l.updateTrusted(l.trusted, *logRoot) {
+		l.trusted = *logRoot
+		glog.Infof("Trusted root updated to TreeSize %v", l.trusted.TreeSize)
+	}
+	return logRoot, nil
+}
+
+// SetUpdatePredicate allows tests to have finer grained control over when
+// the trusted root is updated.
```
pavelkalinnikov: Add something like "Must be called before any other method."?

core/client/tracker/tracker.go
```diff
@@ -0,0 +1,113 @@
...
+func (l *LogTracker) SetUpdatePredicate(f UpdateTrustedPredicate) {
+	l.updateTrusted = f
+}
+
+// isNewer returns true when newRoot is newer than cntRoot.
+func isNewer(cntRoot, newRoot types.LogRootV1) bool {
+	if newRoot.TreeSize > cntRoot.TreeSize {
```
pavelkalinnikov: Or simply:
```golang
return newRoot.TreeSize > cntRoot.TreeSize ||
		(newRoot.TreeSize == cntRoot.TreeSize &&
			newRoot.TimestampNanos > cntRoot.TimestampNanos)
```

core/client/verifier/verifier.go
```diff
@@ -42,23 +45,31 @@ var (
 	ErrNilProof = errors.New("nil proof")
 )
 
+type LogTracker interface {
```
pavelkalinnikov: Add type and method comments? (some linters force it for exported types)

core/client/verifier/verifier.go
```diff
@@ -42,23 +45,31 @@ var (
...
+	LastVerifiedLogRoot() *pb.LogRootRequest
+	VerifyLogRoot(req *pb.LogRootRequest, root *pb.LogRoot) (*types.LogRootV1, error)
+}
+
 // Verifier is a client helper library for verifying request and responses.
```
pavelkalinnikov: requests

core/integration/client_tests.go
```diff
@@ -166,10 +171,48 @@ func TestBatchUpdate(ctx context.Context, env *Env, t *testing.T) []*tpb.Action
...
+	return cli, logTracker
+}
+
 // TestEmptyGetAndUpdate verifies set/get semantics.
 func TestEmptyGetAndUpdate(ctx context.Context, env *Env, t *testing.T) []*tpb.Action {
 	go runSequencer(ctx, t, env.Directory.DirectoryId, env)
 
+	cli, logTracker := NewClientWithTracker(t, env)
+	logTracker.SetUpdatePredicate(func(_, newRoot types.LogRootV1) bool {
+		return newRoot.TreeSize%5 == 1
```
pavelkalinnikov: Why do we need to test this predicate, and not the original one?

core/client/tracker/tracker.go
gdbelvin: The verification is a critical failure.  
The previous client would block trying to obtain a mutex and timeout. 
This implementation will allow clients to make calls optimistically and then fail the verification if things get out of sync. 

In its current state, the tracker and client libraries are not designed to be used in a multi-threaded context.   However, this is changing with #1311.  

The Next PR will try to give nieve multi-threaded support by *not* updating the trusted root.

core/client/tracker/tracker.go
gdbelvin: It can be set at any time, but I'll suggest setting it before other calls.

core/client/tracker/tracker.go
gdbelvin: I find this harder to read?

core/integration/client_tests.go
gdbelvin: Good question.
a) The original predicate is tested in the tracker's unit tests. 
b) The purpose of this is to produce more interesting consistency proofs in the test vectors.  If we updated the log root every time, the consistency proofs would always be empty.

core/integration/client_tests.go
```diff
@@ -166,10 +171,49 @@ func TestBatchUpdate(ctx context.Context, env *Env, t *testing.T) []*tpb.Action
...
+	return cli, logTracker
+}
+
 // TestEmptyGetAndUpdate verifies set/get semantics.
 func TestEmptyGetAndUpdate(ctx context.Context, env *Env, t *testing.T) []*tpb.Action {
 	go runSequencer(ctx, t, env.Directory.DirectoryId, env)
 
+	cli, logTracker := NewClientWithTracker(t, env)
+	logTracker.SetUpdatePredicate(func(_, newRoot types.LogRootV1) bool {
+		// Only update occationally in order to produce interesting consistency proofs.
```
golangcibot: `occationally` is a misspelling of `occasionally` (from `misspell`)

core/integration/client_tests.go
```diff
@@ -315,6 +353,12 @@ func TestBatchGetUser(ctx context.Context, env *Env, t *testing.T) []*tpb.Action
 	authorizedKeys1 := testutil.VerifyKeysetFromPEMs(testPubKey1)
 	transcript := []*tpb.Action{}
 
+	cli, logTracker := NewClientWithTracker(t, env)
+	logTracker.SetUpdatePredicate(func(_, newRoot types.LogRootV1) bool {
+		// Only update occationally in order to produce interesting consistency proofs.
```
golangcibot: `occationally` is a misspelling of `occasionally` (from `misspell`)

PR #1320, opened 2019-07-15
Remove VerifySignedMapRoot from VerifierInterface

gdbelvin:
All verifications should be rooted in the last verified log root.

Pass already verified `mapRoot` to `RevisionMutations`.  There's no need to re-verify it without the context of a last verified log root.

### DIFF
--- a/core/client/client.go
+++ b/core/client/client.go
@@ -29,7 +29,6 @@ import (
 
 	"github.com/google/keytransparency/core/client/verifier"
 	"github.com/google/keytransparency/core/mutator/entry"
-	"github.com/google/trillian"
 
 	"github.com/google/trillian/client/backoff"
 	"github.com/google/trillian/types"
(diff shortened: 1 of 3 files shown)


core/client/client.go
```diff
@@ -77,8 +76,6 @@ type VerifierInterface interface {
 	// VerifyRevision verifies that revision is correctly signed and included in the append only log.
 	// VerifyRevision also verifies that revision.LogRoot is consistent with the last trusted SignedLogRoot.
 	VerifyRevision(revision *pb.Revision, trusted types.LogRootV1) (*types.LogRootV1, *types.MapRootV1, error)
-	// VerifySignedMapRoot verifies the signature on the SignedMapRoot.
-	VerifySignedMapRoot(smr *trillian.SignedMapRoot) (*types.MapRootV1, error)
```
pavelkalinnikov: Do you still need the implementations of this method?

gdbelvin: We don't, and it sort of doesn't exit in the sense that it's a property of the Trillian map verifier which the verifier inherits.  I can pull it out to make it more explicit if you'd like?

PR #1323, opened 2019-07-16
Verify Revisions in StreamRevisions

gdbelvin:
By verifying revisions inside of StreamRevisions, 
we can reuse VerifyGetRevision and have better control
over the updating of LogRoot

### DIFF
--- a/cmd/keytransparency-monitor/main.go
+++ b/cmd/keytransparency-monitor/main.go
@@ -24,7 +24,6 @@ import (
 	"github.com/golang/glog"
 	"github.com/google/trillian/crypto"
 	"github.com/google/trillian/crypto/keys/pem"
-	"github.com/google/trillian/types"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
@@ -85,10 +84,8 @@ func main() {
 		glog.Exitf("Failed to initialize monitor: %v", err)
 	}
 
-	// TODO(gbelvin): persist trusted roots
-	trusted := types.LogRootV1{}
 	go func() {
-		if err := mon.ProcessLoop(ctx, trusted); err != nil {
+		if err := mon.ProcessLoop(ctx, 0); err != nil {
 			glog.Errorf("ProcessLoop: %v", err)
 		}
 	}()
--- a/cmd/keytransparency-monitor/main.go
+++ b/cmd/keytransparency-monitor/main.go
@@ -88,7 +88,7 @@ func main() {
 	// TODO(gbelvin): persist trusted roots
 	trusted := types.LogRootV1{}
 	go func() {
-		if err := mon.ProcessLoop(ctx, *directoryID, trusted); err != nil {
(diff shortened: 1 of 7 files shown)


cmd/keytransparency-monitor/main.go
```diff
@@ -85,10 +84,8 @@ func main() {
 		glog.Exitf("Failed to initialize monitor: %v", err)
 	}
 
-	// TODO(gbelvin): persist trusted roots
```
pavelkalinnikov: Is this TODOne? If the monitor restarts, it will start over from rev 0.

core/client/get_and_verify.go
```diff
@@ -94,7 +94,7 @@ func (c *Client) VerifiedGetLatestRevision(ctx context.Context) (*types.LogRootV
 // VerifiedGetRevision fetches the requested revision from the key server.
 // It also verifies the consistency of the latest log root against the last seen log root.
 // Returns the latest log root and the requested map root.
```
pavelkalinnikov: Please update the comment.

core/client/mutations.go
```diff
@@ -39,16 +39,13 @@ type RevisionMutations struct {
 // StreamRevisions repeatedly fetches revisions and sends them to out until GetRevision
 // returns an error other than NotFound or until ctx.Done is closed.  When
 // GetRevision returns NotFound, it waits one pollPeriod before trying again.
-func (c *Client) StreamRevisions(ctx context.Context, directoryID string, startRevision int64, out chan<- *pb.Revision) error {
+func (c *Client) StreamRevisions(ctx context.Context, startRevision int64, out chan<- *types.MapRootV1) error {
 	defer close(out)
 	wait := time.NewTicker(c.RetryDelay).C
 	for i := startRevision; ; {
 		// time out if we exceed the poll period:
```
pavelkalinnikov: Can you format this comment?

Also, is it accurate? It's not obvious from the `VerifiedGetRevision` invocation that it has some timing out built in.

core/client/mutations.go
```diff
@@ -39,16 +39,13 @@ type RevisionMutations struct {
 // StreamRevisions repeatedly fetches revisions and sends them to out until GetRevision
 // returns an error other than NotFound or until ctx.Done is closed.  When
 // GetRevision returns NotFound, it waits one pollPeriod before trying again.
-func (c *Client) StreamRevisions(ctx context.Context, directoryID string, startRevision int64, out chan<- *pb.Revision) error {
+func (c *Client) StreamRevisions(ctx context.Context, startRevision int64, out chan<- *types.MapRootV1) error {
 	defer close(out)
 	wait := time.NewTicker(c.RetryDelay).C
```
pavelkalinnikov: Does the ticker leak if you don't `Stop` it?

cmd/keytransparency-monitor/main.go
gdbelvin: This is now refactored into the client which will be refactored into a log tracker soon

core/client/mutations.go
gdbelvin: It's not accurate

PR #1324, opened 2019-07-16
Remove unused logVerifier

gdbelvin:


gdbelvin:
Indeed. Though it's not quite as bad as it looks.
The monitor does have integration tests.

PR #1325, opened 2019-07-18
Cache go mod download

gdbelvin:


cmd/keytransparency-server/Dockerfile
```diff
@@ -1,9 +1,13 @@
 FROM golang:1.12 as build
 
 WORKDIR /go/src/github.com/google/keytransparency
```
AlCutter: Would be neat if these all shared the same `build` stage, is that possible?

gdbelvin: How do I do that?

gdbelvin: Or what would be the value in doing that?

cmd/keytransparency-monitor/Dockerfile
```diff
@@ -1,9 +1,13 @@
 FROM golang:1.12 as build
 
 WORKDIR /go/src/github.com/google/keytransparency
-COPY . .
+COPY go.mod .
+COPY go.sum .
```
RJPercival: Can combine these two steps as `COPY go.mod go.sum ./`

cmd/keytransparency-server/Dockerfile
AlCutter: I was thinking that if there were something like:

```
FROM ... as base
...
go mod download
COPY ...
```

which was then shared by all the job-specific Dockerfiles, e.g.:
```

FROM base as build:
go get .../${thing}

FROM distroless/...
COPY --from=build ...
```
 Then you'd save having to rebuild the same `base` image/re-fetching the same modules etc. repeatedly.

PR #1327, opened 2019-07-19
Tracker Factory for dependency injection

gdbelvin:
Allow callers of NewFromConfig to supply a tracker factory.


core/client/verifier/verifier.go
```diff
@@ -53,6 +52,8 @@ type LogTracker interface {
 	VerifyLogRoot(state *pb.LogRootRequest, newRoot *pb.LogRoot) (*types.LogRootV1, error)
 }
 
+type LogTrackerFactory func(*tclient.LogVerifier) LogTracker
```
mhutchinson: Given the length of the conversation we had around this yesterday, I think a short description here would be worthwhile. It only needs to say that this allows different implementation strategies, including a no-op implementation, for a Tracker to be injected based on a verifier. It feels obvious now, but I'm not sure that would be true without the context.

core/client/gobindclient/client.go
```diff
@@ -107,7 +110,9 @@ func AddKtServer(ktURL string, insecureTLS bool, ktTLSCertPEM []byte, directoryI
 		}
 	}
 
-	client, err := client.NewFromConfig(ktClient, config)
+	client, err := client.NewFromConfig(ktClient, config,
+		func(lv *tclient.LogVerifier) verifier.LogTracker { return tracker.New(lv) },
```
mhutchinson: Potentially in a future CL, I recommend renaming tracker.New to something like tracker.NewSynchronousHead or equivalent. This is something that may have many different implementations that could all be reasonable given different requirements.

core/client/tracker/tracker.go
```diff
@@ -46,8 +46,8 @@ type LogTracker struct {
 	mu            sync.RWMutex
 }
 
-// New creates a log tracker from no trusted root.
-func New(lv LogRootVerifier) *LogTracker {
+// NewSyncronous creates a log tracker from no trusted root.
+func NewSyncronous(lv LogRootVerifier) *LogTracker {
```
mhutchinson: s/Syncronous/Synchronous/

core/client/verifier/verifier.go
```diff
@@ -53,6 +52,10 @@ type LogTracker interface {
 	VerifyLogRoot(state *pb.LogRootRequest, newRoot *pb.LogRoot) (*types.LogRootV1, error)
 }
 
+// LogTrackerFactory allows the caller of NewFromDirectory to supply different
+// tracker implementations.  eg. no-op, single-syncronous-head, and multi-head tracking.
```
golangcibot: `syncronous` is a misspelling of `synchronous` (from `misspell`)

PR #1331, opened 2019-08-05
Use Travis Build Stages

gdbelvin:
This speeds up the build
Isolates installs build specific stages
Speeds up error identification to specific stages

Depends on #1330

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,28 @@ env:
   - CLOUDSDK_CORE_DISABLE_PROMPTS=1
   - GO111MODULE=on
 
+jobs:
+  include:
+    - name: "build"
+      script: go build ./...
+    - name: "lint"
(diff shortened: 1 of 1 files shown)


gdbelvin:
Depends on #1330

.travis.yml
```diff
@@ -18,6 +18,29 @@ env:
...
+    - name: "coverage"
+      script:
+      - go test ./... -coverprofile=coverage1.txt
+      - go test ./impl/integration/ -coverprofile=coverage2.txt -coverpkg=./...
+      after_success: bash <(curl -s https://codecov.io/bash)
+    - name: "docker integration test"
+      install: skip
+      script: ./scripts/integration.sh
+    - stage: deploy
+      if: branch = master AND type != pull_request
```
pavelkalinnikov: Is this equivalent to the previous behavior, or more restrictive? Maybe split into 2 PRs in the latter case?

gdbelvin: This is exactly equivalent to previous behavior. 
I don't know why Travis made it seem like it wasn't

PR #1332, opened 2019-08-05
Fix deploy scripts

gdbelvin:
Depends on #1330


scripts/deploy.sh
```diff
@@ -35,6 +35,7 @@ gcloud --quiet auth configure-docker
 
 
 echo "Generating keys..."
+rm -f ./genfiles/*
```
mhutchinson: Before running a command like this, it seems prudent to check/enforce the current directory is what you expect it to be.

gdbelvin:
Added some checks -- not sure if they are rigorous enough though. 
Key Transparency can be checked out in any directory now that Go modules are a thing. This makes creating strong checks around what directory the script expects to be run in more difficult.

scripts/deploy.sh
```diff
@@ -33,8 +33,11 @@ gcloud --quiet config set container/cluster ${CLUSTER_NAME_CI}
 gcloud --quiet container clusters get-credentials ${CLUSTER_NAME_CI}
 gcloud --quiet auth configure-docker
 
+# Test current directory before deleting anything
+test $(basename $(pwd)) == "keytransparency" || exit
```
pavelkalinnikov: Should this exit with a non-zero code?

gdbelvin: Done

PR #1334, opened 2019-08-08
Check for data races in travis

gdbelvin:
This PR both adds race tests, and fixes a few outstanding race conditions. 

Blocked on #1336 -- Fixing some problem with GOPROXY

### DIFF
--- a/core/keyserver/keyserver.go
+++ b/core/keyserver/keyserver.go
@@ -654,7 +654,7 @@ func (s *Server) BatchQueueUserUpdate(ctx context.Context, in *pb.BatchQueueUser
 						glog.Warningf("Invalid UpdateEntryRequest: %v", err)
 						errors <- status.Errorf(codes.InvalidArgument, "Invalid mutation")
 					}
-					if err = validateEntryUpdate(u, vrfPriv); err != nil {
+					if err := validateEntryUpdate(u, vrfPriv); err != nil {
 						glog.Warningf("Invalid UpdateEntryRequest: %v", err)
 						errors <- status.Errorf(codes.InvalidArgument, "Invalid request")
 					}
--- a/core/keyserver/keyserver.go
+++ b/core/keyserver/keyserver.go
@@ -18,7 +18,6 @@ package keyserver
 import (
 	"context"
 	"fmt"
-	"log"
 	"runtime"
 	"sync"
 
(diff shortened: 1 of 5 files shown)


core/keyserver/keyserver.go
```diff
@@ -636,36 +635,30 @@ func (s *Server) BatchQueueUserUpdate(ctx context.Context, in *pb.BatchQueueUser
...
-						errors <- status.Errorf(codes.InvalidArgument, "Invalid request")
-					}
-				}
-			}()
-		}
-	}()
-	for e := range errors {
-		return nil, e
+	g, _ := errgroup.WithContext(ctx)
+	for u := range updates {
```
pavelkalinnikov: Btw, the `updates` channel seems redundant here, this could iterate the `in.Updates` slice directly.

core/keyserver/keyserver.go
```diff
@@ -636,36 +635,30 @@ func (s *Server) BatchQueueUserUpdate(ctx context.Context, in *pb.BatchQueueUser
...
 	_, tdone := monitoring.StartSpan(ctx, "BatchQueueUserUpdate.verify")
 	updates := make(chan *pb.EntryUpdate)
-	errors := make(chan error)
 	go func() {
 		defer close(updates)
 		for _, u := range in.Updates {
 			updates <- u
 		}
 	}()
-	go func() {
```
pavelkalinnikov: What was the race in this block? If there is none, can this be in a separate PR?

gdbelvin: The race in this block was that `Wait()` was being called concurrent to `Add()`

core/keyserver/keyserver.go
gdbelvin: good point

core/keyserver/keyserver.go
pavelkalinnikov: Doesn't look like it. All `Add`s were done synchronously, strictly before the deferred `Wait` was called.

gdbelvin: Sorry - I was talking about the race condition in core/client/batch_get_and_verify.go

The race condition in this block was `err = validateEntryUpdate`. The fix: `err := validateEntryUpdate`

pavelkalinnikov: Oh, this is a good one :)

Well the refactoring is a bit out of scope then. But LGTM, as I don't want to block you.

PR #1335, opened 2019-08-08
Factor out defining revisions into an async job

pavelkalinnikov:


pavelkalinnikov:
First cut. PTAL.

pavelkalinnikov:
Some integrations tests need tweaking.

core/sequencer/server.go
```diff
@@ -222,72 +223,58 @@ func (s *Server) unappliedMetric(ctx context.Context, directoryID string, maxCou
...
+	resp := &spb.DefineRevisionsResponse{
+		Begin: int64(root.Revision) + 1,
+		End:   high,
 	}
 
 	// Don't create new revisions if there are ones waiting to be applied.
-	if len(outstanding) > 0 {
-		return &spb.DefineRevisionsResponse{OutstandingRevisions: outstanding}, nil
+	// TODO(#1056): Allow going ahead to some extent.
+	if resp.End > resp.Begin {
```
pavelkalinnikov: We probably should allow at lest 1 outstanding revision here, otherwise this PR will only make things 1.5 times slower.

core/sequencer/sequencer.go
```diff
@@ -121,16 +121,31 @@ func (s *Sequencer) ForAllMasterships(ctx context.Context, f func(ctx context.Co
...
 }
 
-// RunBatchForAllMasterships runs KeyTransparencySequencerClient.RunBatch on all
-// directories this sequencer is currently master for.
-func (s *Sequencer) RunBatchForAllMasterships(ctx context.Context, batchSize int32) error {
-	glog.Infof("RunBatchForAllMasterships")
+// DefineRevisionsForAllMasterships runs KeyTransparencySequencerClient's
+// DefineRevisions method on all directories that this sequencer is currently
+// master for.
+func (s *Sequencer) DefineRevisionsForAllMasterships(ctx context.Context, batchSize int32) error {
```
gdbelvin: I think there's a more generic RunForAllMasterships function that might reduce some code duplication

core/sequencer/sequencer_api.proto
```diff
@@ -75,8 +70,29 @@ message DefineRevisionsRequest {
...
+  // begin is the first outstanding revision (inclusive).
+  int64 begin = 2;
+  // end is the last outstanding revision (exclusive). It is equal to begin if
+  // there are no outstanding revisions.
+  int64 end = 3;
+}
+
+// GetNewRevisionsRequest contains information needed to query the set of
+// outstanding revisions.
+message GetNewRevisionsRequest {
```
jtoohill: Maybe "GetUnappliedRevisionsRequest/Response" or "GetOutstandingRevisionsRequest/Response"? "New" could be a misnomer, as revisions may be defined but not applied for a long time. Whatever you pick, be consistent with the field naming above.

core/sequencer/sequencer_api.proto
```diff
@@ -49,13 +49,8 @@ message MapMetadata {
...
   // directory_id is the directory to run for.
   string directory_id = 1;
-  // min_batch is the minimum number of items in a batch.
-  // If less than min_batch items are available, nothing happens.
-  // TODO(#1047): Replace with timeout so items in the log get processed
-  // eventually.
-  int32 min_batch = 2;
-  // max_batch is the maximum number of items in a batch.
-  int32 max_batch = 3;
+  // Reserved for the deprecated min_batch and max_batch fields.
```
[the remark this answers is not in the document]
pavelkalinnikov: Done.

core/sequencer/sequencer_api.proto
pavelkalinnikov: Done.

core/sequencer/sequencer_api.proto
```diff
@@ -75,8 +70,29 @@ message DefineRevisionsRequest {
 
 // DefineRevisionResponse contains information about freshly defined revisions.
 message DefineRevisionsResponse {
-  // outsanding_revisions a list of all the defined revisions which are not yet applied.
-  repeated int64 outstanding_revisions = 1;
+  // Reserved for the deprecated outstanding_revisions field.
+  reserved 1;
+  // begin is the first outstanding revision (inclusive).
+  int64 begin = 2;
```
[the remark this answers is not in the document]
pavelkalinnikov: Done.

core/integration/client_tests.go
```diff
@@ -72,11 +72,7 @@ func runSequencer(ctx context.Context, t *testing.T, dirID string, env *Env) {
 	ticker := time.NewTicker(100 * time.Millisecond)
 	defer ticker.Stop()
 	sequencer.PeriodicallyRun(ctx, ticker.C, func(ctx context.Context) {
```
gdbelvin: Do you also need to run `DefineRevision` in addition to `RunBatch` here?

gdbelvin:
Real integration test is failing with `Error: update failed: context deadline exceeded` -- are all the right go routines getting launched?

core/sequencer/server.go
```diff
@@ -304,20 +291,56 @@ func (s *Server) DefineRevisions(ctx context.Context,
...
+		Begin: int64(root.Revision) + 1,
+		End:   high,
+	}
+	if revs.End < revs.Begin { // In case there is a race.
+		revs.End = revs.Begin // Avoid ugly results.
+	}
+	return revs, nil
+}
+
+// getRootAndRev returns the latest map root, and the first unprocessed
```
gdbelvin: The name is a bit confusing -- maybe because I made it do something confusing?
Perhaps this should be split into two functions?

LatestAppliedMapRevision
LatestDefinedBatchRevision

core/sequencer/sequencer_api.proto
```diff
@@ -75,8 +70,29 @@ message DefineRevisionsRequest {
 
 // DefineRevisionResponse contains information about freshly defined revisions.
 message DefineRevisionsResponse {
-  // outsanding_revisions a list of all the defined revisions which are not yet applied.
-  repeated int64 outstanding_revisions = 1;
+  // unapplied_begin is the first outstanding revision (inclusive).
+  int64 unapplied_begin = 2;
```
gdbelvin: I find this a bit harder to grok.  
Was there a particular reason for the change?

core/sequencer/server.go
```diff
@@ -222,72 +223,58 @@ func (s *Server) unappliedMetric(ctx context.Context, directoryID string, maxCou
...
+	// Get the list of unprocessed revisions.
+	root, high, err := s.getRootAndRev(ctx, in.DirectoryId)
 	if err != nil {
 		return nil, err
 	}
-	outstanding := []int64{}
-	for rev := int64(latestMapRoot.Revision) + 1; rev <= highestRev; rev++ {
-		outstanding = append(outstanding, rev)
+	resp := &spb.DefineRevisionsResponse{
+		UnappliedBegin: int64(root.Revision) + 1,
```
gdbelvin:
```suggestion
		UnappliedBegin: int64(root.Revision) + 1,  // The next map revision to apply
```

core/sequencer/server.go
```diff
@@ -222,72 +223,58 @@ func (s *Server) unappliedMetric(ctx context.Context, directoryID string, maxCou
...
+		UnappliedEnd:   high,
```
gdbelvin:
```suggestion
		UnappliedEnd:   high,  // The highest pre-defined revision from the batch definitions table. 
```

core/sequencer/server.go
```diff
@@ -222,72 +223,58 @@ func (s *Server) unappliedMetric(ctx context.Context, directoryID string, maxCou
...
-	}
-	_, latestMapRoot, err := mapClient.GetAndVerifyLatestMapRoot(ctx)
-	if err != nil {
-		return nil, err
-	}
-
-	// Collect a list of unapplied revisions.
-	highestRev, err := s.batcher.HighestRev(ctx, in.DirectoryId)
+	// Get the list of unprocessed revisions.
+	root, high, err := s.getRootAndRev(ctx, in.DirectoryId)
```
gdbelvin: Would it be better to call GetUnappliedRevisions here?

core/integration/client_tests.go
```diff
@@ -523,8 +519,6 @@ func (env *Env) setupHistory(ctx context.Context, directory *pb.Directory, userI
 			}
 			if _, err := env.Sequencer.RunBatch(ctx, &spb.RunBatchRequest{
```
gdbelvin: And here and the other places in this file.  Maybe a helper function is needed

pavelkalinnikov: Done, a helper was added in #1341.

core/sequencer/sequencer_api.proto
pavelkalinnikov: `outstanding_revisions` always contains a contiguous range of revisions, so this is it just to avoid waste of space. Let me see if I can detach this to a prereq PR.

core/sequencer/sequencer_api.proto
```diff
@@ -75,8 +68,29 @@ message DefineRevisionsRequest {
...
+// outstanding revisions.
+message GetUnappliedRevisionsRequest {
+  // directory_id is the directory to examine the outstanding revisions for.
+  string directory_id = 1;
+}
+
+// GetUnappliedRevisionResponse contains the set of outstanding revisions.
+message GetUnappliedRevisionsResponse {
+  // begin is the first outstanding revision (inclusive).
+  int64 begin = 1;
```
gdbelvin: Nit: these names should match the names in DefineRevisionsResponse

core/sequencer/server.go
```diff
@@ -222,72 +223,58 @@ func (s *Server) unappliedMetric(ctx context.Context, directoryID string, maxCou
...
-		MinBatch:    in.MinBatch,
-		MaxBatch:    in.MaxBatch})
+	revs, err := s.loopback.GetUnappliedRevisions(ctx,
+		&spb.GetUnappliedRevisionsRequest{DirectoryId: in.DirectoryId})
 	if err != nil {
 		return nil, err
 	}
 
-	unappliedRevisions.Set(float64(len(defResp.OutstandingRevisions)))
+	count, end := revs.End-revs.Begin, revs.End
```
gdbelvin: This changes the behavior slightly -- could this be it's own PR?

core/sequencer/server.go
pavelkalinnikov: Not relevant after another refactoring.

core/sequencer/server.go
pavelkalinnikov: Not relevant after another refactoring.

core/sequencer/server.go
pavelkalinnikov: I think the metric should be more or less the same.

core/sequencer/sequencer_api.proto
pavelkalinnikov: Got back to lists of revisions.

core/sequencer/sequencer_api.proto
pavelkalinnikov: Using revision lists again.

cmd/keytransparency-sequencer/main.go
```diff
@@ -231,11 +231,17 @@ func runSequencer(ctx context.Context, conn *grpc.ClientConn,
 	if err := signer.AddAllDirectories(ctx); err != nil {
 		glog.Errorf("runSequencer(AddAllDirectories): %v", err)
 	}
+
 	go sequencer.PeriodicallyRun(ctx, time.Tick(*refresh), func(ctx context.Context) {
 		if err := signer.AddAllDirectories(ctx); err != nil {
 			glog.Errorf("PeriodicallyRun(AddAllDirectories): %v", err)
 		}
-		if err := signer.RunBatchForAllMasterships(ctx, int32(*batchSize)); err != nil {
+		if err := signer.DefineRevisionsForAllMasterships(ctx, int32(*batchSize)); err != nil {
```
pavelkalinnikov: Should we separate this from `AddAllDirectories`, to save a bit of latency?

pavelkalinnikov: I think `AddAllDirectories` should also run with a more relaxed cadence, because directories are rarely (almost never) added.

pavelkalinnikov: #1348

core/sequencer/sequencer_api.proto
```diff
@@ -75,8 +68,23 @@ message DefineRevisionsRequest {
...
-  repeated int64 outstanding_revisions = 1;
+  // revisions is the list of newly defined revisions.
+  repeated int64 revisions = 2;
+  // Deprecated tags, do not reuse.
+  reserved 1;
+}
+
+// GetUnappliedRevisionsRequest contains information needed to query the set of
+// outstanding revisions.
+message GetUnappliedRevisionsRequest {
```
gdbelvin: `ListUnappliedRevisions`

gdbelvin: PageSize

pavelkalinnikov: Not relevant anymore, as we decided to remove lists of revisions.

pavelkalinnikov:
Since this PR has gone through many transformations/merges, I will squash it and rebase on #1349.

pavelkalinnikov:
Please take another look.

PR #1336, opened 2019-08-08
Fix GOPROXY problem

gdbelvin:
The GOPROXY has begun to return 410's for some of our dependencies.

This PR attempts to fix this by updating to fresher dependencies, more of them
being explicit versions, and therefore a higher likelihood of being
found by the proxy.


gdbelvin:
Related to google/trillian#1785 

I think Trillian still has a new pseudo-versions that are incorrect.

PR #1337, opened 2019-08-14
Fix GOPROXY problem

mhutchinson:
The GOPROXY has begun to return 410's for some of our dependencies.

This PR attempts to fix this by updating to fresher dependencies, more of them being explicit versions, and therefore a higher likelihood of being found by the proxy.

This is a clone of #1336 because Gary is on holiday and Travis failed the build of that branch so we can't submit.


mhutchinson:
This all builds fine on my machine. But I can get it to list the problematic import of blackfriday by doing the following:
go clean -cache -modcache
go get -v ./...  # This updates go.mod
go list -m all

The problem seems to be that Trillian at 1.3.0 has some dodgy dependencies, including the "github.com/russross/blackfriday/v2 v2.0.1+incompatible" (which cannot be fetched - https://proxy.golang.org/github.com/russross/blackfriday/v2/@v/v2.0.1+incompatible.info) that I'm growing so fond of. I don't think this can be worked around except without pushing a new Trillian release that includes https://github.com/google/trillian/pull/1785.

For future reference, the way I diagnosed the reason that the +incompatible dependency snuck in is via the very handy graph feature:
go mod graph | grep blackfriday

PR #1341, opened 2019-08-20
integration: Factor out helper to run batch and publish

pavelkalinnikov:


core/integration/client_tests.go
```diff
@@ -521,22 +532,14 @@ func (env *Env) setupHistory(ctx context.Context, directory *pb.Directory, userI
...
-			}
-			_, err = env.Sequencer.PublishRevisions(ctx, &spb.PublishRevisionsRequest{DirectoryId: directory.DirectoryId, Block: true})
-			if err != nil {
-				return fmt.Errorf("sequencer.PublishRevisions(%v): %v", i, err)
+			if err := runBatchAndPublish(ctx, directory.DirectoryId, 1, 1, true, env); err != nil {
+				return fmt.Errorf("runBatchAndPublish(%v): %v", i, err)
 			}
-		} else if _, err := env.Sequencer.RunBatch(ctx, &spb.RunBatchRequest{
-			DirectoryId: directory.DirectoryId,
+		} else {
```
golangcibot: elseif: can replace 'else {if cond {}}' with 'else if cond {}' (from `gocritic`)

core/integration/client_tests.go
```diff
@@ -516,30 +516,29 @@ func (env *Env) setupHistory(ctx context.Context, directory *pb.Directory, userI
...
 				return fmt.Errorf("runBatchAndPublish(empty): %v", err)
 			}
+			continue
+		}
+		u := &client.User{
+			UserID:         userID,
+			PublicKeyData:  p,
+			AuthorizedKeys: authorizedKeys,
+		}
+		cctx, cancel := context.WithTimeout(ctx, env.Timeout)
```
pavelkalinnikov: I wonder whether this context is used as intended, because:

- It is only canceled after all loop iterations are done.
- It it used only with `CreateMutation` but not the other methods.

core/integration/client_tests.go
```diff
@@ -67,23 +67,34 @@ LOA+tLe/MbwZ69SRdG6Rx92f9tbC6dz7UVsyI7vIjS+961sELA6FeR91lA==
 -----END PUBLIC KEY-----`
 )
 
+func runBatchAndPublish(ctx context.Context, dirID string, mn, mx int32, block bool, env *Env) error {
+	rbReq := &spb.RunBatchRequest{
+		DirectoryId: dirID,
+		MinBatch:    mn,
+		MaxBatch:    mx,
+	}
+	convert := func(err error, prefix string) error {
```
gdbelvin: Does https://godoc.org/google.golang.org/grpc/status#FromContextError help?

core/integration/client_tests.go
```diff
@@ -67,23 +67,34 @@ LOA+tLe/MbwZ69SRdG6Rx92f9tbC6dz7UVsyI7vIjS+961sELA6FeR91lA==
 -----END PUBLIC KEY-----`
 )
 
+func runBatchAndPublish(ctx context.Context, dirID string, mn, mx int32, block bool, env *Env) error {
```
gdbelvin: I like to put env right after context for consistency

core/integration/client_tests.go
```diff
@@ -67,23 +67,34 @@ LOA+tLe/MbwZ69SRdG6Rx92f9tbC6dz7UVsyI7vIjS+961sELA6FeR91lA==
...
+	rbReq := &spb.RunBatchRequest{
+		DirectoryId: dirID,
+		MinBatch:    mn,
+		MaxBatch:    mx,
+	}
+	convert := func(err error, prefix string) error {
+		if err != nil && err != context.Canceled && status.Code(err) != codes.Canceled {
+			return fmt.Errorf("%v: %v", prefix, err)
```
gdbelvin: To preserve the grpc error code:
```
st := status.FromErr(err)
return status.Errorf(st.Code(), "message prefix: %v", st.Message())
```

core/integration/client_tests.go
pavelkalinnikov: It looks like `dirID` is always taken from `env.Directory`. Is that right?

pavelkalinnikov: It is indeed. I removed it.

pavelkalinnikov: Done.

core/integration/client_tests.go
pavelkalinnikov: Done.

core/integration/client_tests.go
pavelkalinnikov: Nah, direct checks seem more convenient. Do you see a better way to express this?

core/integration/client_tests.go
gdbelvin: I think it got added a long time ago when some of these RPCs would hang. 
It would be nice if the lower RPC layers had a way to set a 1 minute timeout on all RPCs, but I can't think of a way to cleanly do that.

PR #1344, opened 2019-08-22
Read batch definitions from batch storage

pavelkalinnikov:
The latest batch definition should start from the highest defined batch, not the highest applied batch. 

Related to #1335


gdbelvin:
server_test.go:160: DefineRevisions(): batch 2 not found

PR #1345, opened 2019-08-23
Batch storage integration tests

gdbelvin:
Move the batch storage tests to an integration test suite.

Add test to ensure that the storage layer returns codes.NotFound for missing rows

### DIFF
--- a/core/integration/storagetest/batch.go
+++ b/core/integration/storagetest/batch.go
@@ -4,9 +4,12 @@ import (
 	"context"
 	"testing"
 
-	spb "github.com/google/keytransparency/core/sequencer/sequencer_go_proto"
+	"github.com/golang/protobuf/proto"
+	"github.com/google/go-cmp/cmp"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+
+	spb "github.com/google/keytransparency/core/sequencer/sequencer_go_proto"
 )
 
 // Batcher writes batch definitions to storage.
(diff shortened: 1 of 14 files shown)


core/integration/storagetest/batch.go
```diff
@@ -0,0 +1,128 @@
...
+		"TestReadBatch":  b.TestReadBatch,
+		"TestHighestRev": b.TestHighestRev,
+	} {
+		ms := storageFactory(ctx, t)
+		t.Run(name, func(t *testing.T) { f(ctx, t, ms) })
+	}
+}
+
+// BatchTests is a suite of tests to run against
+type BatchTests struct{}
```
pavelkalinnikov: What's the purpose of this type? Kind of like a "namespace"?

gdbelvin: The purpose of this type is to allow runtime discovery of `Test*` methods. 
Unlike the compiler, Go cannot do runtime reflection on modules, only concrete types. 
I've keep the runtime reflection bits out of this PR to keep it simple.

core/integration/storagetest/batch.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/golang/protobuf/proto"
+	"github.com/google/go-cmp/cmp"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	spb "github.com/google/keytransparency/core/sequencer/sequencer_go_proto"
+)
+
+// Batcher writes batch definitions to storage.
+type Batcher interface {
```
pavelkalinnikov: Why not just reference this same interface defined in `core`? I don't see a clear benefit of copy-pasting it to yet another place, esp. to tests. The idea should be presumably testing **the** interface, not some made-up interface that happens to be a superset. Also, keeping them in sync is an extra effort.

core/integration/storagetest/batch.go
pavelkalinnikov: Is the idea to replace `RunBatchStorageTests` hand-writtent-table-driven function with a code that scans methods using reflection? Maybe add a TODO to clarify that?

core/integration/storagetest/batch.go
pavelkalinnikov: Although not having to import that package is kinda nice...

gdbelvin: I'll make it a type alias

PR #1349, opened 2019-08-30
Use pair of edge revisions instead of listing all

pavelkalinnikov:
This PR replaces the slice of unapplied revisions returned from `DefineRevisions` call with an equivalent tuple of "watermark" revisions. This removes the danger of this list growing uncontrollably, and makes it easier to change the semantic of this method in #1335.


core/sequencer/sequencer_api.proto
```diff
@@ -75,8 +75,18 @@ message DefineRevisionsRequest {
 
 // DefineRevisionResponse contains information about freshly defined revisions.
 message DefineRevisionsResponse {
-  // outsanding_revisions a list of all the defined revisions which are not yet applied.
-  repeated int64 outstanding_revisions = 1;
+  // map_revision is the current map revision that was fully applied. Any
```
gdbelvin: current / highest map revision that has been fully applied

core/sequencer/sequencer_api.proto
```diff
@@ -75,8 +75,18 @@ message DefineRevisionsRequest {
...
+  // unapplied revisions are greater than map_revision.
+  int64 map_revision = 2;
```
gdbelvin: `highest_applied_revision`?

core/sequencer/sequencer_api.proto
```diff
@@ -75,8 +75,18 @@ message DefineRevisionsRequest {
...
+  // old_defined is the last defined revision before this call. Any newly
+  // defined revisions are greater than old_defined.
+  int64 old_defined = 3;
+  // new_defined is the last defined revision after this call. No new revisions
+  // are defined if new_defined <= old_defined.
+  int64 new_defined = 4;
```
gdbelvin: "new" is slightly ambiguous.  Is it the number of new defined revisions or is it the new version of defined? 

Would `highest_defined_revision` be clearer?

core/sequencer/server.go
```diff
@@ -274,14 +274,16 @@ func (s *Server) DefineRevisions(ctx context.Context,
...
-	if len(outstanding) > 0 {
-		return &spb.DefineRevisionsResponse{OutstandingRevisions: outstanding}, nil
+	resp := &spb.DefineRevisionsResponse{
+		MapRevision: int64(latestMapRoot.Revision),
+		OldDefined:  highestRev,
+		NewDefined:  highestRev,
+	}
+	// Don't create new revisions if there are ones waiting to be applied, or the
+	// highest defined revision is lagging behind for some reason.
+	if resp.OldDefined != resp.MapRevision {
```
gdbelvin: Should this logic be with respect to NewDefined?

core/sequencer/sequencer_api.proto
```diff
@@ -75,8 +75,18 @@ message DefineRevisionsRequest {
...
 // DefineRevisionResponse contains information about freshly defined revisions.
 message DefineRevisionsResponse {
-  // outsanding_revisions a list of all the defined revisions which are not yet applied.
-  repeated int64 outstanding_revisions = 1;
+  // map_revision is the current map revision that was fully applied. Any
+  // unapplied revisions are greater than map_revision.
+  int64 map_revision = 2;
+  // old_defined is the last defined revision before this call. Any newly
+  // defined revisions are greater than old_defined.
+  int64 old_defined = 3;
```
gdbelvin: I'm a bit unclear on why this field is needed. Should it be removed? 
If it is only to measure when new revisions have been added would a more direct "new_revisions" field be appropriate?

core/sequencer/sequencer_api.proto
pavelkalinnikov: How about `applied_to`?

pavelkalinnikov: Renamed to `highest_applied`. I think the "revision" word is redundant.

core/sequencer/sequencer_api.proto
pavelkalinnikov: Removed.

core/sequencer/sequencer_api.proto
pavelkalinnikov: Renamed to `highest_defined`.

core/sequencer/server.go
pavelkalinnikov: Now irrelevant.

PR #1350, opened 2019-08-30
Allow defining and applying revisions concurrently

pavelkalinnikov:


core/sequencer/server.go
```diff
@@ -236,12 +236,17 @@ func (s *Server) DefineRevisions(ctx context.Context,
...
-	if resp.HighestDefined != resp.HighestApplied {
+	// Do nothing if the highest defined revision is lagging behind for some
+	// reason. It will catch up later.
+	if resp.HighestDefined < resp.HighestApplied {
 		return resp, nil
 	}
-	// TODO(#1056): Allow going ahead to some extent.
+	// Allow at most one pending revision, which enables applying a revision and
+	// defining the next one simultaneously.
+	if resp.HighestDefined > resp.HighestApplied+1 {
```
gdbelvin: Could you pull this out into a constant or a flag? I can imagine this being a value we'd want to tweek for performance

PR #1353, opened 2019-09-11
storagetest factory now takes domainID

gdbelvin:
Some storage implementations require a domain setup step
This switch allows the storage factory to do the nessesary
setup for the test.

### DIFF
--- a/core/integration/storagetest/batch.go
+++ b/core/integration/storagetest/batch.go
@@ -39,8 +39,9 @@ func RunBatchStorageTests(t *testing.T, factory BatchStorageFactory) {
 type BatchTests struct{}
 
 func (*BatchTests) TestNotFound(ctx context.Context, t *testing.T, f BatchStorageFactory) {
-	b := f(ctx, t, "testnotfounddir")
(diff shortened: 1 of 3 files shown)


core/integration/storagetest/batch.go
```diff
@@ -31,24 +31,25 @@ func RunBatchStorageTests(t *testing.T, storageFactory BatchStorageFactory) {
...
 	}
 }
 
 // BatchTests is a suite of tests to run against
 type BatchTests struct{}
 
-func (*BatchTests) TestNotFound(ctx context.Context, t *testing.T, b Batcher) {
+func (*BatchTests) TestNotFound(ctx context.Context, t *testing.T, f BatchStorageFactory) {
+	b := f(ctx, t, "testnotfounddir")
 	_, err := b.ReadBatch(ctx, "nodir", 0)
```
mhutchinson: For consistency can "nodir" be the same as "testnotfounddir" string?

gdbelvin: done

PR #1354, opened 2019-09-11
Close db after environment is closed

gdbelvin:
Don't close the db before using it


gdbelvin:
I have no idea how the integration tests passed before...

PR #1360, opened 2019-09-25
Dedup EntryUpdate that contains the same data and keyset (#1359)

libinjG:
This is an even simpler change than we discussed.


core/mutator/entry/mutator.go
```diff
@@ -87,6 +87,10 @@ func ReduceFn(leaves []*pb.EntryUpdate, msgs []*pb.EntryUpdate,
 	// Filter for mutations that are valid.
 	newEntries := make([]*pb.EntryUpdate, 0, len(msgs))
 	for i, msg := range msgs {
+		// Skip if there is no change in data.
+		if len(leaves) > 0 && bytes.Equal(leaves[0].GetCommitted().GetData(), msg.GetCommitted().GetData()) {
```
mhutchinson: This worries me that it will skip updates that intentionally keep the data the same, but modify, e.g. the 'authorized_keyset' property. Do we need a deeper equality check than just the data?

libinjG: I think an additional check of the "authorized_keyset" should be sufficient. Agree?

mhutchinson: @gdbelvin - what are your thoughts on this? What aspects of a KT record are important for equality?

gdbelvin: For more correct code, do the `if len(leaves) > 0` check separately and return an error if it is false. 
This is already done for you on line 78. Consider setting a new oldCommitment value there. 

What we're doing here is defining equality for two EntryUpdates.  Perhaps we should create an explicit `Equals` method for them and check
- The commitment - as we are doing here
- AuthorizedKeyset
- Index

libinjG: I added the check for AuthorizedKeyset. The "index" should be equal already since the "reduce" is based on the index.

core/mutator/entry/mutator.go
```diff
@@ -87,6 +87,12 @@ func ReduceFn(leaves []*pb.EntryUpdate, msgs []*pb.EntryUpdate,
 	// Filter for mutations that are valid.
 	newEntries := make([]*pb.EntryUpdate, 0, len(msgs))
 	for i, msg := range msgs {
+		// Skip if there is no change in data.
+		if len(leaves) > 0 {
+			if isEqual, _ := isEntryUpdateEqual(leaves[0], msg); isEqual {
```
gdbelvin: please use `emitErr` so that errors can be surfaced.

core/mutator/entry/mutator.go
```diff
@@ -179,3 +185,20 @@ func verifyKeys(handle *keyset.Handle, data []byte, sigs [][]byte) error {
 	}
 	return mutator.ErrUnauthorized
 }
+
+// isEntryUpdateEqual checks whether two EntryUpdates (for the same index) are
+// equal. They are considered equal if they share the same data and authorized keyset.
+func isEntryUpdateEqual(e1, e2 *pb.EntryUpdate) (bool, error) {
```
gdbelvin: Please verify that index is indeed the same.

gdbelvin: Same with UserID

gdbelvin:
LGTM

PR #1364, opened 2019-10-03
Queue integration test

gdbelvin:
By moving the queue tests from `impl/` to `core/integration` we can ensure that any queue implementation meets the API contract.


core/integration/storagetest/queue.go
```diff
@@ -0,0 +1,157 @@
...
+func mustMarshal(t *testing.T, p proto.Message) []byte {
+	t.Helper()
+	b, err := proto.Marshal(p)
+	if err != nil {
+		t.Fatalf("proto.Marshal(): %v", err)
+	}
+	return b
+}
+
+func (QueueTests) TestReadLog(ctx context.Context, t *testing.T, newForTest QueueStorageFactory) {
```
golangcibot: 99-127 lines are duplicate of `core/integration/storagetest/queue.go:129-157` (from `dupl`)

core/integration/storagetest/queue.go
```diff
@@ -0,0 +1,157 @@
...
+		if err != nil {
+			t.Fatalf("ReadLog(%v): %v", tc.batchSize, err)
+		}
+		if got, want := len(rows), tc.count; got != want {
+			t.Fatalf("ReadLog(%v): len: %v, want %v", tc.batchSize, got, want)
+		}
+	}
+}
+
+func (QueueTests) TestConcurrentWrites(ctx context.Context, t *testing.T, newForTest QueueStorageFactory) {
```
golangcibot: 129-157 lines are duplicate of `core/integration/storagetest/queue.go:99-127` (from `dupl`)

core/integration/storagetest/queue.go
```diff
@@ -0,0 +1,126 @@
...
+	"google.golang.org/grpc/status"
+
+	pb "github.com/google/keytransparency/core/api/v1/keytransparency_go_proto"
+)
+
+// Queuer writes items to the queue.
+type Queuer interface {
+	keyserver.MutationLogs
+	adminserver.LogsAdmin
+}
```
mhutchinson: I'm finding this quite hard to grok. This would seem simpler to me if we split this test into two:
- one which tests the MutationLogs contracts hold up (SetWritable vs ListLogs)
- one which tests the LogsAdmin contracts hold up (Send vs ReadLog)

As it is, this has two test cases which test non-intersecting sets of methods on different interfaces, and introduces the name `Queuer` which it tests, which doesn't appear in any of the code actually under test.

gdbelvin: Good call. Separated admin and storage test suites

gdbelvin:
At the moment these test suites might look funny because they only have one test, but future PRs will add to them

core/integration/storagetest/batch.go
```diff
@@ -1,3 +1,17 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
```
mhutchinson: s/2016/2019/

core/integration/storagetest/mutation_log.go
```diff
@@ -0,0 +1,84 @@
...
+	"testing"
+	"time"
+
+	"github.com/golang/protobuf/proto"
+	"github.com/google/keytransparency/core/keyserver"
+
+	pb "github.com/google/keytransparency/core/api/v1/keytransparency_go_proto"
+)
+
+type QueueStorageFactory func(ctx context.Context, t *testing.T, dirID string, logIDs ...int64) keyserver.MutationLogs
```
mhutchinson: QueueStorageFactory can be more clearly named MutationLogsFactory I think?

Passim for other Queue stuff.

core/integration/storagetest/mutation_log.go
```diff
@@ -0,0 +1,84 @@
...
+	b, err := proto.Marshal(p)
+	if err != nil {
+		t.Fatalf("proto.Marshal(): %v", err)
+	}
+	return b
+}
+
+func (QueueTests) TestReadLog(ctx context.Context, t *testing.T, newForTest QueueStorageFactory) {
+	directoryID := "TestReadLog"
+	logID := int64(5)
```
mhutchinson: 5 is a bit of a magic number here

core/integration/storagetest/mutation_log.go
```diff
@@ -0,0 +1,84 @@
...
+func mustMarshal(t *testing.T, p proto.Message) []byte {
+	t.Helper()
+	b, err := proto.Marshal(p)
+	if err != nil {
+		t.Fatalf("proto.Marshal(): %v", err)
+	}
+	return b
+}
+
+func (QueueTests) TestReadLog(ctx context.Context, t *testing.T, newForTest QueueStorageFactory) {
```
mhutchinson: Given that these tests are performing high level conformance of some fairly extensive iceberg code, I suggest adding some documentation on what this is testing and why is reasonable. Justification: $randomEngineer doesn't know a lot about this & updates their DB from v1.2 to v1.3; the batchSize 4 test below fails and prints "ReadLog(4): len 12, want 6". Poor engineer then has no idea what the devil is going on and goes to the pub. Tests stay broken.

core/integration/storagetest/mutation_log.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+	}
+
+	for _, tc := range []struct {
+		batchSize int32
+		count     int
+	}{
+		{batchSize: 0, count: 0},
+		{batchSize: 1, count: 3},
+		{batchSize: 4, count: 6},
```
mhutchinson: I can wrap my head around why 0 => 0 and 100 => 30, but why 1=>3 and 4=>6 is difficult for me. I could understand it by digging into the code, but the TLDR here is that the docs for ReadLog don't specify enough of a contract that a test could reasonably be asserting something as a blackbox.

core/integration/storagetest/mutation_log_admin.go
```diff
@@ -0,0 +1,89 @@
...
+	"context"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/keytransparency/core/adminserver"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+type QueueAdminFactory func(ctx context.Context, t *testing.T, dirID string, logIDs ...int64) adminserver.LogsAdmin
```
mhutchinson: I think Queue should be removed here. LogsAdminFactory seems better. Passim for other Queue stuff in this file.

core/integration/storagetest/mutation_log_admin.go
```diff
@@ -0,0 +1,89 @@
...
+
+// QueueTests is a suite of tests to run against
+type QueueAdminTests struct{}
+
+func (QueueAdminTests) TestSetWritable(ctx context.Context, t *testing.T, f QueueAdminFactory) {
+	directoryID := "TestSetWritable"
+	for _, tc := range []struct {
+		desc       string
+		logIDs     []int64
+		set        map[int64]bool
```
mhutchinson: Docs on what `set` implies would be good. It looks like entries in `logIDs` that are not in here are left in their default state, and otherwise are explicitly set to writable based on this map. The fact that it's called set and is implemented as a map is confusingly ambiguous; Go often uses map[x]bool to denote a set, but I think you mean set in the active form and not the noun (even "the noun" is still be ambiguous. it's nothing to do with badgers, or hardening glue, for example).

core/integration/storagetest/mutation_log_admin.go
```diff
@@ -0,0 +1,89 @@
...
+	} {
+		t.Run(tc.desc, func(t *testing.T) {
+			m := f(ctx, t, directoryID, tc.logIDs...)
+			wantLogs := make(map[int64]bool)
+			for _, logID := range tc.wantLogIDs {
+				wantLogs[logID] = true
+			}
+
+			for logID, enabled := range tc.set {
+				if err := m.SetWritable(ctx, directoryID, logID, enabled); err != nil {
```
mhutchinson: How about a test that call SetWritable on a non-existent log?

core/integration/storagetest/mutation_log.go
gdbelvin: Fair enough.  Names should match.

core/integration/storagetest/mutation_log.go
gdbelvin: Indeed it is. I can comment it as such.  We just need a random logID .to write to and read from

core/integration/storagetest/mutation_log_admin.go
gdbelvin: I'll move this to a verb `setWritable`

core/integration/storagetest/mutation_log_admin.go
gdbelvin: Good call

core/integration/storagetest/mutation_log_admin.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+}
+
+type logsAdminTests struct{}
+
+func (logsAdminTests) TestSetWritable(ctx context.Context, t *testing.T, f LogsAdminFactory) {
+	directoryID := "TestSetWritable"
+	m := f(ctx, t, directoryID, 1)
+	if st := status.Convert(m.SetWritable(ctx, directoryID, 2, true)); st.Code() != codes.NotFound {
+		t.Errorf("SetWritable(non-existant logid): %v, want %v", st, codes.NotFound)
```
golangcibot: `existant` is a misspelling of `existent` (from `misspell`)
```suggestion

```

impl/sql/mutationstorage/queue.go
```diff
@@ -33,9 +33,16 @@ import (
 // SetWritable enables or disables new writes from going to logID.
 func (m *Mutations) SetWritable(ctx context.Context, directoryID string, logID int64, enabled bool) error {
 	glog.Errorf("mutationstorage: SetWritable(%v, %v, enabled: %v)", directoryID, logID, enabled)
-	_, err := m.db.ExecContext(ctx,
+	result, err := m.db.ExecContext(ctx,
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

gdbelvin:
PTAL

core/integration/storagetest/mutation_log.go
```diff
@@ -0,0 +1,84 @@
...
+
+	"github.com/golang/protobuf/proto"
+	"github.com/google/keytransparency/core/keyserver"
+
+	pb "github.com/google/keytransparency/core/api/v1/keytransparency_go_proto"
+)
+
+type MutationLogsFactory func(ctx context.Context, t *testing.T, dirID string, logIDs ...int64) keyserver.MutationLogs
+
+// RunMutationLogsTests runs all the queue tests against the provided storage implementation.
```
pavelkalinnikov: Please be consistent w.r.t. queues/mutations/logs etc. The name `RunMutationLogsTests` doesn't look like it's running "queue tests". Passim for the whole PR.

core/integration/storagetest/mutation_log.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+	}
+
+	for _, tc := range []struct {
+		limit int32
+		count int
+	}{
+		{limit: 0, count: 0},
+		{limit: 1, count: 3},    // We asked for 1 item, which gets us into the first batch, so we return 3 items.
+		{limit: 3, count: 3},    // We asked for 1 item, which gets us into the first batch, so we return 3 items.
```
pavelkalinnikov: `asked for 3 items`

core/integration/storagetest/mutation_log.go
```diff
@@ -0,0 +1,84 @@
...
+		{limit: 4, count: 6},    // Reading 4 items gets us into the second batch of size 3
```
pavelkalinnikov: Please add period.

core/integration/storagetest/mutation_log_admin.go
```diff
@@ -0,0 +1,95 @@
+// Copyright 2019 Google Inc. All Rights Reserved.
```
pavelkalinnikov: Maybe rename to `mutation_logs_admin.go` for consistency?

core/integration/storagetest/mutation_log_admin.go
```diff
@@ -0,0 +1,95 @@
...
+			}
+			if err != nil {
+				return
+			}
+			logs := make(map[int64]bool)
+			for _, log := range logIDs {
+				logs[log] = true
+			}
+			if got, want := logs, wantLogs; !cmp.Equal(got, want) {
+				t.Errorf("randLog(): %v, want %v", got, want)
```
pavelkalinnikov: `"ListLogs(): ..."`

core/keyserver/keyserver.go
```diff
@@ -72,11 +72,15 @@ type WriteWatermark struct {
 
 // MutationLogs provides sets of time ordered message logs.
 type MutationLogs interface {
-	// Send submits an item to a random log.
+	// Send submits groups of item atomically to a random log.
```
pavelkalinnikov: Please clarify. Does this send the whole group, or multiple sub-groups? Also, could use `mutations` instead of `items` to match the parameter name.

core/keyserver/keyserver.go
```diff
@@ -72,11 +72,15 @@ type WriteWatermark struct {
...
+	// TODO(gbelvin): Create a batch level object to make it clear that this a batch of updates.
 	Send(ctx context.Context, directoryID string, mutation ...*pb.EntryUpdate) (*WriteWatermark, error)
-	// ReadLog returns the messages in the (low, high] range stored in the specified log.
+	// ReadLog returns the messages in the (low, high] range stored in the
```
pavelkalinnikov: Please fix the double spaces. Also, my wrapper wraps this to just 3 lines.

```
	// ReadLog returns the messages in the (low, high] range stored in the
	// specified log. ReadLog always returns complete units of the original
	// batch, and will return more items than limit if needed to do so.
```

core/keyserver/keyserver.go
```diff
@@ -72,11 +72,15 @@ type WriteWatermark struct {
...
+	// specified log.  ReadLog always returns complete units of the
```
pavelkalinnikov: Please be clearer on what a "complete unit" is. Does this correspond to the batches submitted through `Send`?

impl/sql/mutationstorage/queue_test.go
```diff
@@ -20,77 +20,45 @@ import (
...
 	"github.com/google/go-cmp/cmp"
+	"github.com/google/keytransparency/core/adminserver"
+	"github.com/google/keytransparency/core/integration/storagetest"
+	"github.com/google/keytransparency/core/keyserver"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
 	pb "github.com/google/keytransparency/core/api/v1/keytransparency_go_proto"
 	_ "github.com/mattn/go-sqlite3"
 )
```
pavelkalinnikov: Consider renaming this file to "mutation_logs_test.go" or something consistent with the other names.

impl/sql/mutationstorage/queue_test.go
```diff
@@ -20,77 +20,45 @@ import (
...
 
-const directoryID = "default"
-
-func newForTest(ctx context.Context, t testing.TB, logIDs ...int64) *Mutations {
+func newForTest(ctx context.Context, t testing.TB, dirID string, logIDs ...int64) *Mutations {
 	m, err := New(newDB(t))
 	if err != nil {
-		t.Fatalf("Failed to create Mutations: %v", err)
+		t.Fatalf("Failed to create mutations: %v", err)
```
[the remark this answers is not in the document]
gdbelvin: mutation storage

core/integration/storagetest/mutation_log.go
gdbelvin: removed references to queue

core/keyserver/keyserver.go
gdbelvin: ``` // Send submits the whole group of mutations atomically to a random log.```

gdbelvin:
Thanks for being so thorough. I took another pass.

core/integration/storagetest/mutation_logs_admin.go
```diff
@@ -0,0 +1,95 @@
...
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/keytransparency/core/adminserver"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+type LogsAdminFactory func(ctx context.Context, t *testing.T, dirID string, logIDs ...int64) adminserver.LogsAdmin
+
+// RunLogsAdminTests runs all the queue admin tests against the provided storage implementation.
```
pavelkalinnikov: s/queue/logs/ ?

core/keyserver/keyserver.go
```diff
@@ -72,11 +72,15 @@ type WriteWatermark struct {
...
 // MutationLogs provides sets of time ordered message logs.
 type MutationLogs interface {
-	// Send submits an item to a random log.
+	// Send submits the whole group of mutations atomically to a random log.
+	// TODO(gbelvin): Create a batch level object to make it clear that this a batch of updates.
 	Send(ctx context.Context, directoryID string, mutation ...*pb.EntryUpdate) (*WriteWatermark, error)
-	// ReadLog returns the messages in the (low, high] range stored in the specified log.
+	// ReadLog returns the messages in the (low, high] range stored in the
+	// specified log. ReadLog always returns complete units of the original
+	// batchs sent via Send, and will return  more items than limit if
```
pavelkalinnikov:
```suggestion
	// batches sent via Send, and will return more items than limit if
```

gdbelvin:
Agreed that the API could use a revisit. 
From a principle of least surprise point of view, you're absolutely right, getting a non-deterministic amount of data back is *weird*.  An API that requests proper units of batches would be much better. But from the perspective of trying to control the throughput of the system, batches of non-deterministic size are also a problem.

PR #1358, opened 2019-09-20
Concurrency test for MutationLog

gdbelvin:
Ensure that the queue can support multiple writes at the same time. 
- New Concurrent conformance test
- Adds concurrency to the MySql implementation by retrying when encountering db locked errors.

Most interesting to me is that the concurrency test for high contention completes relatively quickly in under 2s.

PR #1366, opened 2019-10-16
Define watermarks in terms of microseconds

gdbelvin:
This helps define a better, tested interface for the logs.

Depends on #1358

### DIFF
--- a/core/integration/storagetest/mutation_logs.go
+++ b/core/integration/storagetest/mutation_logs.go
@@ -21,6 +21,8 @@ import (
 
 	"github.com/golang/protobuf/proto"
 	"github.com/google/keytransparency/core/keyserver"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 
 	pb "github.com/google/keytransparency/core/api/v1/keytransparency_go_proto"
 )
@@ -33,7 +35,8 @@ func RunMutationLogsTests(t *testing.T, factory MutationLogsFactory) {
 	b := &mutationLogsTests{}
 	for name, f := range map[string]func(ctx context.Context, t *testing.T, f MutationLogsFactory){
 		// TODO(gbelvin): Discover test methods via reflection.
-		"TestReadLog": b.TestReadLog,
+		"TestReadLog":              b.TestReadLog,
+		"TestReadLogHighWatermark": b.TestReadLogHighWatermark,
 	} {
 		t.Run(name, func(t *testing.T) { f(ctx, t, factory) })
 	}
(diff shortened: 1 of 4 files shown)


core/keyserver/keyserver.go
```diff
@@ -76,7 +76,8 @@ type MutationLogs interface {
 	// TODO(gbelvin): Create a batch level object to make it clear that this a batch of updates.
 	Send(ctx context.Context, directoryID string, mutation ...*pb.EntryUpdate) (*WriteWatermark, error)
 	// ReadLog returns the messages in the (low, high] range stored in the
-	// specified log. ReadLog always returns complete units of the original
+	// specified log. Low and high are measured in milliseconds since the
```
mhutchinson: Is this API locked in? If not, I strongly recommend using a Time object instead of int64. In my experience types > comments for APIs.

impl/sql/mutationstorage/mutation_logs.go
```diff
@@ -202,6 +206,9 @@ func (m *Mutations) HighWatermark(ctx context.Context, directoryID string, logID
 // ReadLog may return more rows than batchSize in order to fetch all the rows at a particular timestamp.
 func (m *Mutations) ReadLog(ctx context.Context, directoryID string,
 	logID, low, high int64, batchSize int32) ([]*mutator.LogMessage, error) {
+	if high > watermark(time.Now().Add(time.Minute)) {
```
mhutchinson: This feels really arbitrary. Locking the code to time.Now() to check types feels like a medicine that has some unwelcome side-effects.

core/keyserver/keyserver.go
mhutchinson: We discussed in person, but I recommend doing the migration to time.Time and then revisiting whether this PR is needed.

core/integration/storagetest/mutation_logs.go
```diff
@@ -50,6 +53,19 @@ func mustMarshal(t *testing.T, p proto.Message) []byte {
 	return b
 }
 
+// TestReadLogHighWatermark ensures that erroneous values for high watermark return errors.
+func (mutationLogsTests) TestReadLogHighWatermark(ctx context.Context, t *testing.T, newForTest MutationLogsFactory) {
+	directoryID := "TestReadLogHighWatermark"
+	logID := int64(5) // Any log ID.
+	m := newForTest(ctx, t, directoryID, logID)
+	limit := int32(2)
+	_, err := m.ReadLog(ctx, directoryID, logID, 0, time.Now().UnixNano(), limit)
```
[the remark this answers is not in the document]
gdbelvin: I'll remove this test and the associated sanity check in the implementation as requested

core/keyserver/keyserver.go
gdbelvin: Could I suggest keeping this API change for a new PR? 
The change touches quite a bit

mhutchinson: The changes in this PR are all related to the granularity of the time unit, and changing to time.Time will affect all of the same lines. If you feel strongly about this then we can make this change first, but I'm concerned that with this approach silly mistakes can happen (e.g. this comment says milliseconds but it should say microseconds). This wouldn't apply if we used time.Time.

gdbelvin:
Superseded by #1369

PR #1371, opened 2019-10-28
Use testdb in integration tests

gdbelvin:
Use the KT test database which sets the right flags.

Don't use the same flag for the database as Trillian.


impl/integration/env.go
```diff
@@ -115,14 +116,11 @@ func keyFromPEM(p string) *any.Any {
 }
 
 // NewEnv sets up common resources for tests.
-func NewEnv(ctx context.Context) (*Env, error) {
+func NewEnv(ctx context.Context, t testing.TB) (*Env, error) {
```
mhutchinson: Should `t.Helper()` be called here?

impl/sql/testdb/testdb.go
```diff
@@ -27,12 +26,13 @@ import (
 	ktsql "github.com/google/keytransparency/impl/sql"
 )
 
-var dataSourceURI = flag.String("test_mysql_uri", "root@tcp(127.0.0.1)/", "The MySQL URI to use when running tests")
+var dataSourceURI = "root@tcp(127.0.0.1)/" //  "The MySQL URI to use when running tests"
```
mhutchinson: Is it really desired to not have any flag control over this value?

gdbelvin: Options:
1) Have a KT specific flag
2) Export the flag that Trillian defines

gdbelvin: Going for option 1

impl/sql/testdb/testdb.go
```diff
@@ -27,11 +27,12 @@ import (
 	ktsql "github.com/google/keytransparency/impl/sql"
 )
 
-var dataSourceURI = flag.String("test_mysql_uri", "root@tcp(127.0.0.1)/", "The MySQL URI to use when running tests")
+var dataSourceURI = flag.String("test_kt_uri", "root@tcp(127.0.0.1)/", "The MySQL URI to use when running tests")
```
mhutchinson: flag should have `mysql` or `db` or something in it.

PR #1376, opened 2019-11-05
Move the responsibility to pick an input log from storage to the frontend

gdbelvin:
Allow clients to specify the mutation input log
This permits quality of service prioritization because
input logs are processed in low to high order.

### DIFF
--- a/core/keyserver/keyserver.go
+++ b/core/keyserver/keyserver.go
@@ -658,7 +658,8 @@ func (s *Server) BatchQueueUserUpdate(ctx context.Context, in *pb.BatchQueueUser
 	}
 	tdone()
 
-	// Save mutation to the database.
+	// Pick a random logID.  Note, this effectively picks a random QoS. See issue #1377.
+	// TODO(gbelvin): Define an explicit QoS / Load ballancing API.
 	wmLogID, err := s.randLog(ctx, directory.DirectoryID)
 	if st := status.Convert(err); st.Code() != codes.OK {
 		return nil, status.Errorf(st.Code(), "Could not pick a log to write to: %v", err)
--- a/core/keyserver/keyserver.go
+++ b/core/keyserver/keyserver.go
@@ -70,7 +70,7 @@ type MutationLogs interface {
 	// Send submits the whole group of mutations atomically to a random log.
 	// TODO(gbelvin): Create a batch level object to make it clear that this a batch of updates.
 	// Returns the logID and timestamp that the mutation batch got written at.
-	Send(ctx context.Context, directoryID string, mutation ...*pb.EntryUpdate) (int64, time.Time, error)
+	Send(ctx context.Context, directoryID string, logID int64, mutation ...*pb.EntryUpdate) (int64, time.Time, error)
 	// ReadLog returns the messages in the (low, high] range stored in the
 	// specified log. ReadLog always returns complete units of the original
 	// batches sent via Send, and will return  more items than limit if
@@ -610,6 +610,7 @@ func (s *Server) QueueEntryUpdate(ctx context.Context, in *pb.UpdateEntryRequest
 	return s.BatchQueueUserUpdate(ctx, &pb.BatchQueueUserUpdateRequest{
 		DirectoryId: in.DirectoryId,
 		Updates:     []*pb.EntryUpdate{in.EntryUpdate},
+		LogId:       in.GetLogId(),
 	})
 }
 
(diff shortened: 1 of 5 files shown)


impl/sql/mutationstorage/mutation_logs.go
```diff
@@ -66,15 +65,11 @@ func (m *Mutations) AddLogs(ctx context.Context, directoryID string, logIDs ...i
 // Send writes mutations to the leading edge (by sequence number) of the mutations table.
 // Returns the logID/watermark pair that was written, or nil if nothing was written.
 // TODO(gbelvin): Make updates a slice.
-func (m *Mutations) Send(ctx context.Context, directoryID string, updates ...*pb.EntryUpdate) (int64, time.Time, error) {
+func (m *Mutations) Send(ctx context.Context, directoryID string, logID int64, updates ...*pb.EntryUpdate) (int64, time.Time, error) {
```
pavelkalinnikov: I think this no longer needs to return the `logID`. Also needs the corresponding comment update.

core/keyserver/keyserver.go
```diff
@@ -666,6 +673,18 @@ func (s *Server) BatchQueueUserUpdate(ctx context.Context, in *pb.BatchQueueUser
 	return &empty.Empty{}, nil
 }
 
+func (s *Server) randLog(ctx context.Context, directoryID string) (int64, error) {
+	// TODO(gbelvin): Cache these results.
+	writable := true
+	logIDs, err := s.logs.ListLogs(ctx, directoryID, writable)
```
mhutchinson: What happens here if there are no writable logs? My naive (but experienced) intuition is that ListLogs will return an empty list with no error, and then panics below trying to find a random number for `rand.Intn(0)`.

core/keyserver/keyserver.go
```diff
@@ -656,7 +659,11 @@ func (s *Server) BatchQueueUserUpdate(ctx context.Context, in *pb.BatchQueueUser
 	tdone()
 
 	// Save mutation to the database.
-	wmLogID, wmTime, err := s.logs.Send(ctx, directory.DirectoryID, in.Updates...)
+	wmLogID, err := s.randLog(ctx, directory.DirectoryID)
```
mhutchinson: This is a natural place to link to #1377 - you're making a QoS decision here given the current implementation and it's worth noting that.

impl/sql/mutationstorage/mutation_logs_test.go
```diff
@@ -57,45 +56,6 @@ func TestLogsAdminIntegration(t *testing.T) {
 		})
 }
 
-func TestRandLog(t *testing.T) {
```
mhutchinson: I feel like the replacement to this test ought to be something that confirms the entry was written to the log you specified? If you reintroduced the randomness in the mutationstorage layer and ignored the specified logID, would any test fail?

gdbelvin: I think that case is already covered by the storage layer integration tests: 
core/integration/storagetest/mutation_logs.go

core/keyserver/keyserver.go
gdbelvin: ListLogs will return codes.NotFound when there are no writable logs:
core/integration/storagetest/mutation_logs_admin.go

core/keyserver/keyserver.go
mhutchinson: Please address this, otherwise approval.

core/keyserver/revisions_test.go
```diff
@@ -71,8 +71,16 @@ func (b batchStorage) ReadBatch(ctx context.Context, dirID string, rev int64) (*
...
 type mutations map[int64][]*mutator.LogMessage // Map of logID to Slice of LogMessages
 
-func (m *mutations) Send(ctx context.Context, dirID string, mutation ...*pb.EntryUpdate) (int64, time.Time, error) {
-	return 0, time.Time{}, errors.New("unimplemented")
+func (m *mutations) Send(ctx context.Context, dirID string, _ int64, mutation ...*pb.EntryUpdate) (time.Time, error) {
+	return time.Time{}, errors.New("unimplemented")
+}
+
+func (m *mutations) ListLogs(ctx context.Context, dirID string, _ bool) ([]int64, error) {
+	logIDs := []int64{}
```
pavelkalinnikov: Could pre-allocate `len(*m)` items here.

core/keyserver/keyserver.go
```diff
@@ -666,6 +674,18 @@ func (s *Server) BatchQueueUserUpdate(ctx context.Context, in *pb.BatchQueueUser
 	return &empty.Empty{}, nil
 }
 
+func (s *Server) randLog(ctx context.Context, directoryID string) (int64, error) {
+	// TODO(gbelvin): Cache these results.
```
pavelkalinnikov: How badly does this read harm the write performance?

PR #1378, opened 2019-11-06
Do pagination the right way

gdbelvin:
Pagination should not be modifying the continuation tokens. 
They are dangerous and not for consumption.

### DIFF
--- a/core/sequencer/server.go
+++ b/core/sequencer/server.go
@@ -132,9 +132,7 @@ type LogsReader interface {
 	// or all logIDs associated with directoryID if writable is false.
 	ListLogs(ctx context.Context, directoryID string, writable bool) ([]int64, error)
 
-	// ReadLog returns the lowest messages in the (low, high] range stored in the
-	// specified log, up to batchSize.  Paginate by setting low to the
-	// highest LogMessage returned in the previous page.
+	// ReadLog returns the lowest messages in the (low, high] range stored in the specified log, up to batchSize.
 	ReadLog(ctx context.Context, directoryID string, logID int64, low, high time.Time,
 		batchSize int32) ([]*mutator.LogMessage, error)
 }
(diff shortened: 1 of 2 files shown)


core/sequencer/server.go
```diff
@@ -343,21 +341,22 @@ func (s *Server) readMessages(ctx context.Context, source *spb.MapMetadata_Sourc
 	ss := metadata.Source(source)
 	low := ss.StartTime()
 	high := ss.EndTime()
-	// Loop until less than chunkSize items are returned.
-	for count := chunkSize; count == chunkSize; {
-		batch, err := s.logs.ReadLog(ctx, directoryID, source.LogId, low, high, chunkSize)
+	for moreToRead := true; moreToRead; {
+		// Request one more item than chunksize so we can find the next page token
```
mhutchinson: nits:
1) chunk**S**ize
2) Period at end of comments

core/sequencer/server.go
```diff
@@ -343,21 +341,22 @@ func (s *Server) readMessages(ctx context.Context, source *spb.MapMetadata_Sourc
...
 			return fmt.Errorf("logs.ReadLog(): %v", err)
 		}
-		count = int32(len(batch))
 		glog.Infof("ReadLog(dir: %v log: %v, (%v, %v], %v) count: %v",
-			directoryID, source.LogId, low, high, chunkSize, count)
+			directoryID, source.LogId, low, high, chunkSize, len(batch))
+		moreToRead = int32(len(batch)) == (chunkSize + 1)
+		if moreToRead {
+			low = batch[chunkSize].ID  // Use the last row as the start of the next read.
+			batch = batch[0:chunkSize] // Don't emit the next page token.
```
pavelkalinnikov: `batch[:chunkSize]`

core/sequencer/server.go
```diff
@@ -343,21 +341,22 @@ func (s *Server) readMessages(ctx context.Context, source *spb.MapMetadata_Sourc
 	ss := metadata.Source(source)
 	low := ss.StartTime()
 	high := ss.EndTime()
-	// Loop until less than chunkSize items are returned.
-	for count := chunkSize; count == chunkSize; {
-		batch, err := s.logs.ReadLog(ctx, directoryID, source.LogId, low, high, chunkSize)
+	for moreToRead := true; moreToRead; {
+		// Request one more item than chunksize so we can find the next page token
+		batch, err := s.logs.ReadLog(ctx, directoryID, source.LogId, low, high, chunkSize+1)
```
mhutchinson: It would feel more honest to me if this queried using `chunkSize` without the `+1`. I think the logic still checks out (assuming you don't add the 1 in the check below either).

Reasoning: I assume chunkSize is picked based on what the DB can handle. In the case of the SQL optimizer fun we had recently, I would be somewhat upset if I picked a number under that limit, but then the code between me and the DB started incrementing that size for $reasons, and triggered the issue.

core/sequencer/server.go
pavelkalinnikov: chunkSize

pavelkalinnikov: period

core/sequencer/server.go
pavelkalinnikov: Can you double check `ReadLog`'s inclusion/exclusion semantics? The current comment says `(low, high]`, which means you don't need to read one extra item to figure out the next `low`, just set it to be the last ID. Moreover, setting `low` to `chunkSize+1`th ID makes the next `ReadLog` to skip it.

gdbelvin: The comment was incorrect. It should be `[low, high)`

gdbelvin: I'm not sure chunkSize - 1 would be much better, since it causes issues around the lower end.  

The +1, -1 semantics are bound up in figuring out whether there are more items to read, to distinguish the  difference between an exact batch and there being more items to read.

PR #1380, opened 2019-11-07
Support and test mutation log queries at intermediate timestamps

gdbelvin:
Standardize on quantums of time.Microsecond

### DIFF
--- a/core/integration/storagetest/mutation_logs.go
+++ b/core/integration/storagetest/mutation_logs.go
@@ -16,6 +16,7 @@ package storagetest
 
 import (
 	"context"
+	"fmt"
 	"testing"
 	"time"
 
@@ -70,7 +71,7 @@ func (mutationLogsTests) TestReadLog(ctx context.Context, t *testing.T, newForTe
 		}
 	}
 
-	for _, tc := range []struct {
+	for i, tc := range []struct {
 		limit int32
 		want  int
 	}{
@@ -80,13 +81,15 @@ func (mutationLogsTests) TestReadLog(ctx context.Context, t *testing.T, newForTe
 		{limit: 4, want: 6},    // Reading 4 items gets us into the second batch of size 3.
 		{limit: 100, want: 30}, // Reading all the items gets us the 30 items we wrote.
 	} {
-		rows, err := m.ReadLog(ctx, directoryID, logID, minWatermark, time.Now(), tc.limit)
-		if err != nil {
-			t.Fatalf("ReadLog(%v): %v", tc.limit, err)
-		}
-		if got := len(rows); got != tc.want {
-			t.Fatalf("ReadLog(%v): len: %v, want %v", tc.limit, got, tc.want)
-		}
+		t.Run(fmt.Sprintf("%d", i), func(t *testing.T) {
+			rows, err := m.ReadLog(ctx, directoryID, logID, minWatermark, time.Now(), tc.limit)
+			if err != nil {
+				t.Fatalf("ReadLog(%v): %v", tc.limit, err)
+			}
+			if got := len(rows); got != tc.want {
+				t.Fatalf("ReadLog(%v): len: %v, want %v", tc.limit, got, tc.want)
+			}
+		})
 	}
 }
 
@@ -107,7 +110,7 @@ func (mutationLogsTests) TestReadLogExact(ctx context.Context, t *testing.T, new
 		idx = append(idx, ts)
 	}
 
-	for _, tc := range []struct {
+	for i, tc := range []struct {
 		low, high time.Time
 		want      []byte
 	}{
@@ -119,17 +122,19 @@ func (mutationLogsTests) TestReadLogExact(ctx context.Context, t *testing.T, new
 		{low: idx[0].Add(1), high: idx[9], want: []byte{1, 2, 3, 4, 5, 6, 7, 8}},
 		{low: idx[0].Add(1), high: idx[9].Add(1), want: []byte{1, 2, 3, 4, 5, 6, 7, 8, 9}},
 	} {
-		rows, err := m.ReadLog(ctx, directoryID, logID, tc.low, tc.high, 100)
-		if err != nil {
-			t.Fatalf("ReadLog(): %v", err)
-		}
-		got := make([]byte, 0, len(rows))
-		for _, r := range rows {
-			i := r.Mutation.Entry[0]
-			got = append(got, i)
-		}
-		if !cmp.Equal(got, tc.want) {
-			t.Fatalf("ReadLog(%v,%v): got %v, want %v", tc.low, tc.high, got, tc.want)
-		}
+		t.Run(fmt.Sprintf("%d", i), func(t *testing.T) {
+			rows, err := m.ReadLog(ctx, directoryID, logID, tc.low, tc.high, 100)
+			if err != nil {
+				t.Fatalf("ReadLog(): %v", err)
+			}
+			got := make([]byte, 0, len(rows))
+			for _, r := range rows {
+				i := r.Mutation.Entry[0]
+				got = append(got, i)
+			}
+			if !cmp.Equal(got, tc.want) {
+				t.Fatalf("ReadLog(%v,%v): got %v, want %v", tc.low, tc.high, got, tc.want)
+			}
+		})
 	}
 }
--- a/core/integration/storagetest/mutation_logs.go
+++ b/core/integration/storagetest/mutation_logs.go
@@ -108,15 +108,18 @@ func (mutationLogsTests) TestReadLogExact(ctx context.Context, t *testing.T, new
 	}
 
 	for _, tc := range []struct {
-		low, high int
+		low, high time.Time
 		want      []byte
 	}{
-		{low: 0, high: 0, want: []byte{}},
-		{low: 0, high: 1, want: []byte{0}},
-		{low: 0, high: 9, want: []byte{0, 1, 2, 3, 4, 5, 6, 7, 8}},
-		{low: 1, high: 9, want: []byte{1, 2, 3, 4, 5, 6, 7, 8}},
+		{low: idx[0], high: idx[0], want: []byte{}},
+		{low: idx[0], high: idx[1], want: []byte{0}},
+		{low: idx[0], high: idx[9], want: []byte{0, 1, 2, 3, 4, 5, 6, 7, 8}},
+		{low: idx[1], high: idx[9], want: []byte{1, 2, 3, 4, 5, 6, 7, 8}},
+		// Ensure that adding 1 correctly modifies the range semantics.
+		{low: idx[0].Add(1), high: idx[9], want: []byte{1, 2, 3, 4, 5, 6, 7, 8}},
+		{low: idx[0].Add(1), high: idx[9].Add(1), want: []byte{1, 2, 3, 4, 5, 6, 7, 8, 9}},
 	} {
-		rows, err := m.ReadLog(ctx, directoryID, logID, idx[tc.low], idx[tc.high], 100)
+		rows, err := m.ReadLog(ctx, directoryID, logID, tc.low, tc.high, 100)
 		if err != nil {
 			t.Fatalf("ReadLog(): %v", err)
 		}
--- a/core/integration/storagetest/mutation_logs_reader.go
+++ b/core/integration/storagetest/mutation_logs_reader.go
@@ -31,6 +31,8 @@ type LogsReadWriter interface {
 	// Send submits the whole group of mutations atomically to a log.
 	// TODO(gbelvin): Create a batch level object to make it clear that this a batch of updates.
 	// Returns the timestamp that the mutation batch got written at.
+	// This timestamp can be used as an argument to the lower bound of ReadLog [low, high).
+	// To acquire a value to  use  for the  upper bound of ReadLog [low, high), use HighWatermark.
 	Send(ctx context.Context, directoryID string, logID int64, mutation ...*pb.EntryUpdate) (time.Time, error)
 }
 
@@ -44,7 +46,9 @@ func RunMutationLogsReaderTests(t *testing.T, factory logsRWFactory) {
 	type TestFunc func(ctx context.Context, t *testing.T, f logsRWFactory)
 	for name, f := range map[string]TestFunc{
 		// TODO(gbelvin): Discover test methods via reflection.
-		"TestHighWatermark": b.TestHighWatermark,
+		"TestHighWatermarkPreserve": b.TestHighWatermarkPreserve,
+		"TestHighWatermarkRead":     b.TestHighWatermarkRead,
+		"TestHighWatermarkBatch":    b.TestHighWatermarkBatch,
 	} {
 		t.Run(name, func(t *testing.T) { f(ctx, t, factory) })
 	}
@@ -52,26 +56,18 @@ func RunMutationLogsReaderTests(t *testing.T, factory logsRWFactory) {
 
 type mutationLogsReaderTests struct{}
 
-// TestHighWatermark ensures that reads respect the low inclusive, high exclusive API.
-// nolint:gocyclo
-func (mutationLogsReaderTests) TestHighWatermark(ctx context.Context, t *testing.T, newForTest logsRWFactory) {
-	directoryID := "TestHighWatermark"
-	logID := int64(1)
-	m, done := newForTest(ctx, t, directoryID, logID)
-	defer done(ctx)
-
+func setupWatermarks(ctx context.Context, t *testing.T, m LogsReadWriter, dirID string, logID int64, maxIndex int) ([]time.Time, []time.Time) {
+	t.Helper()
 	// Setup the test by writing 10 items to the mutation log and
 	// collecting the reported high water mark after each write.
-	sent := []time.Time{} // Timesttamps that Send reported.
+	sent := []time.Time{} // Timestamps that Send reported.
 	hwm := []time.Time{}  // High water marks collected after each Send.
-	maxIndex := 9
 	for i := 0; i <= maxIndex; i++ {
-		ts, err := m.Send(ctx, directoryID, logID,
-			&pb.EntryUpdate{Mutation: &pb.SignedEntry{Entry: []byte{byte(i)}}})
+		ts, err := m.Send(ctx, dirID, logID, &pb.EntryUpdate{Mutation: &pb.SignedEntry{Entry: []byte{byte(i)}}})
 		if err != nil {
 			t.Fatalf("Send(%v): %v", logID, err)
 		}
-		count, wm, err := m.HighWatermark(ctx, directoryID, logID, minWatermark, 100 /*batchSize*/)
+		count, wm, err := m.HighWatermark(ctx, dirID, logID, minWatermark, 100 /*batchSize*/)
 		if err != nil {
 			t.Fatalf("HighWatermark(): %v", err)
 		}
@@ -81,9 +77,19 @@ func (mutationLogsReaderTests) TestHighWatermark(ctx context.Context, t *testing
 		sent = append(sent, ts)
 		hwm = append(hwm, wm)
 	}
+	return sent, hwm
+}
+
+// Tests that query HighWatermarks with varying parameters and validate results directly.
+func (mutationLogsReaderTests) TestHighWatermarkPreserve(ctx context.Context, t *testing.T, newForTest logsRWFactory) {
+	directoryID := "TestHighWatermarkPreserve"
+	logID := int64(1)
+	m, done := newForTest(ctx, t, directoryID, logID)
+	defer done(ctx)
+	maxIndex := 9
+	sent, _ := setupWatermarks(ctx, t, m, directoryID, logID, maxIndex)
 
 	arbitraryTime := time.Date(1, 2, 3, 4, 5, 6, 7, time.UTC)
-	// Tests that query HighWatermarks with varying parameters and validate results directly.
 	for _, tc := range []struct {
 		desc  string
 		start time.Time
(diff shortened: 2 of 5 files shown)


core/integration/storagetest/mutation_logs_reader.go
```diff
@@ -0,0 +1,100 @@
...
+	"context"
+	"testing"
+	"time"
+
+	"github.com/google/keytransparency/core/sequencer"
+
+	pb "github.com/google/keytransparency/core/api/v1/keytransparency_go_proto"
+)
+
+type LogsReader interface {
```
pavelkalinnikov: Please add a type comment.

core/integration/storagetest/mutation_logs_reader.go
```diff
@@ -0,0 +1,100 @@
...
+	sequencer.LogsReader
+	// Send submits the whole group of mutations atomically to a random log.
```
pavelkalinnikov: Please update the comment (the "random" bit, and the return values).

core/integration/storagetest/mutation_logs_reader.go
```diff
@@ -0,0 +1,100 @@
...
+	// TODO(gbelvin): Create a batch level object to make it clear that this a batch of updates.
+	// Returns the logID and timestamp that the mutation batch got written at.
+	Send(ctx context.Context, directoryID string, logID int64, mutation ...*pb.EntryUpdate) (time.Time, error)
+}
+
+// mutationLogsReaderFactory returns a new database object, and a function for cleaning it up.
+type mutationLogsReaderFactory func(ctx context.Context, t *testing.T, dirID string, logIDs ...int64) (LogsReader, func(context.Context))
```
pavelkalinnikov: Maybe shorten to 'logsReaderFactory`, since it generates `LogsReader` objects?

core/integration/storagetest/mutation_logs_reader.go
```diff
@@ -0,0 +1,100 @@
...
+
+// RunMutationLogsReaderTests runs all the tests against the provided storage implementation.
+func RunMutationLogsReaderTests(t *testing.T, factory mutationLogsReaderFactory) {
+	ctx := context.Background()
+	b := &mutationLogsReaderTests{}
+	for name, f := range map[string]func(ctx context.Context, t *testing.T, f mutationLogsReaderFactory){
```
pavelkalinnikov: nit: Maybe define the `func` type 1 line above, for easier readability?

core/integration/storagetest/mutation_logs_reader.go
```diff
@@ -0,0 +1,100 @@
...
+		// TODO(gbelvin): Discover test methods via reflection.
+		"TestHighWatermark": b.TestHighWatermark,
+	} {
+		t.Run(name, func(t *testing.T) { f(ctx, t, factory) })
+	}
+}
+
+type mutationLogsReaderTests struct{}
+
+// TestReadLogExact ensures that reads respect the low inclusive, high exclusive API.
```
pavelkalinnikov: Please sync the comment and the function name (and `directoryID` if the name changes).

impl/memory/mutation_logs.go
```diff
@@ -119,7 +119,7 @@ func (m MutationLogs) HighWatermark(_ context.Context, _ string, logID int64, st
 	count := int32(0)
 	high := start // Preserve start time if there are no rows to process.
 	for ; i < len(logShard) && count < batchSize; i++ {
-		high = logShard[i].time.Add(time.Nanosecond) // Returns exclusive n + 1
+		high = logShard[i].time.Add(time.Microsecond) // Returns exclusive n + 1
```
pavelkalinnikov: My first reaction to this line is "Why? The idea was that we just want the minimal watermark that's greater". Then I understood that you changed it to a microsecond because one of the storage implementations uses the microsecond granularity, and so +1 nanosecond would render "equal" because of the time-to-microsecond conversion error.

This is the 3rd reason to not use `time.Time`! Pleeeease. The idea why to use `time.Time` was that it would be independent from storage. The reality is that it's not. What if at some point another storage implementation wants a millisecond granularity? Would you convert all the code to add a millisecond instead of a microsecond? All we want is an abstract int, and do an abstract +1 here.

core/sequencer/server_test.go
```diff
@@ -229,6 +229,7 @@ func TestHighWatermarks(t *testing.T) {
 	fakeLogs, idx := setupLogs(ctx, t, dirID, map[int64]int{0: 10, 1: 20})
 	s := Server{logs: fakeLogs}
 
+	quantum := time.Microsecond
```
pavelkalinnikov: Would these tests break if it was a `Nanosecond`?

gdbelvin: The mysql implementation returns `high` + 1 microsecond, the memory implementation returns `high` + 1 nanosecond.  This PR standardized on microsecond since it was the higher granularity.  I'd like to remove this complexity though.

impl/memory/mutation_logs.go
gdbelvin: The reason for this API leakage is perhaps an ugly hack buried inside of the Highwatermarks function -- Let's discuss that? 

Basically, Highwatermarks *should* return the highest primary key from the batch table, but because we've defined the ReadLog's API in terms of `[low, high)`,  we need to return a high watermark that's *just* beyond the primary key of the highest item.

pavelkalinnikov: So, there are options:

1. `ReadLog` request is `[low, high)`. And then the `HighWatermarks` returns watermarks "just beyond" the last one, because these are going to be used as `low` in follow-up requests.
2. `ReadLog` request is in `(low, high]` range. `HighWatermarks` simply returns the last ID, and it can be used as a follow-up `low`. Cons: a) we can't `ReadLog` at position 0 in this case; b) the logs are sorted/read left-to-right, so `[low, high)` seems more intuitive.
3. Currently: `ReadLog` is in `[low, high)`, `HighWatermarks` returns the last ID, and the application layer does "logical" +1 (which tries to satisfy all existing storage implementations).
4. `ReadLog` is in `[low, high)`, `HighWatermarks` returns the last ID, and the application layer does "physical" +1 (independent of storage implementations; needs watermarks being ints or `Watermark struct` with a defined increment operation).

I think (1) is the preferred option, and if we choose it then the question of having `Watermark struct` is orthogonal (but still strongly voted for by me).

core/integration/storagetest/mutation_logs_reader.go
gdbelvin: I'm not sure why the linter doesn't pick this up...

impl/memory/mutation_logs.go
gdbelvin: I've come up with a way to go with option 1, and also stick with +1

gdbelvin:
Ok, sooo...

I've removed all concepts of what "quantum" the storage layer has from tests, or anywhere else outside of the storage implementation itself.  I think we've arrived at a place where the code in `core` can treat watermarks as opaque objects that it does NOT modify. 

In order to have consistent behavior for the HighWatermarks test across storage implementations, I've settled on returning a timestamp "just beyond" the primary key of the highest item. Since we're using `time.Time`, it makes sense to use `Add(1)` regardless of the time fidelity of the storage layer. 

It is now the storage layer's job to round timestamp inputs up to whatever quantum the storage layer itself decides to use. This preserves all the necessary semantics and (very helpfully) we now have tests that verify this behavior.

impl/sql/mutationstorage/mutation_logs.go
```diff
@@ -163,6 +165,7 @@ func (m *Mutations) send(ctx context.Context, ts time.Time, directoryID string,
 // equal to batchSize items greater than start.
 func (m *Mutations) HighWatermark(ctx context.Context, directoryID string, logID int64,
 	start time.Time, batchSize int32) (int32, time.Time, error) {
+	startQuery := start.Add(quantum - 1).Truncate(quantum)
```
pavelkalinnikov: This code feels very fragile to me, akin to using `float` when you need the precision of `int`. I know it probably works, and tests pass, but I don't feel comfortable when I read it.

core/integration/storagetest/mutation_logs_reader.go
```diff
@@ -0,0 +1,100 @@
...
+	m, done := newForTest(ctx, t, directoryID, logIDs...)
+	defer done(ctx)
+	update := &pb.EntryUpdate{}
+
+	idx := []time.Time{}
+	for i := 0; i < 10; i++ {
+		logID := int64(1)
+		ts, err := m.Send(ctx, directoryID, logID, update)
+		if err != nil {
+			t.Fatalf("m.send(%v): %v", logID, err)
```
pavelkalinnikov: `Send`

impl/sql/mutationstorage/mutation_logs.go
```diff
@@ -173,21 +176,25 @@ func (m *Mutations) HighWatermark(ctx context.Context, directoryID string, logID
 			ORDER BY Q.Time ASC
 			LIMIT ?
 		) AS T1`,
-		directoryID, logID, start, batchSize).
+		directoryID, logID, startQuery, batchSize).
```
pavelkalinnikov: I'm surprized the SQL query takes `time.Time` as a parameter. My understanding was that it was a microseconds int.

Have you considered not using the `DATETIME` type? There is much [details](https://dev.mysql.com/doc/refman/8.0/en/datetime.html) on how it works, which, together with details on how `time.Time` works, make me nervous about missing some unexpected corner cases like leap years/seconds or timezones, or keeping in mind this implicit "microsecond quantum" quirk.

Is MySQL `DATETIME` type one of the reasons you decided to use `time.Time`? I wonder if we could avoid this complexity as follows:

- Use `BIGINT` MySQL type for clarity
- Use `time.Now().UnixNano()` timestamp when writing
- Do not convert it back to `time.Time` when reading, just continue using the int
- Use `Watermark` type to wrap this int in a type-safe manner

gdbelvin: The big motivation for all these changes was managing the discrepancy between the mysql implementation (which used to be BIGINT with UnixNanos), and Spanner, which uses microseconds. These were two very different kinds of "ints", which caused their own problems.  Trying to abstract away the time precision of the storage layer is how we got to time.Time. Migrating to time.Time has also produced a whole new suite of tests to ensure that storage implementations implement these APIs correctly, and have removed quite a few sneaky modifications of these watermarks in the code.

I think your concerns are very legitimate and worth discussing further.

gdbelvin:
An interesting side effect of this PR is that is *requires* watermarks to have nanosecond precision. 
Just thought I'd point that out for posterity.

core/integration/storagetest/mutation_logs_reader.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/google/keytransparency/core/sequencer"
+
+	pb "github.com/google/keytransparency/core/api/v1/keytransparency_go_proto"
+)
+
+// LogsReadWriter supports test's ability to write to and read from the mutation logs.
+type LogsReadWriter interface {
+	sequencer.LogsReader
+	// Send submits the whole group of mutations atomically to a log.
+	// TODO(gbelvin): Create a batch level object to make it clear that this a batch of updates.
```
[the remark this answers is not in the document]
gdbelvin: I think the idea was to create a distinct type to hold batches of mutations. 
This would make the ReadLogs API much clearer since we could return []Batches rather than having the semi-awkard current situation of returning more than `batchSize` of mutations. 

@mhutchinson -- your thoughts?

impl/sql/mutationstorage/mutation_logs.go
gdbelvin: See discussion below

mhutchinson:
@gdbelvin is this ready for review now, or are planning on making the other changes regarding testing semantics for high watermark?

core/integration/storagetest/mutation_logs_reader.go
```diff
@@ -0,0 +1,176 @@
...
+		})
+	}
+
+	// Tests that use the watermarks defined during setup.
+	for _, tc := range []struct {
+		desc     string
+		readHigh time.Time
+		want     []byte
+	}{
+		// Verify that highwatermark can retreive all the data written so far.
```
golangcibot: `retreive` is a misspelling of `retrieve` (from `misspell`)
```suggestion

```

core/integration/storagetest/mutation_logs_reader.go
```diff
@@ -0,0 +1,176 @@
...
+		desc  string
+		start time.Time
+		batch int32
+		want  []byte
+	}{
+		// Verify that limiting batch size controls the number of items returned.
+		{desc: "limit batch", start: sent[0], batch: 2, want: []byte{0, 1}},
+		// Verify that advancing start by 1 with the same batch size advances the results by one.
+		{desc: "start 1", start: sent[1], batch: 2, want: []byte{1, 2}},
+		// Verify that watermarks inbetween primary keys resolve correctly.
```
golangcibot: `inbetween` is a misspelling of `between` (from `misspell`)
```suggestion

```

core/integration/storagetest/mutation_logs_reader.go
```diff
@@ -0,0 +1,176 @@
...
+		"TestHighWatermark": b.TestHighWatermark,
+	} {
+		t.Run(name, func(t *testing.T) { f(ctx, t, factory) })
+	}
+}
+
+type mutationLogsReaderTests struct{}
+
+// TestHighWatermark ensures that reads respect the low inclusive, high exclusive API.
+func (mutationLogsReaderTests) TestHighWatermark(ctx context.Context, t *testing.T, newForTest logsRWFactory) {
```
golangcibot: cyclomatic complexity 19 of func `(mutationLogsReaderTests).TestHighWatermark` is high (> 18) (from `gocyclo`)

core/integration/storagetest/mutation_logs_reader.go
```diff
@@ -0,0 +1,176 @@
...
+			for _, r := range rows {
+				i := r.Mutation.Entry[0]
+				got = append(got, i)
+			}
+			if !cmp.Equal(got, tc.want) {
+				t.Fatalf("ReadLog(%v,%v): got %v, want %v", tc.start, wm, got, tc.want)
+			}
+		})
+	}
+
```
golangcibot: unnecessary trailing newline (from `whitespace`)

gdbelvin:
`TestHighWatermark` has been reworked to avoid any dependency on the time fidelity of the storage implementation. 

This also has the happy side effect of not imposing a particular time fidelity on the batch definition table. (Previous versions of this PR were going to force nanosecond fidelity)

core/integration/storagetest/mutation_logs.go
```diff
@@ -80,13 +81,15 @@ func (mutationLogsTests) TestReadLog(ctx context.Context, t *testing.T, newForTe
...
 		{limit: 100, want: 30}, // Reading all the items gets us the 30 items we wrote.
 	} {
-		rows, err := m.ReadLog(ctx, directoryID, logID, minWatermark, time.Now(), tc.limit)
-		if err != nil {
-			t.Fatalf("ReadLog(%v): %v", tc.limit, err)
-		}
-		if got := len(rows); got != tc.want {
-			t.Fatalf("ReadLog(%v): len: %v, want %v", tc.limit, got, tc.want)
-		}
+		t.Run(fmt.Sprintf("%d", i), func(t *testing.T) {
```
mhutchinson: I feel like I saw Pavel use an empty string instead of the `fmt.Sprintf` description and the framework automatically generates a description?

core/integration/storagetest/mutation_logs_reader.go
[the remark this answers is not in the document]
mhutchinson: If a batch is a first-class object then it seems subtle and indirect that it is constructed by passing in a bunch of `EntryUpdate`s at the same time. This feels like it should be a slice at a minimum.

For the purposes of just this method I wouldn't go beyond turning the varargs `mutation` into a slice, but an API shouldn't be designed a method at a time so I'd need to re-review the whole API & future plans to make a call on the situation.

core/integration/storagetest/mutation_logs_reader.go
```diff
@@ -0,0 +1,177 @@
...
+	pb "github.com/google/keytransparency/core/api/v1/keytransparency_go_proto"
+)
+
+// LogsReadWriter supports test's ability to write to and read from the mutation logs.
+type LogsReadWriter interface {
+	sequencer.LogsReader
+	// Send submits the whole group of mutations atomically to a log.
+	// TODO(gbelvin): Create a batch level object to make it clear that this a batch of updates.
+	// Returns the timestamp that the mutation batch got written at.
```
mhutchinson: I think a note on this returned value about whether it can be used directly in a query, or whether it needs 1 adding to it will help put a stake in the ground about this topic.

core/integration/storagetest/mutation_logs_reader.go
```diff
@@ -0,0 +1,177 @@
...
+// nolint:gocyclo
+func (mutationLogsReaderTests) TestHighWatermark(ctx context.Context, t *testing.T, newForTest logsRWFactory) {
+	directoryID := "TestHighWatermark"
+	logID := int64(1)
+	m, done := newForTest(ctx, t, directoryID, logID)
+	defer done(ctx)
+
+	// Setup the test by writing 10 items to the mutation log and
+	// collecting the reported high water mark after each write.
+	sent := []time.Time{} // Timesttamps that Send reported.
```
mhutchinson: tyop in comment

core/integration/storagetest/mutation_logs_reader.go
```diff
@@ -0,0 +1,177 @@
...
+			if !got.Equal(tc.want) {
+				t.Errorf("HighWatermark(%v, %v) high: %v, want %v", tc.start, tc.batch, got, tc.want)
+			}
+			if count != 0 {
+				t.Errorf("HighWatermark(): count %v, want 0", count)
+			}
+		})
+	}
+
+	// Tests that use the watermarks defined during setup.
```
mhutchinson: It feels like these different test blocks should be different tests. They share the same setup but the blocks are orthogonal.

impl/sql/mutationstorage/mutation_logs.go
```diff
@@ -28,6 +28,10 @@ import (
 	pb "github.com/google/keytransparency/core/api/v1/keytransparency_go_proto"
 )
 
+// quantum is the fidelity of the Timestamp column.
+// See https://dev.mysql.com/doc/refman/8.0/en/fractional-seconds.html
```
mhutchinson: Async: Should `impl/sql` in the directory really be `impl/mysql`? SQL != MySQL, and this may be a trip hazard in the future.

impl/sql/mutationstorage/mutation_logs.go
```diff
@@ -163,6 +164,7 @@ func (m *Mutations) send(ctx context.Context, ts time.Time, directoryID string,
 // equal to batchSize items greater than start.
 func (m *Mutations) HighWatermark(ctx context.Context, directoryID string, logID int64,
 	start time.Time, batchSize int32) (int32, time.Time, error) {
+	startQuery := start.Add(quantum - 1).Truncate(quantum)
```
mhutchinson: Please update PR description. It says "Define ... test", but in reality the more notable point here is the impl is changing.

impl/sql/mutationstorage/mutation_logs.go
```diff
@@ -188,6 +190,10 @@ func (m *Mutations) HighWatermark(ctx context.Context, directoryID string, logID
 // ReadLog may return more rows than batchSize in order to fetch all the rows at a particular timestamp.
 func (m *Mutations) ReadLog(ctx context.Context, directoryID string,
 	logID int64, low, high time.Time, batchSize int32) ([]*mutator.LogMessage, error) {
+	// Advance the low and high marks to the next highest quantum to preserve read semantics.
+	low = low.Add(quantum - 1).Truncate(quantum)
```
mhutchinson: If you don't do this, what changes? I would expect >= to do the right thing...

core/integration/storagetest/mutation_logs.go
gdbelvin: Hmm, I suppose that's true.  For tests with the same name, the framework will prepend `_$i` to the name.  That feels a bit hacky. Here I preserve an explicit link between the test name and the row it's testing.

impl/sql/mutationstorage/mutation_logs.go
gdbelvin: yes, that is a change that's needed

impl/sql/mutationstorage/mutation_logs.go
gdbelvin: The storage implementation will truncate low to the precision of quantum, with the effect that the timestamp will actually go backwards.

core/integration/storagetest/mutation_logs_reader.go
```diff
@@ -0,0 +1,199 @@
...
+)
+
+// LogsReadWriter supports test's ability to write to and read from the mutation logs.
+type LogsReadWriter interface {
+	sequencer.LogsReader
+	// Send submits the whole group of mutations atomically to a log.
+	// TODO(gbelvin): Create a batch level object to make it clear that this a batch of updates.
+	// Returns the timestamp that the mutation batch got written at.
+	// This timestamp can be used as an argument to the lower bound of ReadLog [low, high).
+	// To aquire a value to  use  for the  upper bound of ReadLog [low, high), use HighWatermark.
```
golangcibot: `aquire` is a misspelling of `acquire` (from `misspell`)
```suggestion

```

PR #1381, opened 2019-11-08
Library for converting time.Time to sequencer watermarks

gdbelvin:
This PR effectively makes pb.Watermarks a private object.
By only accessing the watermarks proto through a helper library, we give ourselves the ability to define clear rules for what valid watermarks are.

### DIFF
--- a/core/sequencer/metadata/sourceslice.go
+++ b/core/sequencer/metadata/sourceslice.go
@@ -15,17 +15,54 @@
 package metadata
 
 import (
+	"fmt"
+	"math"
 	"time"
 
 	spb "github.com/google/keytransparency/core/sequencer/sequencer_go_proto"
 )
 
-// Source returns a wrapper for SourceSlice
-func Source(s *spb.MapMetadata_SourceSlice) SourceSlice {
-	return SourceSlice{s: s}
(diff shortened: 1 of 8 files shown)


gdbelvin:
Haha. Yes. I had previously bundled these changes with a functional change
to what the int64s represented.

This PR is only the helper library.

core/sequencer/metadata/sourceslice.go
```diff
@@ -15,17 +15,54 @@
...
+	}}, nil
+}
+
+// FromProto returns a wrapper for SourceSlice
+func FromProto(s *spb.MapMetadata_SourceSlice) *SourceSlice {
+	return &SourceSlice{s: s}
 }
 
 // SourceSlice defines accessor and conversion methods for MapMetadata_SourceSlice
+// TODO(gbelvin): fully migrate to source slices that encode time directly.
```
pavelkalinnikov: Are you going to change the wire format of `MapMetadata` proto? If so, this would be going too far in my opinion. Can we stop the rush and discuss it again or maybe do a small doc with options, and technical reasons for doing it? I feel like this started from a small +-1 error problem, and now being inflated to a bunch of readability and complexity problems.

PR #1383, opened 2019-11-12
Don't emit metric err on read err

gdbelvin:
Emitting the wrong number of dimensions will cause an error.


gdbelvin:
I've now removed the offending line because the error is already covered by returning `err`.

I'm not sure it's worth the effort to fake out the reader and return an error. Especially since the metric function itself doesn't expose an error type.

PR #1384, opened 2019-11-13
Define watermarks as micros

gdbelvin:
Some time ago, different instances of Key Transparency had different units / meanings for the `int64`'s stored in `MapMetadata`.   When we switched to `time.Time` based APIs in 575344f256a6ca2e5fea3a9c3025aa69d57f63e3,  the open source `MapMetadata` silently acquired a definition of unix nanos and this change nearly caused an outage. 

This PR synchronizes the open source definition of `MapMetadata` with the unix micro one we had been defacto using internally.

### DIFF
--- a/core/sequencer/metadata/sourceslice.go
+++ b/core/sequencer/metadata/sourceslice.go
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Package metadata helps enforce a consistent standard of meaning around the map metadata object.
 package metadata
 
 import (
@@ -24,6 +25,7 @@ import (
 
 var minTime = time.Unix(0, math.MinInt64) // 1677-09-21 00:12:43.145224192
 var maxTime = time.Unix(0, math.MaxInt64) // 2262-04-11 23:47:16.854775807
+const nanosPerQuantum = int64(time.Microsecond / time.Nanosecond)
 
 // validateTime determines whether a timestamp is valid.
 // Valid timestamps can be represented by an int64 number of microseconds from the epoch.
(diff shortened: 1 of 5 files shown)


mhutchinson:
This looks like it doesn't just define things, but functionally changes anything for existing clients. Is this expected? How should clients migrate (or did that ship sail)?

core/sequencer/metadata/sourceslice.go
```diff
@@ -50,8 +52,8 @@ func New(logID int64, low, high time.Time) (*SourceSlice, error) {
 	}
 	return &SourceSlice{s: &spb.MapMetadata_SourceSlice{
 		LogId:            logID,
-		LowestInclusive:  low.UnixNano(),
-		HighestExclusive: high.UnixNano(),
+		LowestInclusive:  low.UnixNano() / nanosPerQuantum,
+		HighestExclusive: high.UnixNano() / nanosPerQuantum,
```
pavelkalinnikov: These lines change what's being stored? Does this fix some problem? Please add it to the PR description.

pavelkalinnikov:
Even though this change fixes some compatibility problem in our existing maps, it makes the contract stricter. If it wasn't `time.Time` all the way between storage and API layers, introducing this restriction wouldn't be needed.

pavelkalinnikov:
Do I understand correctly that merging this PR will unblock importing KT and updating our internal map code? If so, I can approve it modulo further discussion about getting rid of time notion in watermarks (except inside the storage implementations).

core/sequencer/metadata/sourceslice.go
gdbelvin: Yes, this does indeed fix a problem.  Documented in the description.

gdbelvin:
Yes, this is blocking importing KT at the moment because of the difference in watermark definitions.  At the moment, the restriction in the contract is necessary. 

Happy to defer further discussions beyond this for the moment.

PR #1403, opened 2019-12-11
Improve mutation log integration tests

gdbelvin:
- Remove dependency on time.Now
- Check validity of insert timestamps
- Support implementations based on unix seconds

### DIFF
--- a/core/integration/storagetest/mutation_logs.go
+++ b/core/integration/storagetest/mutation_logs.go
@@ -54,7 +54,7 @@ func mustMarshal(t *testing.T, p proto.Message) []byte {
 	return b
 }
 
-// The lowest common denominator for the minimum timestamp supported between all implementations.
+// The minimum timestamp supported between all implementations.
 var minWatermark = time.Unix(0, 0)
 
 // TestReadLog ensures that reads happen in atomic units of batch size.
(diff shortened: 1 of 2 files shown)


core/integration/storagetest/mutation_logs.go
```diff
@@ -54,8 +54,8 @@ func mustMarshal(t *testing.T, p proto.Message) []byte {
 	return b
 }
 
-// https://dev.mysql.com/doc/refman/8.0/en/datetime.html
-var minWatermark = time.Date(1000, 1, 1, 0, 0, 0, 0, time.UTC)
+// The lowest common denominator for the minimum timestamp supported between all implementations.
```
mhutchinson: It's not clear what the [LCD](https://en.wikipedia.org/wiki/Lowest_common_denominator) of min timestamps would be?

core/integration/storagetest/mutation_logs.go
```diff
@@ -64,11 +64,17 @@ func (mutationLogsTests) TestReadLog(ctx context.Context, t *testing.T, newForTe
 	m, done := newForTest(ctx, t, directoryID, logID)
 	defer done(ctx)
 	// Write ten batches.
+	var lastTS time.Time
 	for i := byte(0); i < 10; i++ {
 		entry := &pb.EntryUpdate{Mutation: &pb.SignedEntry{Entry: mustMarshal(t, &pb.Entry{Index: []byte{i}})}}
-		if _, err := m.Send(ctx, directoryID, logID, entry, entry, entry); err != nil {
+		ts, err := m.Send(ctx, directoryID, logID, entry, entry, entry)
```
mhutchinson: Optional side-quest: You could use `lastTS` here instead of introducing `ts` and I think it all checks out.

gdbelvin: True, but then I'd have to side declare err. Not sure which one is cleaner.

core/integration/storagetest/mutation_logs.go
gdbelvin: LCD was the wrong word choice.
"Lowest watermark supported by all implementations" is more precise.

PR #1404, opened 2019-12-12
Query logs by watermark instead of Time

pavelkalinnikov:
The reason for having `time.Time` in the storage API is primarily monitoring.
Watermarks are currently used to compute latency of different stages of the
pipeline.

There are many disadvantages though:
- It hurts to review PRs that operate with `time.Time` magic in queries.
- Zero value needs special treatment. 
- There are 3(!) different types for time: Golang's `time.Time`,
  protobuf's Timestamp, and MySQL DATETIME. We depend on conversions
  between these types to work correctly.
- Storage implementations need to do arcane rounding arithmetics with
  the timestamp. It feels like using float instead of int.
- Using `time.Time` in the API prevents non-Time implementations, e.g.
  simple sequence numbers like in Kafka. For the problem that these
  timestamps are solving, it actually makes more sense to use
  non-timestamp types.

The proposed solution is to use obscure `water.Mark` type for queries, and
make timestamp a field of the entry type to support the metrics case. Since
metrics don't need to be as precise as query watermarks, this will massively
simplify the code. This PR removes `time.Time` semantics, and follow-up PRs
will reintroduce `time.Time` in a form of a field.


core/water/mark.go
```diff
@@ -0,0 +1,46 @@
...
+	return Mark{value: value}
+}
+
+// Value returns the uint64 representation of the watermark.
+func (m Mark) Value() uint64 {
+	return m.value
+}
+
+// Add increments the watermark by the passed in value.
+func (m Mark) Add(value uint64) Mark {
```
pavelkalinnikov: Note: This method is used only in tests.

core/water/mark.go
```diff
@@ -0,0 +1,46 @@
...
+	value uint64
+}
+
+// NewMark returns a watermark from the given integer value.
+func NewMark(value uint64) Mark {
+	return Mark{value: value}
+}
+
+// Value returns the uint64 representation of the watermark.
+func (m Mark) Value() uint64 {
```
pavelkalinnikov: Note: This method is used only in tests and storage implementations. Application layer doesn't use it.

core/sequencer/metadata/sourceslice.go
```diff
@@ -16,71 +16,42 @@
...
-// Every valid timestamp can be represented by a time.Time, but the converse is not true.
-func validateTime(ts time.Time) error {
-	if ts.Before(minTime) {
-		return fmt.Errorf("timestamp %v before %v", ts, minTime)
-	}
-	if ts.After(maxTime) {
-		return fmt.Errorf("timestamp %v after %v", ts, maxTime)
-	}
-	return nil
-}
```
pavelkalinnikov: This is my favourite.

impl/sql/mutationstorage/mutations.go
```diff
@@ -37,12 +37,12 @@ var (
...
 		`CREATE TABLE IF NOT EXISTS Queue (
-		DirectoryID VARCHAR(30)   NOT NULL,
-		LogID    BIGINT           NOT NULL,
-		Time     DATETIME(6)      NOT NULL,
-		LocalID  BIGINT           NOT NULL,
-		Mutation BLOB             NOT NULL,
-		PRIMARY KEY(DirectoryID, LogID, Time, LocalID )
+		DirectoryID VARCHAR(30) NOT NULL,
+		LogID       BIGINT      NOT NULL,
+		Timestamp   BIGINT      NOT NULL, -- In nanoseconds from Unix epoch.
```
pavelkalinnikov: Maybe call it `TimeNanos` for more verbosity.

core/sequencer/server.go
```diff
@@ -130,17 +124,20 @@ type Watermarks map[int64]int64
...
 // LogsReader reads messages in multiple logs.
 type LogsReader interface {
-	// HighWatermark returns the number of items and the highest primary
-	// key up to batchSize items after start (exclusive).
-	HighWatermark(ctx context.Context, directoryID string, logID int64, start time.Time,
-		batchSize int32) (count int32, watermark time.Time, err error)
+	// HighWatermark counts up to batchSize entries in the specified log, located
+	// strictly after the given start watermark (i.e. exclusive). Returns the
+	// number of entries found, and the watermark covering the last entry.
+	// TODO(pavelkalinnikov): Return watermark just beyond the last entry.
```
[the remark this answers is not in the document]
gdbelvin: I'm not sure this comment is correct

gdbelvin:
Not a criticism, just thinking out-loud. The cons of this approach:

- No standard meaning for watermark objects.
   - Which means that monitoring will need to be database specific in order to interpret them.
   - A continual temptation to imply meaning on these opaque objects
 - Previous objections that have been removed by the work to migrate to `time.Time`
   - Spurious leakage of `time.Now` and `time.UnixNano` into the code base, effectively granting meaning to these opaque objects even while claiming that they are opaque.

gdbelvin:
@pavelkalinnikov  Could you
- [ ] Update the mysql implementation to be based on microseconds so that a) it matches spanner, and b) ensures that we have a diversity of time granularities in the testable code base?

gdbelvin:
- [ ] I would also request a conversion API in the storage layer to convert the opaque object into a `time.Time` for monitoring purposes.

core/sequencer/server.go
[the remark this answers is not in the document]
pavelkalinnikov: Check out the new comment which formalizes the semantics.

core/sequencer/server.go
```diff
@@ -130,17 +124,31 @@ type Watermarks map[int64]int64
...
+	HighWatermark(ctx context.Context, directoryID string, logID int64, start water.Mark,
+		batchSize int32) (count int32, high water.Mark, err error)
 
 	// ListLogs returns the logIDs associated with directoryID that have their write bits set,
 	// or all logIDs associated with directoryID if writable is false.
 	ListLogs(ctx context.Context, directoryID string, writable bool) ([]int64, error)
 
-	// ReadLog returns the lowest messages in the [low, high) range stored in the specified log, up to batchSize.
-	ReadLog(ctx context.Context, directoryID string, logID int64, low, high time.Time,
+	// ReadLog returns the lowest messages in the [low, high) watermarks range of
```
pavelkalinnikov: This comments also worth making more formal.

pavelkalinnikov:
> @pavelkalinnikov could you
> 
> * [ ]  Update the mysql implementation to be microsecond based?
>   
>   * This would ensure diversity of time granularities exist in the tested code base
>   * Ensure that external tests exist for a microsecond based implementation.

I made the MySQL implementation in nanoseconds.
Our internal implementation is in microseconds.
The in-memorhy implementation is in sequence numbers.

So we are covering all cases we care about.



> * [ ]  I would also request a conversion API in the storage layer to convert the opaque object into a `time.Time` for monitoring purposes.

I am doing this in a follow-up PR. We agreed to do it slightly differently: the `LogMessage` type will contain a `time.Time` field which will be used by monitoring.

impl/sql/mutationstorage/mutation_logs_test.go
```diff
@@ -105,24 +106,24 @@ func TestSend(t *testing.T) {
 	m, done := newForTest(ctx, t, directoryID, 1, 2)
 	defer done(ctx)
 	update := []byte("bar")
-	ts1 := time.Now().Truncate(time.Microsecond)
-	ts2 := ts1.Add(1 * time.Microsecond)
-	ts3 := ts2.Add(1 * time.Microsecond)
+	wm1 := water.NewMark(uint64(time.Now().UnixNano()))
```
gdbelvin: Should this be micros?

pavelkalinnikov: Done.

PR #1405, opened 2019-12-12
Standardize on unix seconds for metrics

gdbelvin:
Achieve monitoring consistency


gdbelvin:
Temporary patch while I ponder #1404

PR #1410, opened 2019-12-17
Set local ID for messages in memory impl

pavelkalinnikov:


pavelkalinnikov:
I wonder if we should have some tests exercising uniqueness of (ID, LocalID).

pavelkalinnikov:
Added some test coverage.

PR #1411, opened 2019-12-17
Add histogram metric for submit-apply latency

pavelkalinnikov:


core/sequencer/server.go
```diff
@@ -471,6 +477,11 @@ func (s *Server) ApplyRevision(ctx context.Context, in *spb.ApplyRevisionRequest
 	}
 	glog.V(2).Infof("CreateRevision: WriteLeaves:{Revision: %v}", in.Revision)
 
+	writtenAt := time.Now()
+	for _, li := range logItems {
+		appliedLatency.Observe(writtenAt.Sub(li.CreatedAt).Seconds(), in.DirectoryId, strconv.FormatInt(li.LogID, 10))
```
gdbelvin: What do you think about using `time.Since`?  I could see an argument against it in that each loop would use a different version of "now"

```suggestion
		appliedLatency.Observe(time.Since(li.CreatedAt).Seconds(), in.DirectoryId, strconv.FormatInt(li.LogID, 10))
```

pavelkalinnikov: `time.Since` is a tiny bit more pessimistic, as the things have been written already, but each subsequent `time.Now` will be further off. I ran `time.Now()` 500 times locally, and the last one is off by 2ms from the first one. Not too bad, but I'd rather not have it.

PR #1418, opened 2019-12-20
Setup liveness and readiness k8 probes

gdbelvin:
As a precondition for using a GKE load ballancer, we must define readiness probes.

https://cloud.google.com/kubernetes-engine/docs/concepts/ingress#limitations
https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

### DIFF
--- a/cmd/keytransparency-sequencer/server.go
+++ b/cmd/keytransparency-sequencer/server.go
@@ -16,6 +16,7 @@ package main
 
 import (
 	"context"
+	"database/sql"
 	"net"
 	"net/http"
 
@@ -26,9 +27,11 @@ import (
 	"google.golang.org/grpc"
 )
 
-func serveHTTPMetric(addr string) {
+func serveHTTPMetric(addr string, sqldb *sql.DB) {
 	metricMux := http.NewServeMux()
 	metricMux.Handle("/metrics", promhttp.Handler())
+	metricMux.Handle("/healthz", serverutil.Healthz())
+	metricMux.Handle("/readyz", serverutil.Readyz(sqldb))
 
 	glog.Infof("Hosting metrics on %v", addr)
 	if err := http.ListenAndServe(addr, metricMux); err != nil {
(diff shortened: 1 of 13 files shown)


gdbelvin:
Depends on #1420

cmd/serverutil/healthz.go
```diff
@@ -0,0 +1,24 @@
+// Copyright 2019 Google Inc. All Rights Reserved.
```
pavelkalinnikov:
```suggestion
// Copyright 2020 Google Inc. All Rights Reserved.
```
passim

cmd/keytransparency-sequencer/server.go
```diff
@@ -26,9 +27,11 @@ import (
 	"google.golang.org/grpc"
 )
 
-func serveHTTPMetric(addr string) {
+func serveHTTPMetric(addr string, sqldb *sql.DB) {
```
pavelkalinnikov: Please make "metrics" plural. Also, maybe make this name a bit more generic, e.g. `serveHTTPMetricsAndStatus`?

cmd/keytransparency-sequencer/server.go
```diff
@@ -26,9 +27,11 @@ import (
...
 	metricMux := http.NewServeMux()
```
pavelkalinnikov: nit: I would just call this `mux`.

cmd/keytransparency-sequencer/server.go
```diff
@@ -26,9 +27,11 @@ import (
...
 	metricMux.Handle("/metrics", promhttp.Handler())
+	metricMux.Handle("/healthz", serverutil.Healthz())
+	metricMux.Handle("/readyz", serverutil.Readyz(sqldb))
 
 	glog.Infof("Hosting metrics on %v", addr)
```
pavelkalinnikov:
```suggestion
	glog.Infof("Hosting server status and metrics on %v", addr)
```

PR #1421, opened 2020-01-02
Less flakey codecov

gdbelvin:
Code coverage seems to bounce around a bit.
Allow it do bounce around by 1% without going red.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -3,8 +3,6 @@ ignore:
   - "**/*.pb.gw.go"  # ignore generated protos
 coverage:
   range: 60..80
-  round: down
-  precision: 2
   status:
     project:
       default:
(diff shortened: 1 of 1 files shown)


codecov.yml
```diff
@@ -1,3 +1,14 @@
 ignore: 
   - "**/*.pb.go"     # ignore generated protos
   - "**/*.pb.gw.go"  # ignore generated protos
+coverage:
+  range: 60..80
+  round: down
```
NatalieDoduc: maybe remove `round` and `precision` as they are set the the default values?

codecov.yml
```diff
@@ -1,3 +1,14 @@
 ignore: 
   - "**/*.pb.go"     # ignore generated protos
   - "**/*.pb.gw.go"  # ignore generated protos
+coverage:
+  range: 60..80
```
NatalieDoduc: why an upper bound of 80? or am I misreading this config line?

gdbelvin: An upper bound of 80 means that code coverage above 80% will be marked as green.
This can be adjusted as the code coverage for the project grows over time.

PR #1423, opened 2020-01-03
Turn on monitor integration tests

gdbelvin:
Assert that the monitor container comes up successfully in the integration tests.

Previously the monitor would fail on startup if a gRPC request to the server failed. Now the monitor will retry the command for up to 1 minute.

### DIFF
--- a/cmd/keytransparency-monitor/main.go
+++ b/cmd/keytransparency-monitor/main.go
@@ -66,6 +66,7 @@ func main() {
 	}
 	ktClient := pb.NewKeyTransparencyClient(cc)
 
+	// TODO(gbelvin): implement backoff for this
 	config, err := ktClient.GetDirectory(ctx, &pb.GetDirectoryRequest{DirectoryId: *directoryID})
 	if err != nil {
 		glog.Exitf("Could not read directory info %v:", err)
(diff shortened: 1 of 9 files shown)


cmd/keytransparency-monitor/main.go
```diff
@@ -66,10 +69,19 @@ func main() {
...
+	// TODO(gbelvin): implement backoff for this
+	b := backoff.Backoff{
+		Min:    time.Millisecond,
+		Max:    time.Second,
+		Factor: 1.5,
 	}
+	var config *pb.Directory
+	cctx, cancel := context.WithTimeout(ctx, time.Minute)
+	defer cancel()
+	b.Retry(cctx, func() (err error) {
```
golangcibot: Error return value of `b.Retry` is not checked (from `errcheck`)

cmd/keytransparency-monitor/main.go
```diff
@@ -66,10 +69,19 @@ func main() {
 	}
 	ktClient := pb.NewKeyTransparencyClient(cc)
 
-	config, err := ktClient.GetDirectory(ctx, &pb.GetDirectoryRequest{DirectoryId: *directoryID})
-	if err != nil {
-		glog.Exitf("Could not read directory info %v:", err)
+	// TODO(gbelvin): implement backoff for this
```
NatalieDoduc: Can you be more specific on what more needs to be implemented?

gdbelvin: Implemented and removed comment.

PR #1425, opened 2020-01-03
Remove kompose k8 cruft

gdbelvin:
The Kubernetes configs were originally created with `kompose convert` but the k8 configs have since significantly diverged due to sidecars, local and gke specific configs. 

This PR removes kompose metadata that is no longer correct. 

The k8 configs and the docker-compose configs are currently maintained by hand because there is not an exact 1:1 feature mapping between them.

### DIFF
--- a/deploy/kubernetes/base/db-deployment.yaml
+++ b/deploy/kubernetes/base/db-deployment.yaml
@@ -1,10 +1,6 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  annotations:
-    kompose.cmd: kompose convert --file ../../docker-compose.yml
-    kompose.version: 1.14.0 ()
-  creationTimestamp: null
   labels:
     io.kompose.service: db
   name: db
--- a/deploy/kubernetes/base/init-pod.yaml
+++ b/deploy/kubernetes/base/init-pod.yaml
@@ -1,7 +1,6 @@
 apiVersion: v1
 kind: Pod
 metadata:
-  creationTimestamp: null
   labels:
     io.kompose.service: init
   name: init
--- a/deploy/kubernetes/base/db-service.yaml
+++ b/deploy/kubernetes/base/db-service.yaml
@@ -1,10 +1,6 @@
 apiVersion: v1
 kind: Service
 metadata:
-  annotations:
-    kompose.cmd: kompose convert --file ../../docker-compose.yml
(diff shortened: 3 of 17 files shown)


deploy/kubernetes/base/init-pod.yaml
```diff
@@ -1,7 +1,6 @@
 apiVersion: v1
 kind: Pod
 metadata:
-  creationTimestamp: null
```
NatalieDoduc: what did this do? can you also add a mention about `creationTimestamp` in the PR description?

deploy/kubernetes/base/db-deployment.yaml
```diff
@@ -1,10 +1,6 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  annotations:
-    kompose.cmd: kompose convert --file ../../docker-compose.yml
```
NatalieDoduc: I'm guessing this was added as a hint to how the .yaml file was created? If that's the case, it seems mildly useful to keep (if possibly outdated, from the looks of the path) Maybe add this information as part of ./deploy/kubernetes/README.md if maintaining these annotations proves too cumbersome,

Just curious what the log bloating you refer to is.

deploy/kubernetes/base/init-pod.yaml
gdbelvin: I don't know what this does, but from what I read in the k8 documentation, empty and default fields in the configs do no need be be specified.

deploy/kubernetes/base/db-deployment.yaml
gdbelvin: This was the original command that generated these files, but the two configs have now significantly diverged to the point that it is no longer useful to document this.

deploy/kubernetes/base/init-pod.yaml
NatalieDoduc: Could you amend the PR description to mention this change as well?

deploy/kubernetes/base/db-deployment.yaml
NatalieDoduc: Ah ha - so these files have now been manually changed, since their original `kompose convert` creation. Makes sense to not document in that case. One question, though, what is the relationship if any between the docker-compose file and the kube files, then? Why do we use one over the other?

Issue #1200, opened 2019-02-08
[docker-compose] Trillian healthchecks don't work


DazWilkin:
This is my fault :-(

Endeavoring to minimize the Docker image sizes of the Trillian "trio" and not being aware (at the time) that these were used for anything other than Kubernetes (its healthchecks don't suffer Docker's limitation), I converted the Dockerfiles to multi-stage builds and used [Distroless](https://github.com/GoogleContainerTools/distroless) for the runtime.

Distroless images do not include a shell. Docker Healthchecks require (`CMD`) a shell to run.

So the healthchecks fail when used by Docker and Docker Compose, e.g:
```
docker-compose ps
WARNING: Some services (init) use the 'deploy' key, which will be ignored. Compose does not support 'deploy' configuration - use `docker stack deploy` to deploy to a swarm.
            Name                          Command                   State                            Ports                    
------------------------------------------------------------------------------------------------------------------------------
keytransparency_log-server_1   /trillian_log_server --mys ...   Up (unhealthy)
keytransparency_log-signer_1   /trillian_log_signer --mys ...   Up (unhealthy)
keytransparency_map-server_1   /trillian_map_server --mys ...   Up (unhealthy)
...
keytransparency_sequencer_1    /go/bin/keytransparency-se ...   Up (healthy)     0.0.0.0:8083->8081/tcp
keytransparency_server_1       /go/bin/keytransparency-se ...   Up (healthy)     0.0.0.0:443->8080/tcp, 0.0.0.0:8081->8081/tcp
prometheus-to-sd               /monitor --stackdriver-pre ...   Exit 255
```

**Recommend**
The easiest (albeit not good) solution is to remove the healthchecks from the Trillian servers:
https://github.com/google/keytransparency/blob/5ac39fe21be1c6f34b20b8e9b8242db00800212b/docker-compose.yml#L62
https://github.com/google/keytransparency/blob/5ac39fe21be1c6f34b20b8e9b8242db00800212b/docker-compose.yml#L87
https://github.com/google/keytransparency/blob/5ac39fe21be1c6f34b20b8e9b8242db00800212b/docker-compose.yml#L109

An alternative would be to bundle a simple healthcheck binary in the images that may be used instead:
https://github.com/DazWilkin/golang-healthcheck

gdbelvin:
Should this be converted into a trillian issue to use Distroless for docker images?

DazWilkin:
Yes :-)

PR #1427, opened 2020-01-07
Fix Docker HEALTHCHECK

gdbelvin:
- cmd/healthcheck binary to make http(s) requests.
- Move HEALTHCHECK to Dockerfile
- Verify health in test scripts
- Remove unsupported Trillian health checks https://github.com/google/keytransparency/issues/1200

### DIFF
--- a/cmd/healthcheck/main.go
+++ b/cmd/healthcheck/main.go
@@ -1,0 +1,41 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// healthcheck makes a HTTP(S) request to the supplied url and exits with 0
+// (success) if the HTTP response code was in the 2xx range or 1 (unhealthy).
+//
+// TLS certificate errors are ignored in order to support self-signed certs.
+package main
+
+import (
+	"crypto/tls"
+	"log"
+	"net/http"
+	"os"
+)
+
+func main() {
+	if len(os.Args) < 2 {
+		log.Fatal("Expected URL as command-line argument")
+	}
+	url := os.Args[1]
+	http.DefaultTransport.(*http.Transport).TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
+	resp, err := http.Get(url)
+	if err != nil {
+		log.Fatalf("http.Get(%v): %v", url, err)
+	}
+	if resp.StatusCode >= 200 && resp.StatusCode <= 299 {
+		log.Fatalf("HTTP status %v was not in the 2xx range", resp.StatusCode)
+	}
+}
--- a/cmd/healthcheck/main.go
+++ b/cmd/healthcheck/main.go
@@ -33,7 +33,7 @@ func main() {
 	//nolint:gas
 	http.DefaultTransport.(*http.Transport).TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
 
-	// TODO(gbelvin): sanitize user input to avoice SSRF attacks.
+	//nolint:gas
 	resp, err := http.Get(url)
 	if err != nil {
 		log.Fatalf("http.Get(%v): %v", url, err)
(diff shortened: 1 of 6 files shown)


cmd/healthcheck/main.go
```diff
@@ -0,0 +1,41 @@
...
+	"net/http"
+	"os"
+)
+
+func main() {
+	if len(os.Args) < 2 {
+		log.Fatal("Expected URL as command-line argument")
+	}
+	url := os.Args[1]
+	http.DefaultTransport.(*http.Transport).TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
```
golangcibot: G402: TLS InsecureSkipVerify set true. (from `gosec`)

cmd/healthcheck/main.go
```diff
@@ -0,0 +1,41 @@
...
+	resp, err := http.Get(url)
```
golangcibot: G107: Potential HTTP request made with variable url (from `gosec`)

docker-compose.yml
```diff
@@ -132,11 +117,6 @@ services:
       - --v=1
     labels:
       kompose.service.type: LoadBalancer
-    healthcheck:
-      test: ["CMD", "curl", "-f", "http://localhost:8081/readyz"]
-      interval: 30s
-      timeout: 10s
-      retries: 5
```
RJPercival: These values do not match those used by Docker's HEALTHCHECK directive. Is that ok?

docker-compose.yml
```diff
@@ -164,11 +144,6 @@ services:
     ports:
       - "8080"
       - "8081"
-    healthcheck:
-      test: ["CMD", "curl", "-f", "http://localhost:8081/readyz"]
-      interval: 30s
-      timeout: 10s
-      retries: 5
```
RJPercival: These values do not match those used by Docker's HEALTHCHECK directive. Is that ok?

cmd/healthcheck/main.go
```diff
@@ -0,0 +1,44 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
```
RJPercival:
```suggestion
// Copyright 2020 Google Inc. All Rights Reserved.
```

cmd/healthcheck/main.go
```diff
@@ -0,0 +1,44 @@
...
+	"os"
+)
+
+func main() {
+	if len(os.Args) < 2 {
+		log.Fatal("Expected URL as command-line argument")
+	}
+	url := os.Args[1]
+	//nolint:gas
+	http.DefaultTransport.(*http.Transport).TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
```
RJPercival: nit: This could break in a future version of Go - there's no guarantee that `http.DefaultTransport` will always be a `http.Transport` instance. The `http` documentation suggests the following code for configuring TLS:
```go
tr := &http.Transport{
	TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
}
client := &http.Client{Transport: tr}
```
However, I'm fine with leaving this as-is - odds are that it won't break, and it's not that bad if it does.

cmd/healthcheck/main.go
```diff
@@ -0,0 +1,44 @@
...
+
+	//nolint:gas
+	resp, err := http.Get(url)
+	if err != nil {
+		log.Fatalf("http.Get(%v): %v", url, err)
+	}
+	if resp.StatusCode < 200 || resp.StatusCode > 299 {
+		log.Fatalf("HTTP status %v was not in the 2xx range", resp.StatusCode)
```
RJPercival: Might also be helpful to log the response body, in case it has a useful error message in it.

docker-compose.yml
gdbelvin: I'm ok reverting to the defaults for the health check.  These values have currently produced checks that fail a bit too quickly

docker-compose.yml
gdbelvin: see above

cmd/healthcheck/main.go
gdbelvin: I'll change it

cmd/healthcheck/main.go
gdbelvin: Reading the body introduces additional error paths. Not sure if the complexity is worth it.

cmd/keytransparency-server/Dockerfile
```diff
@@ -7,11 +7,14 @@ RUN go mod download
...
 
 FROM gcr.io/distroless/base
 
 COPY --from=build /go/bin/keytransparency-server /
+COPY --from=build /go/bin/healthcheck /
 
 ENTRYPOINT ["/keytransparency-server"]
+HEALTHCHECK CMD ["/healthcheck","http://localhost:8081/healthz"]
 
 EXPOSE 8080
```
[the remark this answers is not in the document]
gdbelvin: done

gdbelvin:
Updated to address comments

PR #1430, opened 2020-01-10
Cleanup old docker images

gdbelvin:
Delete docker images older than 30 days

### DIFF
--- a/scripts/gcrgc.sh
+++ b/scripts/gcrgc.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+
+# Copyright © 2017 Google Inc.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


scripts/gcrgc.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+
+# Copyright © 2017 Google Inc.
```
taknira: It's not 2017 anymore :)

scripts/gcrgc.sh
```diff
@@ -0,0 +1,53 @@
...
+# limitations under the License.
+#
+# Souce: https://gist.github.com/ahmetb/7ce6d741bd5baa194a3fac6b1fec8bb7
+
+IFS=$'\n\t'
+set -eou pipefail
+
+if [[ "$#" -ne 2 || "${1}" == '-h' || "${1}" == '--help' ]]; then
+  cat >&2 <<"EOF"
+gcrgc.sh cleans up tagged or untagged images pushed before specified date
```
taknira: Out of interest, does 'clean up' mean delete?

gdbelvin: yes

scripts/gcrgc.sh
gdbelvin: That's from the original file

PR #1431, opened 2020-01-15
Use ingress object for HTTPS loadballancing

gdbelvin:
This PR shifts the responsibility for terminating TLS connections from the job binaries to an ingress this will significantly reduce the complexity of acquiring a legitimate TLS certificate.

### DIFF
--- a/cmd/keytransparency-sequencer/main.go
+++ b/cmd/keytransparency-sequencer/main.go
@@ -138,7 +138,7 @@ func main() {
 	)
 
 	// Listen and create empty grpc client connection.
-	lis, conn, done, err := serverutil.Listen(ctx, *listenAddr)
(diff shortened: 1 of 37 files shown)


cmd/keytransparency-sequencer/main.go
```diff
@@ -140,7 +138,7 @@ func main() {
 	)
 
 	// Listen and create empty grpc client connection.
-	lis, conn, done, err := serverutil.ListenTLS(ctx, *addr, *certFile, *keyFile)
+	lis, conn, done, err := serverutil.Listen(ctx, *listenAddr)
```
golangcibot: undeclared name: `listenAddr` (from `typecheck`)

PR #1434, opened 2020-01-17
Use TLS Listener

gdbelvin:
By moving TLS setup to the listener, we better isolate the crypto bits, and make things more API compatible with cmux.

Interesting Notes For Posterity:
- To support HTTP2, we need to explicitly ask for it in the `NextProtos` of the TLS config

### DIFF
--- a/cmd/keytransparency-monitor/main.go
+++ b/cmd/keytransparency-monitor/main.go
@@ -123,7 +123,7 @@ func main() {
 	grpc_prometheus.Register(grpcServer)
 	grpc_prometheus.EnableHandlingTimeHistogram()
 
-	lis, conn, done, err := serverutil.Listen(ctx, *addr, *certFile)
+	lis, conn, done, err := serverutil.ListenTLS(ctx, *addr, *certFile, *keyFile)
 	if err != nil {
 		glog.Fatalf("Listen(%v): %v", *addr, err)
 	}
@@ -132,8 +132,7 @@ func main() {
 	g, gctx := errgroup.WithContext(ctx)
 	g.Go(func() error { return serverutil.ServeHTTPMetrics(*metricsAddr, serverutil.Healthz()) })
 	g.Go(func() error {
-		return serverutil.ServeHTTPAPIAndGRPC(gctx, lis, *keyFile, *certFile,
-			grpcServer, conn, mopb.RegisterMonitorHandler)
+		return serverutil.ServeHTTPAPIAndGRPC(gctx, lis, grpcServer, conn, mopb.RegisterMonitorHandler)
 	})
 	glog.Errorf("Monitor exiting: %v", g.Wait())
 }
--- a/cmd/keytransparency-sequencer/main.go
+++ b/cmd/keytransparency-sequencer/main.go
@@ -139,7 +139,7 @@ func main() {
 	)
 
 	// Listen and create empty grpc client connection.
-	lis, conn, done, err := serverutil.ListenTLS(ctx, *listenAddr, *certFile, *keyFile)
+	lis, conn, done, err := serverutil.ListenTLS(ctx, *addr, *certFile, *keyFile)
 	if err != nil {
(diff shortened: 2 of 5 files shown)


cmd/serverutil/serverutil.go
```diff
@@ -17,6 +17,7 @@ package serverutil
```
NatalieDoduc: I'm curious about how you're using the `cmd` pattern. Seems like this particular use is actually a library, rather than a binary?

cmd/keytransparency-monitor/main.go
```diff
@@ -20,11 +20,11 @@ import (
 	"crypto/tls"
 	"flag"
 	"net"
-	"net/http"
 	"time"
 
 	"github.com/golang/glog"
 	"github.com/google/trillian/crypto/keys/pem"
+	"golang.org/x/sync/errgroup"
```
NatalieDoduc: I recently found out about this guy, it's really good!

cmd/keytransparency-sequencer/main.go
```diff
@@ -183,21 +167,20 @@ func main() {
...
-		pb.RegisterKeyTransparencyAdminHandlerFromEndpoint,
-	)
-	runSequencer(ctx, conn, directoryStorage)
+	g, gctx := errgroup.WithContext(ctx)
+	g.Go(func() error { return serverutil.ServeHTTPMetrics(*metricsAddr, serverutil.Readyz(sqldb)) })
+	g.Go(func() error {
+		return serverutil.ServeHTTPAPIAndGRPC(gctx, lis, grpcServer, conn,
+			pb.RegisterKeyTransparencyAdminHandler)
+	})
+	g.Go(func() error { return runSequencer(gctx, conn, directoryStorage) })
```
NatalieDoduc: what's the impact (if any) of running runSequencer in a goroutine as opposed to inline like before?

cmd/keytransparency-sequencer/main.go
```diff
@@ -241,9 +224,10 @@ func runSequencer(ctx context.Context, conn *grpc.ClientConn,
 		}
 	})
 
-	sequencer.PeriodicallyRun(ctx, time.Tick(*refresh), func(ctx context.Context) {
+	go sequencer.PeriodicallyRun(ctx, time.Tick(*refresh), func(ctx context.Context) {
```
NatalieDoduc: why not use g.Go() here as well?

cmd/serverutil/serverutil.go
gdbelvin: True, this breaks the pattern a bit, I could move to an internal directory?

cmd/keytransparency-sequencer/main.go
gdbelvin: At the moment these functions don't return an error.
I don't *think* we want them to because the first error causes all the other contexts decendent from `errgroup.WithContext` to exit, and we want the sequencer to carry on.

cmd/keytransparency-sequencer/main.go
gdbelvin: The impact is that this function now returns. I suppose it doesn't need to be launched with g.Go does it?

cmd/serverutil/serverutil.go
NatalieDoduc: Yeah, I'd prefer if the `cmd` pattern be respected. `internal` would be nicer indeed - you don't need these to be publicly available, do you?

cmd/keytransparency-sequencer/main.go
```diff
@@ -139,7 +139,7 @@ func main() {
 	)
 
 	// Listen and create empty grpc client connection.
-	lis, conn, done, err := serverutil.Listen(ctx, *addr, *certFile)
+	lis, conn, done, err := serverutil.ListenTLS(ctx, *listenAddr, *certFile, *keyFile)
```
golangcibot: undeclared name: `listenAddr` (from `typecheck`)

PR #1435, opened 2020-01-17
Use cmux

gdbelvin:
Use cmux to differentiate between HTTP and GRPC connections.

cmux also (critically) supports serving gRPC over HTTP2 without TLS.

PR #1442, opened 2020-01-27
Serve HTTP2 health check

gdbelvin:
Prioritize HTTP2 over HTTP1

Revert #1435 and use `http.ServeHTTP` rather than cmux.
- cmux prevents http.Serve from serving HTTP2
- http.Serve will only serve HTTP2 over TLS
   - http.Serve uses a type assertion to validate this
   - cmux interrupts this because it wraps the connection.
   - even if the connection did come from a TLS listener 

The GCE ingress load ballancer requires services to serve 
an HTTP 200 OK at / on the same protocol that the server uses.
To support GRPC, this means using HTTP2 throuought.
HTTP2 also requires TLS, both by golang and GCE ingress.

### DIFF
--- a/cmd/serverutil/serverutil.go
+++ b/cmd/serverutil/serverutil.go
@@ -27,12 +27,12 @@ import (
 	"google.golang.org/grpc"
 )
 
-// GrpcHandlerFunc returns an http.Handler that delegates to grpcServer on incoming gRPC
-// connections or otherHandler otherwise. Copied from cockroachdb.
-func GrpcHandlerFunc(grpcServer http.Handler, otherHandler http.Handler) http.Handler {
+// gRPCHandlerFunc returns an http.Handler that delegates to grpcServer on incoming gRPC
+// connections or otherHandler otherwise.
+func gRPCHandlerFunc(grpcServer http.Handler, otherHandler http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		// This is a partial recreation of gRPC's internal checks.
-		// https://github.com/grpc/grpc-go/blob/master/transport/handler_server.go#L62
+		// https://github.com/grpc/grpc-go/blob/v1.26.0/internal/transport/handler_server.go#L62
 		if r.ProtoMajor == 2 && strings.HasPrefix(
 			r.Header.Get("Content-Type"), "application/grpc") {
 			grpcServer.ServeHTTP(w, r)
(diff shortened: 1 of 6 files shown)


cmd/serverutil/serverutil.go
```diff
@@ -19,19 +19,38 @@ import (
...
+	"strings"
 
 	"github.com/golang/glog"
 	"github.com/grpc-ecosystem/grpc-gateway/runtime"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"google.golang.org/grpc"
 )
 
+// GrpcHandlerFunc returns an http.Handler that delegates to grpcServer on incoming gRPC
+// connections or otherHandler otherwise. Copied from cockroachdb.
```
RJPercival: Does their license allow you to copy their code?

cmd/serverutil/serverutil.go
```diff
@@ -19,19 +19,38 @@ import (
...
+func GrpcHandlerFunc(grpcServer http.Handler, otherHandler http.Handler) http.Handler {
```
RJPercival: nit: Go style dictates that "Grpc" should be all uppercase.

```suggestion
func GRPCHandlerFunc(grpcServer http.Handler, otherHandler http.Handler) http.Handler {
```

cmd/serverutil/serverutil.go
```diff
@@ -19,19 +19,38 @@ import (
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// This is a partial recreation of gRPC's internal checks.
+		// https://github.com/grpc/grpc-go/blob/master/transport/handler_server.go#L62
```
RJPercival: This link leads to a 404.
```suggestion
		// https://github.com/grpc/grpc-go/blob/v1.26.0/internal/transport/handler_server.go#L62
```

go.mod
```diff
@@ -46,7 +46,6 @@ require (
 	github.com/prometheus/procfs v0.0.7 // indirect
 	github.com/russross/blackfriday v2.0.0+incompatible // indirect
 	github.com/securego/gosec v0.0.0-20191119104125-df484bfa9e9f // indirect
-	github.com/soheilhy/cmux v0.1.4
```
RJPercival: nit: Could run `go mod tidy` to drop cmux from go.sum as well.

gdbelvin: go mod tidy doesn't produce any changes

RJPercival: That's surprising, I'd have thought it'd drop it from go.sum when it's no longer used.

PR #1443, opened 2020-01-28
Use K8 ingress for TLS termination

### LINKED ISSUE #1396: Use Let's Encrypt to get a real tls keypair

gdbelvin:
Use K8 ingress for TLS termination

Kubernetes ingress objects support
- Managed TLS Certificates
- Central monitoring and metrics
- Static IP address, which allows DNS setup for `sandbox.keytransparency.dev`.

The backend protocol from the ingress reverse proxy to the binaries use HTTP/2 + TLS

This PR uses `kustomize` to configure k8 configs for both GCE and baremetal deployments. The baremetal deployment is used for local testing and the kubernetes test in travis. Baremetal deployments must locally supply an ingress implementation, this PR uses `nginx`.  The GCE deployment uses the cloud provided GCE ingress implementation.  

I chose 'kustomize` over `helm` because kustomize has better support for multiple deployment environments.  We can revisit this decision later if needed. 

The directory structure:
```
deploy
├── kubernetes
│   ├── base
│   └── overlays
│       ├── gke // Google Cloud
│       └── local // Baremetal for local and travis tests
│           └── ingress-nginx
│               ├── baremetal // nginx configs required for non-cloud deployments
│               ├── cloud-generic // Basic nginx config required for most deployments
│               └── static  // Basic nginx config required for all nginx deployments
```

To verify that this PR worked:
```
$ ./scripts/deploy.sh
$ curl https://sandbox.keytransparency.dev -w '%{http_version}\n'
ok
2 
```
```
$ go install github.com/fullstorydev/grpcurl/cmd/grpcurl
$ grpcurl sandbox.keytransparency.dev:443 list
google.keytransparency.v1.KeyTransparency
grpc.reflection.v1alpha.ServerReflection
```

Fixes #1396


gdbelvin:
After researching `Helm`, I've decided to stick with Kustomize for the following reasons
1. Helm supports environment specific customization through Go Templating. This makes small changes possible, but larger changes quickly become convoluted. 
2. I couldn't see a way to optionally depend on subcharts in an environment specific manner. 
  - To support local testing, we need bare metal features that cloud service providers typically provide. 
3. The more complex uses cases for setting up let's encrypt certs can be managed by relying on GCE. This reduces the effectiveness of local testing, but the local testing story probably deserves a re-think.

deploy/kubernetes/base/server-service.yaml
```diff
@@ -9,11 +9,13 @@ spec:
...
     targetPort: 8080
+    nodePort: 30080
   - name: "http-metrics"
     port: 8081
     targetPort: 8081
+    nodePort: 30081
   selector:
     io.kompose.service: server
-  type: NodePort
+  type: NodePort # Required for ingress
```
NatalieDoduc: why was it already there, then?

deploy/kubernetes/base/kustomization.yaml
```diff
@@ -19,4 +19,5 @@ resources:
 - sequencer-service.yaml
 - server-deployment.yaml
 - server-service.yaml
+- ingress.yaml
```
NatalieDoduc: alphabetical order? here and elsewhere.

deploy/kubernetes/overlays/local/ingress-nginx/baremetal/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
NatalieDoduc: could you add a README explaining why there are 3 flavours under `ingress-nginx`/ and what the distinctions are? I can sort of guess by the name but i'd prefer to avoid the ambiguity.

I think you also mentioned that some of these files were copies of strongly inspired from other files? A link to those base files and the potential changes between base and these copies would be helpful. Maybe a hint in the source to remind to update these as well?

deploy/kubernetes/base/server-service.yaml
gdbelvin: It was already there in the base config because that's how kubernetes in docker connected to it.
It was previously overridden to be a `LoadBallancer` in GCE. I thought it would be good to document what depends on it being a `NodePort`

deploy/kubernetes/base/kustomization.yaml
gdbelvin: done

gdbelvin:
Thanks for your suggestions @NatalieDoduc 

Please see
- Updated `README.md`
  - Dir structure
  - Testing instructions
- A `README.md` for the `ingress-nginx` directory

deploy/kubernetes/base/server-service.yaml
NatalieDoduc: Thanks for the clarification - it's a good idea to explain why this type is required.

Now all flavours, including baremetal are `type: NodePort`, as I can't see it being overridden (well, kustomised). Correct?

deploy/kubernetes/overlays/local/ingress-nginx/README.md
```diff
@@ -0,0 +1,21 @@
...
+1. Split out into their component yaml files.
+2. Added `kustomization.yaml` files tying them together.
+3. Removed conflicting resources.
+
+The kustomize dependency graph looks like so:
+```
+overlays\local -> overlays\local\ingress-nginx\baremetal
+overlays\local\ingress-nginx\baremetal -> overlays\local\ingress-nginx\cloudgeneric
+overlays\local\ingress-nginx\cloudgeneric -> overlays\local\ingress-nginx\static
+```
```
NatalieDoduc: Just curious: Under which circumstances would one need to modify these files?

As discussed offline, let's see how this grows. It's certainly very valuable to have travis validate this in CI, but given the potential divergence between the deployment and this `local test`, we might not be validating much that overlaps with a live deployment. Let's keep in mind a full E2E test deployment.

scripts/kubernetes_test.sh
```diff
@@ -32,8 +32,8 @@ fi
 # Hack to wait for the default service account's creation. https://github.com/kubernetes/kubernetes/issues/66689
 n=0; until ((n >= 60)); do kubectl -n default get serviceaccount default -o name && break; n=$((n + 1)); sleep 1; done; ((n < 60))
 
-kubectl apply -k deploy/kubernetes/overlays/local 
-trap "kubectl delete -k deploy/kubernetes/overlays/local" INT EXIT
+kustomize build deploy/kubernetes/overlays/local/ | kubectl apply -f -
+trap "kustomize build deploy/kubernetes/overlays/local/ | kubectl delete -f -" INT EXIT
```
NatalieDoduc: with `delete` you're turning down the environment. is this what you want - in case you had already running env? no biggie, but just checking.

gdbelvin: That's right - these tests verify full setup / teardown with 1 command. :-)

deploy/kubernetes/overlays/local/ingress-nginx/README.md
gdbelvin: The only time we would want to update these files it to keep up-to-date with upstream changes. 

Agreed that the local test is becoming less useful, but let's keep an eye on it.

PR #1447, opened 2020-02-04
Remove old key generation scripts

gdbelvin:
Replace some very funky scripts with with golang's cleaner `generate_cert.go`


gdbelvin:
It's the same functionality, but much more readable / maintainable.

scripts/deploy.sh
```diff
@@ -40,9 +40,11 @@ test $(basename $(pwd)) == "keytransparency" || exit 1
 if ! kubectl get secret kt-tls; then
   echo "Generating keys..."
   rm -f ./genfiles/*
-  ./scripts/prepare_server.sh -f
+  ./scripts/gen_monitor_keys.sh -f
   kubectl create secret generic kt-monitor --from-file=genfiles/monitor_sign-key.pem
-  kubectl create secret tls kt-tls --cert=genfiles/server.crt --key=genfiles/server.key
+  go run "$(go env GOROOT)/src/crypto/tls/generate_cert.go" --host localhost,127.0.0.1,::
```
NatalieDoduc: with go modules now, is there a way to express this without needing to set the GOROOT?

gdbelvin: `go env GOROOT` does not require the `$GOROOT` environment variable to be set. 

GOROOT points to the go source tree, not the place where software is developed

PR #1451, opened 2020-02-11
Explain directories

### LINKED ISSUE #1207: Minor: Explain directories
The docker-compose `init` service creates a directory called `default` that may then be queried:

```
curl --insecure https://localhost/v1/directories/default
```

Subsequent `keytransparency-client post` commands implicitly use this `default` directory.

It was only through some experimentation that I realized, it's easy to create other directories and these may be referenced using the `directory` flag on the post command.

**Recommend**
Adding some clarity to the README.md about directories and how to use them.
Perhaps, with an example of adding a second directory and then POSTing to it?

gdbelvin:
Fixes #1207 - Explain directories

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,60 @@ development.
 
 ### Client operations
 
+## View a Directory's Public Keys
+The Key Transparency server publishes a separate set of public keys for each directory that it hosts.
+By hosting multiple directores, a single domain can host directories for multiple apps or customers.
+A standardized pattern for discovering domains and directores is a TODO in issue #389.
+
+Within a directory the server uses the following public keys to sign its responses:
+1. `log.public_key` signs the top-most merkle tree root, covering the ordered list of map roots.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,6 +37,60 @@ development.
...
+By hosting multiple directores, a single domain can host directories for multiple apps or customers.
+A standardized pattern for discovering domains and directores is a TODO in issue #389.
+
+Within a directory the server uses the following public keys to sign its responses:
+1. `log.public_key` signs the top-most merkle tree root, covering the ordered list of map roots.
+2. `map.public_key` signs each snapshot of the key-value database in the form of a sparse merkle tree.
+3. `vrf.der` signs outputs of the [Verifiable Random Function](https://en.wikipedia.org/wiki/Verifiable_random_function)
+    which obscures the key values in the key-value database.
+
+A directory's keys default can be retrieved over HTTPS/JSON with curl
```
Mercurrent:
```suggestion
A directory's keys can be retrieved over HTTPS/JSON with curl
```

PR #1460, opened 2020-02-14
Avoid race condition in election test

gdbelvin:
Avoid race condition in election test


RJPercival:
I recommend adding a `go mod tidy` test to Travis to stop your go.mod/go.sum lagging behind changes (see Trillian's Travis config for an example).

gdbelvin:
I've removed the mod file changes to clean up the PR.

PR #1473, opened 2020-02-25
Use Cloud MySQL in GKE

gdbelvin:


gdbelvin:
I believe this is working now :-)

deploy/kubernetes/overlays/local/kustomization.yaml
```diff
@@ -2,3 +2,5 @@ bases:
 - ../../base
 resources:
 - ingress-nginx/baremetal
+- db-deployment.yaml
+- db-service.yaml
```
Mercurrent: you removed/renamed this file, right?

PR #1479, opened 2020-02-28
Fix credential encryption

gdbelvin:
#1478 Messed this up by using a different file name than the original.


.travis.yml
```diff
@@ -53,8 +53,7 @@ jobs:
       install:
       - curl -LO "https://storage.googleapis.com/kubernetes-release/release/v1.16.3/bin/linux/amd64/kubectl" && chmod +x kubectl && sudo mv kubectl /usr/local/bin/
       - curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh" | bash && sudo mv kustomize /usr/local/bin/
-      - openssl aes-256-cbc -K $encrypted_555d9b2948d2_key -iv $encrypted_555d9b2948d2_iv
```
NatalieDoduc: I'm not really familiar with this but you can do multiline in yml by adding > , if that was something you wanted to achieve.

Also - given this is in a travis file, was there any way of testing the command beforehand?

gdbelvin: I'm not sure about `>` for multi-line yaml.  I'm more familiar with `-|` for code blocks in yaml.

Since this is a yaml file that only executes on master branch, it's difficult to test.

PR #1489, opened 2020-03-05
Use cases

gdbelvin:


docs/use-cases.md
```diff
@@ -0,0 +1,19 @@
+# Key Transparency Use Cases
+
+
+Key Transparency ensures, with mathematical certainty, that account owners can
+see all the public keys that have been associated with their account.  
+
+This zero-trust architecture is useful in a variety of scenarios from end-to-end encryption to enterprise account management.
```
pavelkalinnikov: Is it zero trust, or something a bit weaker? Some sort of "trust but verify" thing?
Just want to make sure we don't promise too much here.

docs/use-cases.md
```diff
@@ -0,0 +1,19 @@
+# Key Transparency Use Cases
+
+
+Key Transparency ensures, with mathematical certainty, that account owners can
+see all the public keys that have been associated with their account.
```
pavelkalinnikov: Would it be worth elaborating a bit on what "associated" means? Like, some other user using my key is included here, right?

docs/use-cases.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Scenarios 
+
+|  **Scenario** |  **Description** | 
+|---------------|------------------|
+| Encrypted Messaging | Key Transparency is ideal for user friendly end-to-end encrypted messaging.  By making key management analogous to device management, users do not have to learn any new concepts, and no additional UI beyond device management is needed. Key management fits seamlessly into existing account life-cycle flows, and users are protected without requiring them to take additional actions. |
+| PGP Encrypted Email | Key Transparency was initially built to solve the problem of public key lookup for PGP email encryption. KT has the potential to make PGP significantly more usable than the existing web-of-trust model. |
+| Insider Risk        | Key Transparency removes the possibility that a system administrator, whether accidentally, on purpose, or via a hack, might add or remove public keys (eg. [U2F Security Keys](https://en.wikipedia.org/wiki/Universal_2nd_Factor)) from a user account without detection. This enables new levels of security in authentication systems. |
```
pavelkalinnikov:
```suggestion
| Insider Risk        | Key Transparency removes the possibility that a system administrator, whether accidentally, on purpose, or via a hack, adds or removes public keys (eg. [U2F Security Keys](https://en.wikipedia.org/wiki/Universal_2nd_Factor)) from a user account without detection. This enables new levels of security in authentication systems. |
```

docs/use-cases.md
gdbelvin: Good point. How's "assured system design"?

docs/use-cases.md
gdbelvin: How's "see all the public keys that have been used to authenticate their account or send them messages." ?

PR #1504, opened 2020-03-25
Use error wrapping in backoff

gdbelvin:
Preserve error details when indicating that an error should
be retried.

### DIFF
--- a/internal/backoff/backoff.go
+++ b/internal/backoff/backoff.go
@@ -17,6 +17,7 @@ package backoff
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math/rand"
 	"time"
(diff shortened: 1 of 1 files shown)


internal/backoff/backoff.go
```diff
@@ -26,16 +27,13 @@ import (
...
 
 // RetriableError explicitly instructs Backoff to retry.
-type RetriableError string
-
-// Error returns string representation of the retriable error.
-func (re RetriableError) Error() string {
-	return string(re)
+type RetriableError struct {
+	error
 }
```
pavelkalinnikov: Would `type RetriableError error` work, or you would need to explicitly add the `Error()` method?

gdbelvin: I did try that first.  It breaks `errors.As` unfortunately.

PR #1505, opened 2020-03-25
Fix race condition

gdbelvin:
Masterships calls Resign


pavelkalinnikov:
Was this closed on purpose, or needs review?

gdbelvin:
I closed it on purpose.  I was trying to solve a race condition that rarely
occurs in the test.
I still can't find it, but I convinced myself that this was not the right
fix.

PR #1508, opened 2020-03-27
Rename impl/sql -> impl/mysql

gdbelvin:
More accurate package path.
The implementation is mysql specific, not generic sql.


gdbelvin:
KT is very explicitly v0.2-alpha

PR #1522, opened 2020-04-27
ECVRF and ECVRFParam Structure

gdbelvin:
This PR sets out the rough structure of the ECVRF implementation:

- `ECVRFParams` holds private variables that are used to implement the elliptic curve VRF suites.
- `ECVRFAux` is an interface to reference suite specific algorithms
-  `p256SHA256TAI` is the concrete instance of `ECVRFParams` that implements the `P256_SHA256_TAI` VRF.


gdbelvin:
Superseded by https://github.com/google/draft-irtf-cfrg-vrf/pull/4

PR #1535, opened 2020-05-07
Mutation table Spanner impl

gdbelvin:
Interesting note: Spanner Timestamps are `time.Time` objects -- even though they only have microsecond precision underneath. 

cc @pavelkalinnikov


gdbelvin:
Blocked on https://github.com/googleapis/google-cloud-go/issues/2195

impl/spanner/mutations/table.go
```diff
@@ -0,0 +1,311 @@
...
+			DirectoryID: directoryID,
+			LogID:       logID,
+			Timestamp:   spanner.CommitTimestamp,
+			LocalID:     int64(i),
+			Mutation:    mBytes,
+		})
+		if err != nil {
+			return water.Mark{}, err
+		}
+		ms = append(ms, m)
```
mhutchinson, line 171: Is append strongly preferred in Go over direct index assignment? I would expect `ms` to be created with length & cap as `len(entries)`, then this line would be `ms[i] = m`. Seems more efficient and clearer, but perhaps less idiomatic?

impl/spanner/mutations/table.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+
+	commitTimestamp, err := t.client.ReadWriteTransaction(ctx,
+		func(_ context.Context, txn *spanner.ReadWriteTransaction) error {
+			return txn.BufferWrite(ms)
+		})
+	return timeToMark(commitTimestamp), err
```
mhutchinson, line 178: You may have investigated this and decided it's safe, but calling a function on something which may have been returned with an error seems like it might cause failures? Perhaps timeToMark works on any possible value of time.Time in which case discard.

impl/spanner/mutations/table.go
```diff
@@ -0,0 +1,311 @@
...
+}
+
+// HighWatermark returns the number of items and the highest primary
+// key (exclusive) up to batchSize items after start (inclusive).
+func (t *Table) HighWatermark(ctx context.Context, directoryID string, logID int64, start water.Mark, batchSize int32) (count int32, high water.Mark, err error) {
+	// TODO: Replace with MAX(Revision) when spansql supports aggregate operators.
```
mhutchinson, line 184: There isn't a Revision column, so this comment worries me that I'm missing something big? Or is this a copy/paste and should say Timestamp?

impl/spanner/mutations/table.go
```diff
@@ -0,0 +1,311 @@
...
+	stmt := spanner.NewStatement(`
+	SELECT Timestamp FROM Mutations WHERE
```
mhutchinson, line 186: SQL nit: WHERE statement on the next line? I like to format as:

```
SELECT
  col1,
  col2
FROM
  table
WHERE
  condition1 AND
  condition2
LIMIT N
```

impl/spanner/mutations/table.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+
+	commitTimestamp, err := t.client.ReadWriteTransaction(ctx,
+		func(_ context.Context, txn *spanner.ReadWriteTransaction) error {
+			return txn.BufferWrite(ms)
+		})
+	return timeToMark(commitTimestamp), err
+}
+
+// HighWatermark returns the number of items and the highest primary
```
mhutchinson, line 181: This says it returns the highest primary key, but the primary key is defined to be `PRIMARY KEY(DirectoryID, LogID, Timestamp, LocalID)` and this doesn't allow the caller to understand the LocalID. Should it return that too, or should the comment veer away from primary key terminology?

impl/spanner/mutations/table.go
```diff
@@ -0,0 +1,311 @@
...
+		// When there are no rows, return the start time as the highest timestamp.
+		return 0, start, nil
+	}
+	return int32(cnt), timeToMark(max).Add(1), nil
+}
+
+// ReadLog reads all mutations between [low, high).
+func (t *Table) ReadLog(ctx context.Context, directoryID string, logID int64, low, high water.Mark,
+	limit int32) ([]*mutator.LogMessage, error) {
+	if high.Value() == 0 || limit == 0 {
```
mhutchinson, line 225: Similar to above, I worry that this would be an application bug and should fail fast instead of limping on. But this is a decision that should be consistent across the library and you're in the right position to judge that.

impl/spanner/mutations/table.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+
+	msgs := make([]*mutator.LogMessage, 0)
+	rtx := t.client.ReadOnlyTransaction().
+		WithTimestampBound(spanner.MinReadTimestamp(markToTime(high)))
+	defer rtx.Close()
+	if err := rtx.ReadWithOptions(ctx, mutTable,
+		spanner.KeyRange{
+			Kind:  spanner.ClosedOpen,
+			Start: spanner.Key{directoryID, logID, markToTime(low), 0},
```
mhutchinson, line 236: This works, but there's a funny relationship with this LocalID. It's data type is int64 so it could be negative. But it looks like we know it's constrained to uint range because we generate it in this class.

I wonder if the data type should be uint, or this filter could pass in the smallest int64. Or we leave as-is and accept the minor logical tangle here.

impl/spanner/mutations/table.go
```diff
@@ -0,0 +1,311 @@
...
+}
+
+// ReadLog reads all mutations between [low, high).
+func (t *Table) ReadLog(ctx context.Context, directoryID string, logID int64, low, high water.Mark,
+	limit int32) ([]*mutator.LogMessage, error) {
+	if high.Value() == 0 || limit == 0 {
+		return []*mutator.LogMessage{}, nil
+	}
+
+	msgs := make([]*mutator.LogMessage, 0)
```
mhutchinson, line 229: There's a space/time tradeoff to be made here depending on what likely values are for `limit` and what the fill ratio is likely to be. You could provide `limit` as the capacity to save less time reallocating the underlying array, but risking it being mostly empty in the case where `limit` is huge and the rows read are small.

impl/spanner/mutations/table.go
```diff
@@ -0,0 +1,311 @@
...
+			msg, err := unmarshalRow(r, logID)
+			if err != nil {
+				return err
+			}
+			msgs = append(msgs, msg)
+			return nil
+		}); err != nil {
+			return nil, err
+		}
+	}
```
mhutchinson, line 282: This whole block is surprising to me. I would expect `limit` to be the max number to return, but it appears to not be that. I think some doc string on what it means would be helpful. Or this is an indication that I need to eat now :-)

impl/spanner/mutations/table.go
```diff
@@ -0,0 +1,311 @@
...
+	"github.com/google/keytransparency/core/mutator"
+	"github.com/google/keytransparency/core/water"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/proto"
+
+	pb "github.com/google/keytransparency/core/api/v1/keytransparency_go_proto"
+)
+
+// readStaleness is the maximum staleness of HighWatermarks.
```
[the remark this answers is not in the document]
gdbelvin: readStaleness gives Spanner the opportunity to read from follower servers rather than only master.  I can add a bit to the comment.

impl/spanner/mutations/table.go
```diff
@@ -0,0 +1,311 @@
...
+	rtx := t.client.Single()
+	defer rtx.Close()
+
+	m, err := spanner.UpdateStruct(logTable, LogTableCols{
+		DirectoryID: directoryID,
+		LogID:       logID,
+		WriteToLog:  enabled,
+	})
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
gdbelvin: Yes, it's good practice in go, especially now that we have error wrapping with `%w`

impl/spanner/mutations/table.go
gdbelvin: I think append is preferred with a slice that has the right capacity. 
I understand the instinct to do direct indexing (I started that way), but the team uses append.

impl/spanner/mutations/table.go
gdbelvin: Should say timestamp

impl/spanner/mutations/table.go
gdbelvin: I suppose I'm coming from a maths perspective where we define things for edge cases like "1 is a prime"

impl/spanner/mutations/table.go
gdbelvin: Cloud Spanner only supports `int64` for ints

impl/spanner/mutations/table.go
gdbelvin: limit is a good call

impl/spanner/mutations/table.go
gdbelvin: This certainly deserves documentation. 

This is an edge case I'm not super happy with, but fixing it requires a bit of an overhaul -- and would reduce our ability to control traffic to the map.

impl/spanner/keytransparency.ddl
```diff
@@ -33,11 +33,11 @@ CREATE TABLE Batches(
 ) PRIMARY KEY (DirectoryID, Revision),
   INTERLEAVE IN PARENT Directories ON DELETE CASCADE;
 
--- UnsequencedMutations
-CREATE TABLE UnsequencedMutations (
+-- Mutations
```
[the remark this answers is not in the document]
gdbelvin: removed

impl/spanner/mutations/table.go
```diff
@@ -0,0 +1,311 @@
...
+	nanos := int64(w.Value()) * int64(tsRes)
+	return time.Unix(0, nanos)
+}
+func timeToMark(t time.Time) water.Mark {
+	m := t.Truncate(tsRes)
+	v := uint64(m.UnixNano()) / uint64(tsRes)
+	return water.NewMark(v)
+}
+
+// Table implements a rough queue of mutations.
```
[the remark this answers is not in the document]
gdbelvin: This comment is out of date -- only applied to spanner queues.

impl/spanner/mutations/table.go
```diff
@@ -0,0 +1,311 @@
...
+		return nil, status.Errorf(codes.NotFound, "not found")
+	}
+	return logIDs, nil
+}
+
+// Send submits an item to the queue and returns the commit timestamp.
+func (t *Table) Send(ctx context.Context, directoryID string, logID int64, entries ...*pb.EntryUpdate) (water.Mark, error) {
+	if len(entries) == 0 {
+		return water.Mark{}, nil
+	}
```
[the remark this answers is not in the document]
gdbelvin: Sounds good

impl/spanner/mutations/table.go
gdbelvin: good call

impl/spanner/mutations/table.go
gdbelvin: done

impl/spanner/mutations/table.go
gdbelvin: Limit = 0 and high =0 are valid values to pass in, even though they might indicative of logic errors higher in the stack. Because this function doesn't "know" for sure that they are logic errors, they should not be treated as such here. Happy to discuss this

gdbelvin:
Thanks for your comments! 

As a high level observation, there are parts of this API that I'm less than 100% pleased with, but because they are part of the storage layer API contract, I can't change them here. 

The objective of this PR is to get a spanner implementation working. Once that's in place, we can go back and do a better review of the API contracts as a separate work item.

impl/spanner/mutations/table.go
mhutchinson: ack.

impl/spanner/mutations/table_test.go
```diff
@@ -0,0 +1,241 @@
...
+		t.Fatalf("AddLogs(): %v", err)
+	}
+	time.Sleep(readStaleness)
+
+	return q
+}
+
+func TestNewForTest(t *testing.T) {
+	ctx := context.Background()
+	NewForTest(ctx, t, "NewForTest", 1)
```
mhutchinson, line 74: Supernit: the other directory IDs are all lowercase. Can they be consistent in one direction or the other?

impl/spanner/mutations/table.go
```diff
@@ -0,0 +1,320 @@
...
+		}); err != nil {
+		return nil, err
+	}
+	if len(logIDs) == 0 {
+		return nil, status.Errorf(codes.NotFound, "not found")
+	}
+	return logIDs, nil
+}
+
+// Send submits an item to the queue and returns the commit timestamp.
```
mhutchinson, line 143:
```suggestion
// Send submits a batch of items to the queue and returns the commit timestamp.
```

This concept of a batch is important as it comes up below when returning oversized batches.

impl/spanner/mutations/table.go
```diff
@@ -0,0 +1,320 @@
...
+	}
+	if cnt == 0 {
+		// When there are no rows, return the start time as the highest timestamp.
+		return 0, start, nil
+	}
+	return int32(cnt), timeToMark(max).Add(1), nil
+}
+
+// ReadLog reads all mutations between [low, high).
+// If limit is used, ReadLog ensures that complete batches are returned, which could slightly exceed limit.
```
mhutchinson, line 231:
```suggestion
// If limit is used, ReadLog ensures that complete batches are returned, which could exceed limit.
```

From the perspective of this library, there is no control over how big or small this could be.

May be worth adding clarification that a batch is all entries with the same timestamp, because that wasn't obvious to me.

impl/spanner/mutations/table.go
```diff
@@ -0,0 +1,320 @@
...
+	for _, m := range msgs {
+		if m.ID.Value() > maxTs.Value() {
+			maxTs = m.ID
+			maxLocal = 0
+		}
+		if m.ID.Value() == maxTs.Value() &&
+			m.LocalID > maxLocal {
+			maxLocal = m.LocalID
+		}
+	}
```
mhutchinson, line 271: Looks like this can live in the `if` block below?

impl/spanner/mutations/table.go
mhutchinson: I understand this now and think it's reasonable. I think the missing concept for me was that of the "batch". It almost seems like Send should be called WriteBatch.

impl/spanner/mutations/table_test.go
```diff
@@ -0,0 +1,241 @@
...
+func TestReadBatch(t *testing.T) {
+	const dirID = "readbatch"
+	ctx := context.Background()
+	logID := int64(5)
+	m := NewForTest(ctx, t, dirID, logID)
+	var lastTS water.Mark
+	var err error
+	for i := byte(0); i < 10; i++ {
+		entry := &pb.EntryUpdate{Mutation: &pb.SignedEntry{Entry: mustMarshal(t, &pb.Entry{Index: []byte{i}})}}
+		lastTS, err = m.Send(ctx, dirID, logID, entry, entry, entry)
```
mhutchinson, line 100: I spent over a minute trying to work out why reading one thing led to 3 results. The fact that this has entry repeated is obvious when you see it, and subtle when you don't. Can we pop a comment here that simply says "Write the same entry multiple times to construct a batch" or similar?

impl/spanner/mutations/table_test.go
```diff
@@ -0,0 +1,241 @@
...
+		}
+	}
+
+	for _, tc := range []struct {
+		batchSize int32
+		count     int
+	}{
+		{batchSize: 0, count: 0},
+		{batchSize: 1, count: 3},
+		{batchSize: 4, count: 6}, // A partial read into the next timestamp returns all rows in the timestamp.
```
mhutchinson, line 112: Missing testcase of `batchSize: 3, count: 3`. It looks like it will work, but it's a classic edge case.

impl/spanner/mutations/table_test.go
gdbelvin: Added a "batch" variable to clarify this.  I'll also add a todo to remove the variable size param

PR #1550, opened 2020-05-26
Update README cert generation instructions

gdbelvin:
`scripts/preare_server.sh` was replaced with  `$(go env
GOROOT)/src/crypto/tls/generate_cert.go` in #1447

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,7 +178,9 @@ Create or fetch the public key for your specific application.
 go get github.com/google/keytransparency/...
 go get github.com/google/trillian/...
 cd $(go env GOPATH)/src/github.com/google/keytransparency
-./scripts/prepare_server.sh -f
+pushd genfiles
+go run "$(go env GOROOT)/src/crypto/tls/generate_cert.go" --host localhost,127.0.0.1,::
+popd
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -178,7 +178,9 @@ Create or fetch the public key for your specific application.
 go get github.com/google/keytransparency/...
 go get github.com/google/trillian/...
 cd $(go env GOPATH)/src/github.com/google/keytransparency
-./scripts/prepare_server.sh -f
+pushd genfiles
+go run "$(go env GOROOT)/src/crypto/tls/generate_cert.go" --host localhost,127.0.0.1,::
+popd
```
mhutchinson, line 183: `pushd` and `popd' are not standard bash commands. I think they are standard for zsh, but I'd prefer to stick to standard built-ins for this.

gdbelvin: would `cd dir && cd -` be preferable?

btw -- I think `pushd` is part of bash.
https://linux.die.net/man/1/pushd

mhutchinson: You're right. TIL.

mhutchinson:
LGTM.

PR #1553, opened 2020-05-29
Enable tests with Spanner Emulator and GCP

gdbelvin:


impl/spanner/mutations/table_test.go
```diff
@@ -139,7 +143,7 @@ func TestReadLog(t *testing.T) {
 		low, high water.Mark
 		want      int
 	}{
-		{desc: "read nothing", low: ts1, high: ts1, want: 0},
+		// {desc: "read nothing", low: ts1, high: ts1, want: 0}, // Not supported by Cloud Spanner
```
cindylindeed, line 146: Delete entire line?

gdbelvin: I'll keep it around for when they implement this "feature"

PR #1558, opened 2020-06-05
Storage Factory

gdbelvin:
Support creating storage implementations of various types.

### DIFF
--- a/cmd/keytransparency-sequencer/main.go
+++ b/cmd/keytransparency-sequencer/main.go
@@ -37,14 +37,12 @@ import (
 	"github.com/google/keytransparency/core/adminserver"
 	"github.com/google/keytransparency/core/sequencer"
 	"github.com/google/keytransparency/core/sequencer/election"
-	"github.com/google/keytransparency/impl/mysql/directory"
-	"github.com/google/keytransparency/impl/mysql/mutationstorage"
+	"github.com/google/keytransparency/impl"
 	"github.com/google/keytransparency/internal/forcemaster"
 
 	pb "github.com/google/keytransparency/core/api/v1/keytransparency_go_proto"
 	dir "github.com/google/keytransparency/core/directory"
 	spb "github.com/google/keytransparency/core/sequencer/sequencer_go_proto"
-	ktsql "github.com/google/keytransparency/impl/mysql"
 	etcdelect "github.com/google/trillian/util/election2/etcd"
 	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
 
@@ -63,7 +61,8 @@ var (
 	etcdServers = flag.String("etcd_servers", "", "A comma-separated list of etcd servers; no etcd registration if empty")
 	lockDir     = flag.String("lock_file_path", "/keytransparency/master", "etcd lock file directory path")
 
-	serverDBPath = flag.String("db", "db", "Database connection string")
+	dbPath   = flag.String("db", "", "Database connection string")
+	dbEngine = flag.String("db_engine", "mysql", "Storage implementation. One of ['mysql', 'spanner']")
 
 	// Info to connect to the trillian map and log.
 	mapURL = flag.String("map-url", "", "URL of Trillian Map Server")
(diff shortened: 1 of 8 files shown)


cmd/keytransparency-sequencer/main.go
```diff
@@ -63,7 +61,8 @@ var (
 	etcdServers = flag.String("etcd_servers", "", "A comma-separated list of etcd servers; no etcd registration if empty")
 	lockDir     = flag.String("lock_file_path", "/keytransparency/master", "etcd lock file directory path")
 
-	serverDBPath = flag.String("db", "db", "Database connection string")
+	dbPath   = flag.String("db", "", "Database connection string")
+	dbEngine = flag.String("db_engine", "mysql", "Storage implementation. One of ['mysql', 'spanner']")
```
mhutchinson, line 65: There's a bit of a layer violation here that the supported types are  hardcoded into this docstring but the impl switching is in another file. I don't have suggestions to improve this without making things more complicated though so happy to leave this if you are.

cmd/keytransparency-sequencer/main.go
```diff
@@ -147,11 +137,11 @@ func main() {
 	defer done()
 
 	spb.RegisterKeyTransparencySequencerServer(grpcServer, sequencer.NewServer(
-		directoryStorage,
+		db.Directories,
 		trillian.NewTrillianLogClient(lconn),
 		trillian.NewTrillianMapClient(mconn),
 		trillian.NewTrillianMapWriteClient(mconn),
-		mutations, mutations,
+		db.Batches, db.Logs,
```
mhutchinson, line 144: nit: I feel like separate lines for these 2 params would be more consistent. Unless I'm missing something implicit about why they're on the same line?

cmd/keytransparency-server/main.go
```diff
@@ -84,14 +83,6 @@ func main() {
 	}
 
 	// Create database and helper objects.
```
mhutchinson, line 85: Unless the diff display is lying to me, this comment is obsolete.

impl/storage.go
```diff
@@ -0,0 +1,107 @@
...
+	"context"
+	"fmt"
+
+	"cloud.google.com/go/spanner"
+	"github.com/google/keytransparency/core/sequencer"
+	"github.com/google/keytransparency/core/water"
+	"github.com/google/keytransparency/impl/mysql/mutationstorage"
+	"github.com/google/keytransparency/impl/spanner/batch"
+	"github.com/google/keytransparency/impl/spanner/directory"
+	"github.com/google/keytransparency/impl/spanner/mutations"
```
mhutchinson, line 27: Can we import these with custom names? In the code below it should be clearer which are mysql and which are spanner. `batch` doesn't help me.

impl/storage.go
```diff
@@ -0,0 +1,107 @@
...
+		// Copied methods from adminserver.LogsAdmin because of duplicate method.
+
+		// AddLogs creates and adds new logs for writing to a directory.
+		AddLogs(ctx context.Context, directoryID string, logIDs ...int64) error
+		// SetWritable enables or disables new writes from going to logID.
+		SetWritable(ctx context.Context, directoryID string, logID int64, enabled bool) error
+
+		// Copied methods from keyserver.MutationLogs
+		SendBatch(ctx context.Context, directoryID string, logID int64, batch []*pb.EntryUpdate) (water.Mark, error)
+	}
```
mhutchinson, line 50: Can you get a Go Native to review this? It looks a bit clunky to me.

cmd/keytransparency-sequencer/main.go
gdbelvin: Added a `StorageEngines()` function for this

cmd/keytransparency-sequencer/main.go
gdbelvin: You're right.

impl/storage.go
gdbelvin: Good idea. done

impl/storage.go
gdbelvin: This is a bit of a known issue: https://github.com/golang/go/issues/6977
https://golang.org/ref/spec#Uniqueness_of_identifiers

I think these could be merged if the interfaces were in the same package, because they are in different packages, they are different identifiers.

impl/mysql/open.go
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 // Package sql provides functions for interacting with MySQL.
```
mhutchinson, line 15:
```suggestion
// Package mysql provides functions for interacting with MySQL.
```

PR #1566, opened 2020-06-25
Meet in the middle algorithm

gdbelvin:
Part of KT 2.0

### DIFF
--- a/core/client/MeetInTheMiddle.md
+++ b/core/client/MeetInTheMiddle.md
@@ -1,0 +1,43 @@
+# Meet In The Middle Algorithm
+
+### Purpose
+Enable a **sender** Bob at revision *fetched* and a **receiver** Alice at revision *current* to perform 
+append-only proofs between revisions *fetched* <= *intermediate* <= *current*, without coordinating.
+
+### Background
+Key Transparency 1.0 allowed Bob to fetch at an unknown revision, requiring
+Alice to fetch *all* revisions in order to verify her key history, an O(n)
+amount of work where n is the number of revisonis between verifications. This
+limited the frequency and freshness of directory updates.
+
+The meet in the middle algorithm shares the work between Alice and Bob by selecting Log(n) 
+potential intermediates, with the assurance that one of them will be shared between Alice and Bob.
+This brings the verification work down to a level that is Log(n) efficient, regardless of directory freshness. 
+
+## Algorithm
+
+### Sender Verificatoin
+Clients save the (earliest) revision number associated with the data they have
+fetched.  For Key Transparency, this means that *senders* will store, for each
+contact, the contact's public key, map revision number it was fetched at, and
+the latest revision it has been verified against.
+
+Senders will then periodically verify that the map has not deleted evidence of
+the public key they are relying on by querying the map at select revisions and
+verifying that the contact's history of public keys is append-only.
+
+The revisions to verify are as follows:
+
+   revⁱ = fetched- fetched mod 2ⁱ+ 2ⁱ
+
+### Receiver Verification
+
+Receivers save, for their own keys, the latest revision they have verified.
+
+Receivers periodically (Tuneable to their own security tollarance) verify that
+the current version of their key history contains all keys previously contained
+in the map at select previous versions.  
+
+The revisions to verify are as follows:
+
+   revⁱ = current - current mod 2ⁱ
--- a/core/client/meetinthemiddle.go
+++ b/core/client/meetinthemiddle.go
@@ -22,8 +22,9 @@ package client
 // common.
 //
 // `created` is the (earliest) revision a data item was found in the map.
-//    It is OK to use the revision at which an item was fetched, but the algorithm will be more efficent
-//    if it starts with the revision at which an item was submitted to the map.
+//    It is OK to use the revision at which an item was fetched, but the
+//    algorithm will be more efficient if it starts with the revision at which
+//    an item was submitted to the map.
 // `verified` is the latest revision of a successfully verified consistency proof by this client.
 //    When verified > created, previously verified revisions are omitted from the results.
 // `current` is the current revision of the map. Selected revisions will be <= current.
--- a/core/client/meetinthemiddle.go
+++ b/core/client/meetinthemiddle.go
@@ -1,0 +1,87 @@
+// Copyright 2020 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package client
+
+// Meet in the middle algorithm:
+//
+
+// NewerRevisionsToVerify returns a list of revisions to perform append-only verifications against.
+// The relying party / sender is assured that if there is an entity running the
+// OlderRevisionsToVerify algorithm, they will have at least on revision in
+// common.
+//
+// `created` is the (earliest) revision a data item was found in the map.
+//    It is OK to use the revision at which an item was fetched, but the algorithm will be more efficent
+//    if the algorithm starts with the revision at which an item was submitted to the map.
+// `verified` is the latest revision of a successfully verified consistency proof by this client.
+//    When verified > created, previously verified revisions are omitted from the results.
+// `current` is the current revision of the map. Selected revisions will be <= current.
+func NewerRevisionsToVerify(created, current, verified int64) []int64 {
+	ret := []int64{} // bⁱ = b-b mod 2ⁱ+ 2ⁱ
+	if created < 1 {
+		created = 1 // Revision 0 is empty.
+	}
+	if current < 1 {
+		return []int64{} // Nothing to verify if the map is empty.
+	}
+	if created > current {
+		created = current
+	}
+	for r := created; r <= current && r > 0; r = next(r) {
+		// check = created + 2ⁱ - created mod 2ⁱ
+		if r <= verified {
+			continue
+		}
+		if r > current {
+			break
+		}
+		ret = append(ret, r)
+	}
+	return ret
+}
+
+// OlderRevisionsToVerify returns a list of revisions to perform append-only verifications against.
+// The provider / data owner / data verifier periodically verifies that their data history contained in
+// the current map revision contains all the data committed to in select previous revisions.
+// The provider is assured that any relying party has verified that they data
+// they have fetched is represented in at least one of the selected revisions.
+//
+// `current` is the current revision of the map.
+// `verified` is the latest verified revision for this particular data item.
+//
+// TODO: Consier limiting the max period between checks to the expected value
+// of other party lifetimes -- the maximum amount of time this algorithm can
+// tollerate the other party being offline at an unknown point in the past.
+// Currently this algorithm expects the relying party to keep it's memory and
+// be online at least once every `current = 2ⁱ` revisions. See Tests.
+func OlderRevisionsToVerify(current, verified int64) []int64 {
+	ret := []int64{} // aⁱ = a - a mod 2ⁱ
+	if current < 1 {
+		return []int64{}
+	}
+	for r := current; r >= 0; r = prev(r) {
+		if r <= verified {
+			break
+		}
+		ret = append(ret, r)
+	}
+	return ret
+}
+
+// prev return r-2ˡᵉᵛᵉˡ⁽ʳ⁾
+func prev(rev int64) int64 { return rev & (rev - 1) }
+
+// next return r+2ˡᵉᵛᵉˡ⁽ʳ⁾
+func next(rev int64) int64 { return (rev | (rev - 1)) + 1 }
--- a/core/client/meetinthemiddle.go
+++ b/core/client/meetinthemiddle.go
@@ -18,7 +18,7 @@ package client
 
 // NewerRevisionsToVerify returns a list of revisions to perform append-only verifications against.
 // The relying party / sender is assured that if there is an entity running the
-// OlderRevisionsToVerify algorithm, they will have at least on revision in
+// OlderRevisionsToVerify algorithm, they will have at least one revision in
 // common.
 //
 // `created` is the (earliest) revision a data item was found in the map.
@@ -33,11 +33,7 @@ func NewerRevisionsToVerify(created, current, verified uint64) []uint64 {
 	if created < 1 {
 		created = 1 // Revision 0 is empty.
 	}
-	if created > current {
-		panic("created > current")
-	}
-	for r := created; r <= current && r > 0; r = next(r) {
-		// check = created + 2ⁱ - created mod 2ⁱ
+	for r := created; r <= current; r = next(r) {
 		if r <= verified {
 			continue
 		}
--- a/core/client/meetinthemiddle.go
+++ b/core/client/meetinthemiddle.go
@@ -33,7 +33,7 @@ func NewerRevisionsToVerify(created, current, verified uint64) []uint64 {
 		created = 1 // Revision 0 is empty.
 	}
 	if created > current {
-		created = current
+		panic("created > current")
 	}
 	for r := created; r <= current && r > 0; r = next(r) {
 		// check = created + 2ⁱ - created mod 2ⁱ
--- a/core/client/meetinthemiddle.go
+++ b/core/client/meetinthemiddle.go
@@ -1,4 +1,4 @@
-// Copyright 2020 Google Inc. All Rights Reserved.
+// Copyright 2020 Google LLC. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -14,8 +14,7 @@
 
 package client
 
-// Meet in the middle algorithm:
-//
+// Meet in the middle algorithm: docs/meet-in-the-middle.md
 
 // NewerRevisionsToVerify returns a list of revisions to perform append-only verifications against.
 // The relying party / sender is assured that if there is an entity running the
@@ -24,17 +23,14 @@ package client
 //
 // `created` is the (earliest) revision a data item was found in the map.
 //    It is OK to use the revision at which an item was fetched, but the algorithm will be more efficent
-//    if the algorithm starts with the revision at which an item was submitted to the map.
+//    if it starts with the revision at which an item was submitted to the map.
 // `verified` is the latest revision of a successfully verified consistency proof by this client.
 //    When verified > created, previously verified revisions are omitted from the results.
 // `current` is the current revision of the map. Selected revisions will be <= current.
-func NewerRevisionsToVerify(created, current, verified int64) []int64 {
-	ret := []int64{} // bⁱ = b-b mod 2ⁱ+ 2ⁱ
+func NewerRevisionsToVerify(created, current, verified uint64) []uint64 {
+	ret := []uint64{} // b[i] = b-b mod 2ⁱ+ 2ⁱ
 	if created < 1 {
 		created = 1 // Revision 0 is empty.
-	}
-	if current < 1 {
-		return []int64{} // Nothing to verify if the map is empty.
 	}
 	if created > current {
 		created = current
@@ -44,18 +40,16 @@ func NewerRevisionsToVerify(created, current, verified int64) []int64 {
 		if r <= verified {
 			continue
 		}
-		if r > current {
-			break
-		}
 		ret = append(ret, r)
 	}
 	return ret
 }
 
-// OlderRevisionsToVerify returns a list of revisions to perform append-only verifications against.
-// The provider / data owner / data verifier periodically verifies that their data history contained in
-// the current map revision contains all the data committed to in select previous revisions.
-// The provider is assured that any relying party has verified that they data
+// OlderRevisionsToVerify returns a list of revisions to perform append-only
+// verifications against.  The provider / data owner / data verifier
+// periodically verifies that their data history contained in the current map
+// revision contains all the data committed to in select previous revisions.
+// The provider is assured that any relying party has verified that the data
 // they have fetched is represented in at least one of the selected revisions.
 //
 // `current` is the current revision of the map.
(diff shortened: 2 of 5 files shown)


core/client/MeetInTheMiddle.md
```diff
@@ -0,0 +1,43 @@
+# Meet In The Middle Algorithm
```
pavelkalinnikov, line 1: Link this doc from the KT 2 design doc?

pavelkalinnikov, line 1: Consider moving this to the design docs folder, and adding a link to it from the code comments.

core/client/meetinthemiddle.go
```diff
@@ -0,0 +1,87 @@
...
+//
+
+// NewerRevisionsToVerify returns a list of revisions to perform append-only verifications against.
+// The relying party / sender is assured that if there is an entity running the
+// OlderRevisionsToVerify algorithm, they will have at least on revision in
+// common.
+//
+// `created` is the (earliest) revision a data item was found in the map.
+//    It is OK to use the revision at which an item was fetched, but the algorithm will be more efficent
+//    if the algorithm starts with the revision at which an item was submitted to the map.
```
pavelkalinnikov, line 27:
```suggestion
//    if it starts with the revision at which an item was submitted to the map.
```

core/client/meetinthemiddle.go
```diff
@@ -0,0 +1,87 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package client
+
+// Meet in the middle algorithm:
+//
```
pavelkalinnikov, line 18: Add a link to the doc here?

core/client/meetinthemiddle.go
```diff
@@ -0,0 +1,87 @@
...
+//    When verified > created, previously verified revisions are omitted from the results.
+// `current` is the current revision of the map. Selected revisions will be <= current.
+func NewerRevisionsToVerify(created, current, verified int64) []int64 {
+	ret := []int64{} // bⁱ = b-b mod 2ⁱ+ 2ⁱ
+	if created < 1 {
+		created = 1 // Revision 0 is empty.
+	}
+	if current < 1 {
+		return []int64{} // Nothing to verify if the map is empty.
+	}
```
pavelkalinnikov, line 38: This can be removed. The loop below won't add anything if `current <= 0`.

```suggestion
```

core/client/meetinthemiddle.go
```diff
@@ -0,0 +1,87 @@
...
+		created = current
+	}
+	for r := created; r <= current && r > 0; r = next(r) {
+		// check = created + 2ⁱ - created mod 2ⁱ
+		if r <= verified {
+			continue
+		}
+		if r > current {
+			break
+		}
```
pavelkalinnikov, line 49: This can't happen because `r <= current` according to loop condition.

```suggestion
```

core/client/meetinthemiddle.go
```diff
@@ -0,0 +1,87 @@
...
+	ret := []int64{} // bⁱ = b-b mod 2ⁱ+ 2ⁱ
+	if created < 1 {
+		created = 1 // Revision 0 is empty.
+	}
+	if current < 1 {
+		return []int64{} // Nothing to verify if the map is empty.
+	}
+	if created > current {
+		created = current
+	}
```
pavelkalinnikov, line 41: This case seems weird. It almost feels like we should panic or return error here.

According to function comment, `created` is the earliest version where an item was found. If `current < created`, we don't know if the value is present in `current` (otherwise `created` would be equal to `current` by definition). Moving it backwards in time may lead to "losing" this data item, and verifying that some irrelevant revisions are consistent instead of the one we are interested in.

core/client/meetinthemiddle.go
```diff
@@ -0,0 +1,87 @@
...
+		}
+		ret = append(ret, r)
+	}
+	return ret
+}
+
+// OlderRevisionsToVerify returns a list of revisions to perform append-only verifications against.
+// The provider / data owner / data verifier periodically verifies that their data history contained in
+// the current map revision contains all the data committed to in select previous revisions.
+// The provider is assured that any relying party has verified that they data
```
pavelkalinnikov, line 58:
```suggestion
// The provider is assured that any relying party has verified that the data
```

core/client/meetinthemiddle.go
```diff
@@ -0,0 +1,87 @@
...
+// they have fetched is represented in at least one of the selected revisions.
+//
+// `current` is the current revision of the map.
+// `verified` is the latest verified revision for this particular data item.
+//
+// TODO: Consier limiting the max period between checks to the expected value
+// of other party lifetimes -- the maximum amount of time this algorithm can
+// tollerate the other party being offline at an unknown point in the past.
+// Currently this algorithm expects the relying party to keep it's memory and
```
pavelkalinnikov, line 67:
```suggestion
// Currently this algorithm expects the relying party to keep its memory and
```

core/client/meetinthemiddle.go
```diff
@@ -0,0 +1,87 @@
...
+// be online at least once every `current = 2ⁱ` revisions. See Tests.
+func OlderRevisionsToVerify(current, verified int64) []int64 {
+	ret := []int64{} // aⁱ = a - a mod 2ⁱ
+	if current < 1 {
+		return []int64{}
+	}
+	for r := current; r >= 0; r = prev(r) {
```
pavelkalinnikov, line 74:
```suggestion
	for r := current; r > 0; r = prev(r) {
```
Otherwise, if it weren't for the `break` below, it would not terminate.

core/client/meetinthemiddle.go
```diff
@@ -0,0 +1,87 @@
...
+		if r <= verified {
+			break
+		}
```
pavelkalinnikov, line 77: This could be embedded into the loop condition as:
`for r := current; r > 0 && r > verified; r = prev(r) {`

Or just `for r := current; r > verified; r = prev(r) {`, if you check that `verified >= 0` before the loop (or just make it unsigned).

core/client/meetinthemiddle.go
```diff
@@ -0,0 +1,87 @@
...
+		ret = append(ret, r)
+	}
+	return ret
+}
+
+// prev return r-2ˡᵉᵛᵉˡ⁽ʳ⁾
```
pavelkalinnikov, line 83:
```suggestion
// prev returns r-2ˡᵉᵛᵉˡ⁽ʳ⁾
```

core/client/meetinthemiddle.go
```diff
@@ -0,0 +1,87 @@
...
+func prev(rev int64) int64 { return rev & (rev - 1) }
+
+// next return r+2ˡᵉᵛᵉˡ⁽ʳ⁾
```
pavelkalinnikov, line 86:
```suggestion
// next returns r+2ˡᵉᵛᵉˡ⁽ʳ⁾
```

core/client/meetinthemiddle.go
```diff
@@ -0,0 +1,87 @@
...
+// OlderRevisionsToVerify algorithm, they will have at least on revision in
+// common.
+//
+// `created` is the (earliest) revision a data item was found in the map.
+//    It is OK to use the revision at which an item was fetched, but the algorithm will be more efficent
+//    if the algorithm starts with the revision at which an item was submitted to the map.
+// `verified` is the latest revision of a successfully verified consistency proof by this client.
+//    When verified > created, previously verified revisions are omitted from the results.
+// `current` is the current revision of the map. Selected revisions will be <= current.
+func NewerRevisionsToVerify(created, current, verified int64) []int64 {
```
pavelkalinnikov, line 31: It would be easier to reason about this code if all the parameters were unsigned. The possibility of negative values makes it necessary to check them before running the loops, or make sure the loops terminate.

core/client/MeetInTheMiddle.md
```diff
@@ -0,0 +1,43 @@
...
+Enable a **sender** Bob at revision *fetched* and a **receiver** Alice at revision *current* to perform 
+append-only proofs between revisions *fetched* <= *intermediate* <= *current*, without coordinating.
+
+### Background
+Key Transparency 1.0 allowed Bob to fetch at an unknown revision, requiring
+Alice to fetch *all* revisions in order to verify her key history, an O(n)
+amount of work where n is the number of revisonis between verifications. This
+limited the frequency and freshness of directory updates.
+
+The meet in the middle algorithm shares the work between Alice and Bob by selecting Log(n)
```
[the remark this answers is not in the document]
gdbelvin: done

core/client/MeetInTheMiddle.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Background
+Key Transparency 1.0 allowed Bob to fetch at an unknown revision, requiring
+Alice to fetch *all* revisions in order to verify her key history, an O(n)
+amount of work where n is the number of revisonis between verifications. This
+limited the frequency and freshness of directory updates.
+
+The meet in the middle algorithm shares the work between Alice and Bob by selecting Log(n) 
+potential intermediates, with the assurance that one of them will be shared between Alice and Bob.
+This brings the verification work down to a level that is Log(n) efficient, regardless of directory freshness.
```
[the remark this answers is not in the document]
gdbelvin: done

core/client/MeetInTheMiddle.md
```diff
@@ -0,0 +1,43 @@
...
+limited the frequency and freshness of directory updates.
+
+The meet in the middle algorithm shares the work between Alice and Bob by selecting Log(n) 
+potential intermediates, with the assurance that one of them will be shared between Alice and Bob.
+This brings the verification work down to a level that is Log(n) efficient, regardless of directory freshness. 
+
+## Algorithm
+
+### Sender Verificatoin
+Clients save the (earliest) revision number associated with the data they have
```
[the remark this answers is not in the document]
gdbelvin: done

core/client/MeetInTheMiddle.md
```diff
@@ -0,0 +1,43 @@
...
+contact, the contact's public key, map revision number it was fetched at, and
+the latest revision it has been verified against.
+
+Senders will then periodically verify that the map has not deleted evidence of
+the public key they are relying on by querying the map at select revisions and
+verifying that the contact's history of public keys is append-only.
+
+The revisions to verify are as follows:
+
+   revⁱ = fetched- fetched mod 2ⁱ+ 2ⁱ
```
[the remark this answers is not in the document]
gdbelvin: done

core/client/MeetInTheMiddle.md
```diff
@@ -0,0 +1,43 @@
...
+
+Receivers save, for their own keys, the latest revision they have verified.
+
+Receivers periodically (Tuneable to their own security tollarance) verify that
+the current version of their key history contains all keys previously contained
+in the map at select previous versions.  
+
+The revisions to verify are as follows:
+
+   revⁱ = current - current mod 2ⁱ
```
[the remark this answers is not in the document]
gdbelvin: done

core/client/MeetInTheMiddle.md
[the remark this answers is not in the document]
gdbelvin: done

core/client/meetinthemiddle.go
```diff
@@ -0,0 +1,87 @@
+// Copyright 2020 Google Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
gdbelvin: done

core/client/meetinthemiddle_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+	"fmt"
+	"math/rand"
+	"testing"
+)
+
+func TestMeetInTheMiddleOverlap(t *testing.T) {
+	for i, tc := range []struct {
+		created,
```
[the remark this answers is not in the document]
gdbelvin: done

core/client/MeetInTheMiddle.md
gdbelvin: done

core/client/meetinthemiddle.go
gdbelvin: done

core/client/meetinthemiddle.go
gdbelvin: done

core/client/meetinthemiddle.go
gdbelvin: Thanks

core/client/meetinthemiddle.go
gdbelvin: Thanks

core/client/meetinthemiddle.go
gdbelvin: This is a bit of a special case to help make randomized testing easier. 
My thought was that this would "check" the latest version if `created > current`

core/client/meetinthemiddle.go
gdbelvin: done

core/client/meetinthemiddle.go
gdbelvin: done

core/client/meetinthemiddle.go
gdbelvin: Thanks

core/client/meetinthemiddle.go
gdbelvin: good idea. Made everything uint64

core/client/meetinthemiddle.go
gdbelvin: done

core/client/meetinthemiddle.go
gdbelvin: done

core/client/meetinthemiddle.go
gdbelvin: done

core/client/MeetInTheMiddle.md
gdbelvin: done

core/client/meetinthemiddle.go
gdbelvin: added panic

docs/design_new.md
```diff
@@ -77,18 +80,28 @@ The value log offers the following API:
...
+in a map of size `N` over `T` snapshots contains roughly `O(T * M log N)`
+hashes.
+1. Instead of verifying that the entire map is an append-only operation from
+   previous values, we isolate the work of verification to individual
+   sub-trees. `O(T * 1 log N)`
+1. Rather than verifying every single snapshot, we only verify the snapshots
+   that the sender and receiver used. `O(1 * 1 log N)`
+1. But the snapshots that the sender and receiver used are unknown, so we use a
+   [meet-in-the-middle algorithm](meet-in-the-middle.md) to sample log
+   T of them. `O(log T * log N)`
```
pavelkalinnikov, line 104:
```suggestion
   [meet-in-the-middle algorithm](meet-in-the-middle.md) to sample O(log T)
   revisions in such a way that they intersect. `O(log T * log N)`
```

core/client/meetinthemiddle.go
```diff
@@ -0,0 +1,79 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package client
+
+// Meet in the middle algorithm: docs/meet-in-the-middle.md
+
+// NewerRevisionsToVerify returns a list of revisions to perform append-only verifications against.
+// The relying party / sender is assured that if there is an entity running the
+// OlderRevisionsToVerify algorithm, they will have at least on revision in
```
pavelkalinnikov, line 21:
```suggestion
// OlderRevisionsToVerify algorithm, they will have at least one revision in
```

core/client/meetinthemiddle.go
```diff
@@ -0,0 +1,79 @@
...
+//    It is OK to use the revision at which an item was fetched, but the
+//    algorithm will be more efficient if it starts with the revision at which
+//    an item was submitted to the map.
+// `verified` is the latest revision of a successfully verified consistency proof by this client.
+//    When verified > created, previously verified revisions are omitted from the results.
+// `current` is the current revision of the map. Selected revisions will be <= current.
+func NewerRevisionsToVerify(created, current, verified uint64) []uint64 {
+	ret := []uint64{} // b[i] = b-b mod 2ⁱ+ 2ⁱ
+	if created < 1 {
+		created = 1 // Revision 0 is empty.
```
pavelkalinnikov, line 34: This line is not covered in tests. Maybe start with `created = 0` in stress-test, or add a manual test to `TestMeetInTheMiddleOverlap`?

core/client/meetinthemiddle.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	if created > current {
+		panic("created > current")
+	}
```
pavelkalinnikov, line 38: Actually, deleting this seems fine. The loop below will terminate immediately returning an empty slice. You can specify in the function comment that `created > current` does not make sense as a precaution for the caller.

```suggestion
```

core/client/meetinthemiddle.go
```diff
@@ -0,0 +1,79 @@
...
+	for r := created; r <= current && r > 0; r = next(r) {
```
pavelkalinnikov, line 39: `r` is always positive, according to how it's initialised and `next`

```suggestion
	for r := created; r <= current; r = next(r) {
```

core/client/meetinthemiddle.go
```diff
@@ -0,0 +1,79 @@
...
+		// check = created + 2ⁱ - created mod 2ⁱ
```
pavelkalinnikov, line 40: This is already mentioned in line 32, and the comment for `next`
```suggestion
```

core/client/meetinthemiddle.go
```diff
@@ -0,0 +1,79 @@
...
+//
+// TODO: Consier limiting the max period between checks to the expected value
+// of other party lifetimes -- the maximum amount of time this algorithm can
+// tollerate the other party being offline at an unknown point in the past.
+// Currently this algorithm expects the relying party to keep its memory and
+// be online at least once every `current = 2ⁱ` revisions. See Tests.
+func OlderRevisionsToVerify(current, verified uint64) []uint64 {
+	ret := []uint64{} // aⁱ = a - a mod 2ⁱ
+	if current < 1 {
+		return []uint64{}
```
pavelkalinnikov, line 67: Please cover this case with a test.

core/client/meetinthemiddle.go
```diff
@@ -0,0 +1,79 @@
...
+	if created < 1 {
+		created = 1 // Revision 0 is empty.
+	}
+	if created > current {
+		panic("created > current")
+	}
+	for r := created; r <= current && r > 0; r = next(r) {
+		// check = created + 2ⁱ - created mod 2ⁱ
+		if r <= verified {
+			continue
```
pavelkalinnikov, line 42: For some reason this line is not covered by tests. Probably because `verified == 0` in all tests? Consider covering it as well.

core/client/meetinthemiddle_test.go
```diff
@@ -0,0 +1,91 @@
...
+import (
+	"fmt"
+	"testing"
+)
+
+func TestAllPairsOverlap(t *testing.T) {
+	max := uint64(100)
+	for created := uint64(1); created < max; created++ {
+		for current := created; current < max; current++ {
+			for diff := uint64(0); diff < 1; diff++ {
```
pavelkalinnikov, line 26: It seems that only `diff = 0` is examined. Did you intend trying multiple values?

docs/design_new.md
gdbelvin: ooh, nicely worded :-)

core/client/meetinthemiddle.go
gdbelvin: done

core/client/meetinthemiddle.go
gdbelvin: added a line for it

core/client/meetinthemiddle.go
gdbelvin: dope

core/client/meetinthemiddle.go
gdbelvin: next used to underflow, now it overflows.

core/client/meetinthemiddle.go
gdbelvin: done

core/client/meetinthemiddle.go
gdbelvin: done

core/client/meetinthemiddle.go
gdbelvin: Yes, I haven't added tests for different versions of verified yet. I'll do that now.

core/client/meetinthemiddle_test.go
gdbelvin: Yes. I'm having a hard time clearly articulating what values of diff are permissible. Maybe you can help me think through this.  eg. I get failure when 
`created: 0,  current: 3, diff: 1` with `newer: [1,3], and older: [4]` no overlap.

pavelkalinnikov: I think the reason might be that `newer` doesn't go beyond `current`. But it has to, in order to reach to number 4?

gdbelvin:
Added more tests. see what you think.

core/client/meetinthemiddle_test.go
pavelkalinnikov: tbh, I'd drop the `diff` parameter here. The ending of the `Newer` range should be >= of the beginning of `Older`, otherwise they don't always meet in the middle. Effectively, `diff == 0` here is necessary for it to always work.

pavelkalinnikov: Or you could actually flip it around, and make `diff` "negative". This way it would also work.

gdbelvin: In many cases, it is ok for the value of `current` as seen by the older algorithm to be ~2x the value of `current` as seen by the newer algorithm.  Maybe this is just `2^(floor(log_2(current))) * 2 - 1 ` which for the above case would be be 3

pavelkalinnikov:
LGTM. Does the "diff should be <= 0" argument make sense?

core/client/meetinthemiddle_test.go
pavelkalinnikov: Suppose the second party's `current` is a power of 2, say 1024. The `prev` call will get them straight to 0, so `OlderRevisionsToVerify` will return only [1024]. The only way the caller of `NewerRevisionsToVerify` intersects with it is by providing `current` >= 1024, which is effectively "diff <= 0". It's true in a general sense.

pavelkalinnikov: I.e. I'm saying "works as intended". Sometimes they indeed can intersect earlier, but the edge case above says the intersection is guaranteed only if the first person monitors at least to the second person's `current`.

gdbelvin:
Yes, I think that makes sense. 

In the next revision of this algorithm, I want to think about ways to make it robust when `current` as seen by alice and bob are within some unknown delta of each other.  They can't really coordinate on the exact same value

pavelkalinnikov:
They can be within a delta from each in the current version of the algorithm, and don't have to be the same. The sufficient condition of intersection is that Bob's `current` is >= Alice's `current` at some point. (assuming Bob calls `Newer...` and Alice calls `Older...`).

pavelkalinnikov:
Note that Alice and Bob call `Older...` and `Newer...` periodically, so the sufficient condition translates to: the maximal of Bob's `current` is >= the minimal of Alice's `current` of the ones that >= Bob's `created`. This means that Bob needs to stay online for sufficiently long for Alice to make at least one monitoring after Bob's `current`.

PR #1567, opened 2020-06-26
Fix a broken link and grammar in design.md

ryanwwest:
Even though this is an older design doc, I still thought others might appreciate having the right proto link.


ryanwwest:
@googlebot I signed it!

PR #1569, opened 2020-06-30
Fix off by one error

gdbelvin:
For all Map Revisions, we want each Map Revision written in the log of map
roots such that Map Revision == Leaf Index.

When Max Map Revision > Max Leaf Index, we want to AddSequencedLeaf all the
revisions inbetween.

### DIFF
--- a/core/sequencer/server.go
+++ b/core/sequencer/server.go
@@ -527,7 +527,7 @@ func (s *Server) PublishRevisions(ctx context.Context,
 		end = batch
 	}
 	logRootTrail.Set(float64(latestMapRoot.Revision - logRoot.TreeSize))
-	for rev := logRoot.TreeSize - 1; rev <= end; rev++ {
+	for rev := logRoot.TreeSize; rev <= end; rev++ {
 		rawMapRoot, mapRoot, err := mapClient.GetAndVerifyMapRootByRevision(ctx, int64(rev))
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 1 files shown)


core/sequencer/server.go
```diff
@@ -527,8 +527,9 @@ func (s *Server) PublishRevisions(ctx context.Context,
 		end = batch
 	}
 	logRootTrail.Set(float64(latestMapRoot.Revision - logRoot.TreeSize))
-	for rev := logRoot.TreeSize - 1; rev <= end; rev++ {
-		rawMapRoot, mapRoot, err := mapClient.GetAndVerifyMapRootByRevision(ctx, int64(rev))
+	maxLeafIndex := logRoot.TreeSize - 1
+	for revToWrite := maxLeafIndex + 1; revToWrite <= end; revToWrite++ {
```
pavelkalinnikov, line 531:
```suggestion
	for revToWrite := logRoot.TreeSize; revToWrite <= end; revToWrite++ {
```

pavelkalinnikov:
LGTM % nit

core/sequencer/server.go
gdbelvin: I know this has fewer lines, but since the math confused me for a hot second, I thought I'd split it out as a favor to future readers.

pavelkalinnikov: The new math does not seem less confusing to me. In fact, it triggers some questions: "hm, this line computes `logRoot.TreeSize - 1`. Does this mean it expects `TreeSize` > 0? Aaaah, there is +1 in the next line, so it's fine". I think what happened is that you got to the bottom of the bug, and got a false sense that changing this code will improve readability. IMO, what would really help to clarify things is a comment (akin to this PR description), and the actual code can be as short as possible.

